PR #1, opened 2021-07-28
Mention forked from bitbucket and fix links

Dragorn421:
https://github.com/Dragorn421/fast64/blob/updatereadme/README.md
https://github.com/Dragorn421/fast64/blob/updatereadme/fast64_internal/sm64/README.md
https://github.com/Dragorn421/fast64/blob/updatereadme/fast64_internal/oot/README.md

this first PR is also a good occasion to discuss about merging PRs: I think we should squash merge to keep the history as clean as can be and avoid "wip commits"/broken revisions. also makes it easier to git blame

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
-
 # Fast64
 
 This requires Blender 2.82+.
 
-![alt-text](https://bitbucket.org/kurethedead/fast64/raw/master/images/mario_running.gif)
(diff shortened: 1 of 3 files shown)


thecozies (approved):
looks good, the relative links are much better too

Zelllll (approved):
Looks good to me, how do we set pull requests to be squashed rather than containing every commit?

PR #3, opened 2021-07-30
Custom gSPSegment and improve sequence numbering

sauraen:
- Add 1 to sequence number in export so that in-game sequence number is as expected
- Add two optional custom gSPSegment addresses to material editor

### DIFF
--- a/fast64_internal/oot/oot_f3d_writer.py
+++ b/fast64_internal/oot/oot_f3d_writer.py
@@ -640,9 +640,9 @@ class OOTDynamicMaterialDrawLayerProperty(bpy.types.PropertyGroup):
 	segmentC : bpy.props.BoolProperty()
 	segmentD : bpy.props.BoolProperty()
 	customCall0 : bpy.props.BoolProperty()
-	customCall0_seg : bpy.props.StringProperty()
+	customCall0_seg : bpy.props.StringProperty(description="Segment address of a display list to call, e.g. 0x08000010")
 	customCall1 : bpy.props.BoolProperty()
-	customCall1_seg : bpy.props.StringProperty()
+	customCall1_seg : bpy.props.StringProperty(description="Segment address of a display list to call, e.g. 0x08000010")
 
 # The reason these are separate is for the case when the user changes the material draw layer, but not the 
 # dynamic material calls. This could cause crashes which would be hard to detect.
--- a/fast64_internal/oot/oot_f3d_writer.py
+++ b/fast64_internal/oot/oot_f3d_writer.py
@@ -39,7 +39,7 @@ class OOTModel(FModel):
 				gfxList.commands.append(SPDisplayList(
 					GfxList("0x" + format(i,'X') + '000000', GfxListTag.Material, DLFormat.Static)))
 		for i in range(0, 2):
-			p = "custom" + str(i)
+			p = "customCall" + str(i)
 			if getattr(matDrawLayer, p):
 				gfxList.commands.append(SPDisplayList(
 					GfxList(getattr(matDrawLayer, p + "_seg"), GfxListTag.Material, DLFormat.Static)))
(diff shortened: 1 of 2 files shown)


fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -601,6 +605,11 @@ def drawOOTMaterialDrawLayerProperty(layout, matDrawLayerProp, suffix):
 			i = 8 + colIndex * 3 + rowIndex
 			name = "Segment " + format(i, 'X') + " " + suffix
 			col.prop(matDrawLayerProp, "segment" + format(i, 'X'), text = name)
+		name = "Custom " + str(colIndex) + " " + suffix
```
Dragorn421, line 608:
```suggestion
		name = "Jump to (" + str(colIndex + 1) + ") " + suffix
```
I think "custom" is a pretty bad name to use for everything (way too generic), maybe change `custom` to `customJump`

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -630,6 +639,10 @@ class OOTDynamicMaterialDrawLayerProperty(bpy.types.PropertyGroup):
 	segmentB : bpy.props.BoolProperty()
 	segmentC : bpy.props.BoolProperty()
 	segmentD : bpy.props.BoolProperty()
+	custom0 : bpy.props.BoolProperty()
+	custom0_seg : bpy.props.StringProperty()
```
Dragorn421, line 643:
```suggestion
	custom0_seg : bpy.props.StringProperty(description="Segment address of a display list to jump to")
```

Dragorn421 (changes requested):
Sorry, didn't see the PRs until right now by accident. Didn't enable notifications for the repo

Ideally we would want to allow an arbitrary amount of such "custom" jumps but two is probably plenty enough, and a CollectionProperty needs more work

Dragorn421:
hijacking this pr to test the webhook

Zelllll:
Yep I agreed with Dragorn's suggestions, although eventually I'd like to simply remove all the check boxes too and just let users choose as many segments as they want instead, I think it would be worth breaking compatibility with existing stuff to make that change.

Dragorn421:
First

sauraen:
I started working on changing over the segments to a collection, but I realized that the fast64 infrastructure for creating GUI for editing collections is based on the collection in question being some sub-property of an object. This collection would be a sub-property of a material, which means I'd have to make completely new GUI for this or make substantial changes to the existing infrastructure. Combined with the fact that this will break scenes (or the extra work of checking for both types of segment commands), I don't think it's worth it.

I've changed the names to "custom call". It's more like a function call than a jump, as execution returns to the current DL.

Zelllll:
Hm I was thinking using the same gui as the object list, just simply have an add button for a segment, and also allowing you to reorder them so they get called in the specific order you want

sauraen:
Right, so I tried that, and it does not work, because the GUI infrastructure for displaying the elements of a collection with the add/remove/up/down buttons etc. all depends on the collection you're editing being a sub-property of an object in the scene. This collection would be a sub-property of a material, not an object. So it is not compatible with that GUI infrastructure.

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -630,6 +639,10 @@ class OOTDynamicMaterialDrawLayerProperty(bpy.types.PropertyGroup):
 	segmentB : bpy.props.BoolProperty()
 	segmentC : bpy.props.BoolProperty()
 	segmentD : bpy.props.BoolProperty()
+	customCall0 : bpy.props.BoolProperty()
+	customCall0_seg : bpy.props.StringProperty()
+	customCall1 : bpy.props.BoolProperty()
+	customCall1_seg : bpy.props.StringProperty()
```
Dragorn421, line 645:
```suggestion
	customCall0_seg : bpy.props.StringProperty(description="Segment address of a display list to call")
	customCall1 : bpy.props.BoolProperty()
	customCall1_seg : bpy.props.StringProperty(description="Segment address of a display list to call")
```
I plan to eventually add a description for every property to make it more user friendly, may as well start now.
Those are displayed in the tooltip when hovering over a property, imo it's a great way to document whatever a property is for

Dragorn421 (commented):
I did expect `CollectionProperty` to be some work, but not because of fast64, as I said before it will be fine as is, two custom calls is plenty enough.

sauraen:
I've added the descriptions you requested.

Dragorn421 (approved):

PR #4, opened 2021-07-30
Write OoT scene lighting data as hex

sauraen:
Match decomp tools formatting

### DIFF
--- a/fast64_internal/oot/c_writer/oot_level_c.py
+++ b/fast64_internal/oot/c_writer/oot_level_c.py
@@ -471,7 +471,7 @@ def ootExitListToC(scene, headerIndex):
 	return data
 
 def ootVectorToC(vector):
-	return str(vector[0]) + ', ' + str(vector[1]) + ', ' + str(vector[2])
+	return "0x{:02X}, 0x{:02X}, 0x{:02X}".format(vector[0], vector[1], vector[2])
 
 def ootLightToC(light):
 	return "\t{ " + \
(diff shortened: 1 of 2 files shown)


Dragorn421 (commented):
This includes changes from #3 so we'll get the other PR merged first (or you can edit your branch, as you want), but otherwise this looks good.

sauraen:
I needed to match the lighting settings of a vanilla scene, so I needed to compare the values without putting every one through a hex converter. If someone wants to make a PR to change the decomp toolchain instead, and reject this PR, that's fine with me.

Zelllll:
I'm fine with merging this for now but eventually it needs fixed decomp side, however they decided to ban me so RIP

fig02:
I would just wait personally and we'll change whatever is needed decomp side.
Whats the issue exactly? Lighting stuff in light settings from rooms? So a zapd change?

Zelllll:
Another issue is that the struct is totally wrong, and I fixed it in MM already actually I think, and you also have it fixed in your Kankyo branch except you made a duplicate struct. Anyways a bunch of struct members are erroneously swapped for some reason.

Zelllll:
And also the lighting rgb should be in decimal rather than hex since we always do colors that way, that was the main issue here

fig02:
if its dependent on kankyo stuff, then this hotfix seems appropriate

Zelllll:
Alright this should be good then but I believe Dragorn said that this pr also includes other unrelated changes that need reverted, right?

Dragorn421:
Yeah, this includes changes from #3
either revert the changes in a new commit, or:
`git checkout LightingWriteHex`
`git reset --soft main` (resets the branch history to main but doesn't affect working directory)
`git reset -p` and reset index changes related to #3
`git commit`
`git push --force` (overwrite the current history)

sauraen:
I've rebased the branch as you requested. I'm sorry for the trouble, these branches are from the bitbucket era where merges did not happen often. I needed to not have to switch back and forth between branches to actually use the plugin.

Dragorn421 (approved):

PR #8, opened 2021-08-05
Fixed cutscene editing only for primary scene header

sauraen:
Most operators / GUI code for cutscene editing was hard-coded to use only the main scene header. Adding commands to cutscenes in other headers would actually add them to the cutscene in the main header, and there was no way to add or edit commands in other headers. This has been fixed. Basic cases of editing and exporting multiple cutscenes in one scene have been tested and work.

### DIFF
--- a/fast64_internal/oot/oot_scene_room.py
+++ b/fast64_internal/oot/oot_scene_room.py
@@ -677,7 +677,7 @@ def drawSceneHeaderProperty(layout, sceneProp, dropdownLabel, headerIndex, objNa
 			r.prop(sceneProp, "csTermIdx", text = "Index")
 			r.prop(sceneProp, "csTermStart", text = "Start Frm")
 			r.prop(sceneProp, "csTermEnd", text = "End Frm")
-		tempHeaderIndex = headerIndex - 4 if headerIndex is not None and headerIndex >= 4 else -1
+		tempHeaderIndex = 0 if headerIndex is None else headerIndex
 		for i, p in enumerate(sceneProp.csLists):
 			drawCSListProperty(cutscene, p, i, objName, tempHeaderIndex)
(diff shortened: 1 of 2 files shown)


Zelllll (approved):
Looks good to me

fast64_internal/oot/oot_scene_room.py
```diff
@@ -672,9 +677,10 @@ def drawSceneHeaderProperty(layout, sceneProp, dropdownLabel, headerIndex, objNa
 			r.prop(sceneProp, "csTermIdx", text = "Index")
 			r.prop(sceneProp, "csTermStart", text = "Start Frm")
 			r.prop(sceneProp, "csTermEnd", text = "End Frm")
+		tempHeaderIndex = headerIndex - 4 if headerIndex is not None and headerIndex >= 4 else -1
```
Dragorn421, line 680: does this mean you can't add cutscenes to the "default alternate headers" aka child night, adult day and adult night?

sauraen: Thanks for pointing that out. When I was initially adding cutscene support for some reason I thought the existing fast64 code was set up to only permit an entrance cutscene and cutscenes in cutscene headers (not headers 1-3). So, I was intending to just follow this. However, I've looked again and it seems there's no reason the scene file format couldn't support cutscenes in headers 1-3, even if this is never used in game. So, I'll fix this to add that support.

Dragorn421: then you can get rid of `tempHeaderIndex` and just use `headerIndex` directly, have 0 stand for the main header instead of -1 standing for (all of?) the first four headers

sauraen:
Fixed, added support for alternate headers 1-3.

Dragorn421 (approved):

PR #10, opened 2021-08-08
Fix cs header typo

sauraen:
Fix typo in cutscene header selection in getCollection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
-
 # Fast64
 
 This requires Blender 2.82+.
 
-![alt-text](https://bitbucket.org/kurethedead/fast64/raw/master/images/mario_running.gif)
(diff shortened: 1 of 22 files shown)


sauraen:
Also, I do not want to merge 14 commits, I only want to merge one commit. It is only a one line change. I merged upstream main into my own main, then made the branch just for one change. Not sure why GitHub thinks the last 14 commits apply.

Dragorn421 (approved):
As Zel said on Discord it's because we squash merge to avoid flooding fast64's commit history with all commits of every pr

In this instance it's nice to have since it also means we don't have to care about your PR being 14 or 1 commits, as whatever the diff ends up being will turn into one single commit

I suggest branching from a clean fast64/main in the future (or `git reset fast64/main` to set the history to fast64/main's)

PR #14, opened 2021-08-19
[OoT] Missing one "Scene" not changed to "SceneTableEntry" causing exporter to write Scene in z_scene_table.c

Yanis42:
My bad


Dragorn421 (approved):
kek

PR #15, opened 2021-09-01
Allow memory.c to be in boot/ or game/

thecozies:
Just uses `os.path.exists` to determine where the file is!


Zelllll (approved):
Looks good!

PR #16, opened 2021-09-03
Custom geo commands and object instancing in SM64 Object exporting

thecozies:
Regarding the number of changes:
I enabled "Remove trailing whitespace" in VSCode... There is a TON of trailing whitespace throughout this repo so you may want to disable viewing whitespace changes when reviewing!

Instancing:
In a nutshell, this allows people to use Alt D to create object links, and those links will be exported as the same displaylist. Its a bit tricky because all transformations need to happen in the GeoLayout itself, instead of applying all transformations. This needed to happen in order to get these custom geolayout commands to work, since applying their transformations would break the intended transformation.

Custom Geolayout Commands:
Before this change, there were only a small handful of geolayout commands you could use with objects, while armatures were a low more flexible (no pun intended) and you could basically do any geolayout command wherever you wanted. Now people can attach any number of functions anywhere, use custom macros, scale, billboard, etc. Its pretty useful! Removes the need for most edits after exporting.

Maybe a similar system could be done for OOT. Sharing displaylists is pretty useful.

### DIFF
--- a/fast64_internal/sm64/sm64_geolayout_classes.py
+++ b/fast64_internal/sm64/sm64_geolayout_classes.py
@@ -17,8 +17,13 @@ drawLayerNames = {
 }
 
 def getDrawLayerName(drawLayer):
-	# Cast draw layer to int so it can be mapped to a name
-	layer = int(drawLayer) if isinstance(drawLayer, str) and drawLayer.isnumeric() else drawLayer
+	layer = drawLayer
+	if drawLayer is not None:
+		try:
+			# Cast draw layer to int so it can be mapped to a name
+			layer = int(drawLayer)
+		except ValueError:
+			pass
 	if layer in drawLayerNames:
 		return drawLayerNames[layer]
 	else:
--- a/fast64_internal/sm64/sm64_geolayout_classes.py
+++ b/fast64_internal/sm64/sm64_geolayout_classes.py
@@ -17,8 +17,10 @@ drawLayerNames = {
 }
 
 def getDrawLayerName(drawLayer):
-	if drawLayer in drawLayerNames:
-		return drawLayerNames[drawLayer]
+	# Cast draw layer to int so it can be mapped to a name
+	layer = int(drawLayer) if isinstance(drawLayer, str) and drawLayer.isnumeric() else drawLayer
+	if layer in drawLayerNames:
+		return drawLayerNames[layer]
 	else:
 		return str(drawLayer)
 
@@ -207,7 +209,37 @@ class Geolayout:
 		for node in self.nodes:
 			drawLayers |= node.getDrawLayers()
 		return drawLayers
-		
+
+class BaseDisplayListNode:
+	'''Base displaylist node with common helper functions dealing with displaylists'''
+	dl_ext = 'WITH_DL' # add dl_ext to geo command if command has a displaylist
+
+	def get_dl_address(self):
+		if self.hasDL and (self.dlRef or self.DLmicrocode is not None):
+			return self.dlRef or self.DLmicrocode.startAddress
+		return None
+
+	def get_dl_name(self):
+		if self.hasDL and (self.dlRef or self.DLmicrocode is not None):
+			return self.dlRef or self.DLmicrocode.name
+		return 'NULL'
+	
+	def get_c_func_macro(self, base_cmd: str):
+		return f'{base_cmd}_{self.dl_ext}' if self.hasDL else base_cmd
+
+	def c_func_macro(self, base_cmd: str, *args: str):
+		'''
+			Supply base command and all arguments for command.
+			if self.hasDL:
+				this will add self.dl_ext to the command, and
+				adds the name of the displaylist to the end of the command
+			Example return: 'GEO_YOUR_COMMAND_WITH_DL(arg, arg2),'
+		'''
+		all_args = list(args)
+		if self.hasDL:
+			all_args.append(self.get_dl_name())
+		return f'{self.get_c_func_macro(base_cmd)}({", ".join(all_args)}),'
+
 class TransformNode:
 	def __init__(self, node):
 		self.node = node
@@ -327,7 +359,8 @@ class TransformNode:
 		else:
 			drawLayers = set()
 		for child in self.children:
-			drawLayers |= child.getDrawLayers()
+			if hasattr(child, 'getDrawLayers'): # not every child will have draw layers (e.g. GEO_ASM)
+				drawLayers |= child.getDrawLayers()
 		return drawLayers
 
 class SwitchOverrideNode:
@@ -339,10 +372,11 @@ class SwitchOverrideNode:
 		self.texDimensions = texDimensions # None implies a draw layer override
 
 class JumpNode:
-	def __init__(self, storeReturn, geolayout):
+	def __init__(self, storeReturn, geolayout, geoRef: str = None):
 		self.geolayout = geolayout
 		self.storeReturn = storeReturn
 		self.hasDL = False
+		self.geoRef = geoRef
 	
 	def size(self):
 		return 8
@@ -352,8 +386,8 @@ class JumpNode:
 
 	def to_binary(self, segmentData):
 		if segmentData is not None:
-			startAddress = encodeSegmentedAddr(self.geolayout.startAddress,
-				segmentData)
+			address = self.geoRef or self.geolayout.startAddress
+			startAddress = encodeSegmentedAddr(address, segmentData)
 		else:
 			startAddress = bytearray([0x00] * 4)
 		command = bytearray([GEO_BRANCH, 
@@ -362,8 +396,9 @@ class JumpNode:
 		return command
 
 	def to_c(self):
+		geo_name = self.geoRef or self.geolayout.name
 		return "GEO_BRANCH(" + ('1, ' if self.storeReturn else '0, ') + \
-			self.geolayout.name + '),'
+			geo_name + '),'
 
 def convertAddrToFunc(addr):
 	if addr == '':
@@ -472,8 +507,8 @@ class SwitchNode:
 			str(self.defaultCase) + ', ' +\
 			convertAddrToFunc(self.switchFunc) + '),'
 
-class TranslateRotateNode:
-	def __init__(self, drawLayer, fieldLayout, hasDL, translate, rotate):
+class TranslateRotateNode(BaseDisplayListNode):
+	def __init__(self, drawLayer, fieldLayout, hasDL, translate, rotate, dlRef: str = None):
 
 		self.drawLayer = drawLayer
 		self.fieldLayout = fieldLayout
@@ -484,6 +519,7 @@ class TranslateRotateNode:
 
 		self.fMesh = None
 		self.DLmicrocode = None
+		self.dlRef = dlRef
 	
 	def get_ptr_offsets(self):
 		if self.hasDL:
@@ -515,6 +551,8 @@ class TranslateRotateNode:
 	def to_binary(self, segmentData):
 		params = ((1 if self.hasDL else 0) << 7) & \
 			(self.fieldLayout << 4) | int(self.drawLayer)
+		
+		start_address = self.get_dl_address()
 
 		command = bytearray([GEO_TRANSLATE_ROTATE, params])
 		if self.fieldLayout == 0:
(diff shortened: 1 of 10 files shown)


thecozies:
# Here's how you can hide whitespace changes!

image

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -152,17 +154,18 @@ def getCameraObj(camera):
...
+	fModel.materialRevert = GfxList(fModel.name + "_" + 'material_revert_render_settings',
 		GfxListTag.MaterialRevert, fModel.DLFormat)
-	revertMatAndEndDraw(fModel.materialRevert, 
+	revertMatAndEndDraw(fModel.materialRevert,
 		[DPSetEnvColor(0xFF, 0xFF, 0xFF, 0xFF),
 		DPSetAlphaCompare("G_AC_NONE")])
 
-	drawLayers = geolayoutGraph.getDrawLayers()
+	# Get all draw layers, turn layers into strings (some are ints), deduplicate using a set
+	drawLayers = set([str(layer) for layer in geolayoutGraph.getDrawLayers()])
```
Dragorn421, line 164:
```suggestion
	drawLayers = set(str(layer) for layer in geolayoutGraph.getDrawLayers())
```

also is it certain that at this point `getDrawLayers` is defined? (I saw in other places that it can be undefined for some geonodes)
I don't need a detailed explanation, just yes or no is enough

fast64_internal/sm64/sm64_objects.py
```diff
@@ -132,8 +134,46 @@
...
+
+inlineGeoLayoutObjects = {
+	'Geo ASM'				: InlineGeolayoutObjConfig('Geo ASM', 				FunctionNode),
+	'Geo Branch'			: InlineGeolayoutObjConfig('Geo Branch', 			JumpNode,
+								must_have_geo=True),
+	'Geo Translate/Rotate'	: InlineGeolayoutObjConfig('Geo Translate/Rotate', 	TranslateRotateNode,
+								can_have_dl=True, uses_location=True, uses_rotation=True),
+	'Geo Translate Node'	: InlineGeolayoutObjConfig('Geo Translate Node', 	TranslateNode,
+								can_have_dl=True, uses_location=True),
+	'Geo Rotation Node'		: InlineGeolayoutObjConfig('Geo Rotation Node', 	RotateNode,
```
Dragorn421, line 164: indentation (purely visual) looks messed up in `inlineGeoLayoutObjects` but maybe it's just github

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+def yield_children(obj: bpy.types.Object):
+	yield obj
+	if obj.children:
+		for o in obj.children:
+			yield from yield_children(o)
+
+def iter_from_active():
+	active_obj = bpy.context.view_layer.objects.active # keep original obj reference
+	yield from yield_children(active_obj)
+	bpy.context.view_layer.objects.active = active_obj
```
Dragorn421, line 537: that's pretty weird behavior imo,
I would make code calling `iter_from_active` responsible for preserving the context (active object),
rather than doing it in the generator function and implicitly restore context once iteration is done

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+
+def store_original_mtx():
+	for obj in iter_from_active():
+		obj['original_mtx'] = obj.matrix_local
+
+def rotate_bounds(bounds, mtx: mathutils.Matrix):
+	return [
+		(mathutils.Vector(b[:]) @ mtx).to_tuple()
+		for b in bounds[:]
```
Dragorn421, line 546:
```suggestion
		(mathutils.Vector(b) @ mtx).to_tuple()
		for b in bounds
```
I don't think you need to make copies, do you

also is there a reason for `mathutils.Vector(b) @ mtx` instead of the more common `mtx @ mathutils.Vector(b)`

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+	]
+
+def obj_scale_is_unified(obj):
+	'''Combine scale values into a set to ensure all values are the same'''
+	return len(set(obj.scale)) == 1
+
+def translation_rotation_from_mtx(mtx: mathutils.Matrix):
+	'''Strip scale from matrix'''
+	t, r, _ = mtx.decompose()
+	return Matrix.Translation(t).to_4x4() @ r.to_matrix().to_4x4()
```
Dragorn421, line 556:
```suggestion
	return Matrix.Translation(t) @ r.to_matrix().to_4x4()
```

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+
+def scale_mtx_from_vector(scale: mathutils.Vector):
+	scax = mathutils.Matrix.Scale(scale[0], 4, (1, 0, 0))
+	scay = mathutils.Matrix.Scale(scale[1], 4, (0, 1, 0))
+	scaz = mathutils.Matrix.Scale(scale[2], 4, (0, 0, 1))
+	return scax @ scay @ scaz
```
Dragorn421, line 562:
```suggestion
	return mathutils.Matrix.Diagonal(scale[0:3]).to_4x4()
```

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+
+def copy_object_and_apply(obj: bpy.types.Object, apply_scale = False, apply_modifiers = False):
+	if apply_scale or apply_modifiers:
+		# its a unique mesh, use object name
```
Dragorn421, line 566:
```suggestion
		# it's a unique mesh, use object name
```

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+		prev_active = bpy.context.view_layer.objects.active
+		bpy.context.collection.objects.link(obj_copy)
+		obj_copy.select_set(True)
+		bpy.context.view_layer.objects.active = obj_copy
+		for modifier in obj_copy.modifiers:
+			attemptModifierApply(modifier)
+
+		bpy.context.view_layer.objects.active = prev_active
+
+	mtx = mathutils.Quaternion((1, 0, 0), math.radians(-90.0)).to_matrix().to_4x4()
```
Dragorn421, line 591: what is the purpose of this? going from blender coordinates to sm64 coordinates?
if so I would just hardcode the matrix
```suggestion
	mtx = Matrix(((1,  0, 0, 0),
                    (0,  0, 1, 0),
                    (0, -1, 0, 0),
                    (0,  0, 0, 1)))
```
x -> x
y -> -z
z -> y

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+
+	obj_copy.data.transform(mtx)
+	# Flag used for finding these temp objects
+	obj_copy['temp_export'] = True
+
+	# Override for F3D culling bounds (used in addCullCommand)
+	bounds_mtx = scale_mtx_from_vector([1, -1, -1]) # reverse y/z
+	if apply_scale:
+		bounds_mtx *= scale_mtx_from_vector(obj.scale) # apply scale if needed
+	bounds_mtx = bounds_mtx @ mathutils.Quaternion((1, 0, 0), math.radians(-90.0)).to_matrix().to_4x4()
```
Dragorn421, line 603: may want to store the previous matrix I commented on and reuse it here

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+			has_modifiers = len(obj.modifiers) != 0
+			has_uneven_scale = not obj_scale_is_unified(obj)
+			shares_mesh = obj.data.users > 1
+			if has_modifiers or has_uneven_scale:
+				if shares_mesh and has_modifiers:
+					add_warning(
+						f'Object "{obj.name}" cannot be instanced due to having modifiers so an extra displaylist will be created. Remove modifiers to allow instancing.')
+				if shares_mesh and has_uneven_scale:
+					add_warning(
+						f'Object "{obj.name}" cannot be instanced due to uneven object scaling and an extra displaylist will be created. Set all scale values to the same value to allow instancing.')
```
Dragorn421, line 623: I don't know why uniform scaling currently works but uneven scaling doesn't, but regardless I think uneven scaling could be implemented by multiplying the matrix on the rsp stack by whatever is needed (see `gsSPMatrix`)

Just saying I think it can be done, I'm not asking you to make it happen

fast64_internal/utility.py
```diff
@@ -523,6 +525,118 @@ def deleteIfFound(filePath, stringValue):
...
+			obj.original_name = obj.name
+
+			if obj.data.name not in instanced_meshes:
+				instanced_meshes.add(obj.data.name)
+				copy_object_and_apply(obj)
+
+def get_obj_temp_mesh(obj):
+	for o in bpy.data.objects:
+		if o.get('temp_export') and \
+			o.get('instanced_mesh_name') == obj.get('instanced_mesh_name'):
```
Dragorn421, line 637:
```suggestion
		if (
			o.get('temp_export')
			and o.get('instanced_mesh_name') == obj.get('instanced_mesh_name')
		):
```

fast64_internal/utility.py
```diff
@@ -560,17 +677,48 @@ def duplicateHierarchy(obj, ignoreAttr, includeEmpties, areaIndex):
...
 	isMesh = isinstance(obj.data, bpy.types.Mesh)
-	isEmpty = (obj.data is None) and includeEmpties and \
-		(obj.sm64_obj_type == 'Level Root' or \
-		obj.sm64_obj_type == 'Area Root' or \
-		obj.sm64_obj_type == 'None' or \
-		obj.sm64_obj_type == 'Switch')
+	isEmpty = \
+		(obj.data is None) and \
+		includeEmpties and \
+		checkSM64EmptyUsesGeoLayout(obj.sm64_obj_type)
```
Dragorn421, line 721:
```suggestion
	isEmpty = (
		obj.data is None
		and includeEmpties
		and checkSM64EmptyUsesGeoLayout(obj.sm64_obj_type)
	)
```

fast64_internal/sm64/sm64_geolayout_classes.py
```diff
@@ -17,8 +17,10 @@
 }
 
 def getDrawLayerName(drawLayer):
-	if drawLayer in drawLayerNames:
-		return drawLayerNames[drawLayer]
+	# Cast draw layer to int so it can be mapped to a name
+	layer = int(drawLayer) if isinstance(drawLayer, str) and drawLayer.isnumeric() else drawLayer
```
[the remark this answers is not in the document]
thecozies: haha those tabs sizes are HUGE! and i _think_ that isnumeric is what i'm looking for https://docs.python.org/3/library/stdtypes.html#str.isnumeric

your method makes sense to me, though i think the try/catch is a little bulky. what i was trying to do here is check if drawLayer is a string that only contains numbers. the other possible values here are C macros like `"LAYER_TRANSPARENT"`. I'm not exactly sure how else a `ValueError` could happen here but if a ValueError popped up with this specific condition i think that i would want it to throw an error because it would be some wildly unexpected value. i'm open to changing it though

fast64_internal/sm64/sm64_geolayout_writer.py
thecozies: > also is it certain that at this point getDrawLayers is defined
yup!

also, great suggestion lol

fast64_internal/sm64/sm64_objects.py
thecozies: wow yeah github DESTROYED the formatting.. should i align with spaces instead? my local tab size is 4 and i'm used to using spaces as tabs

image

fast64_internal/utility.py
thecozies: weird yeah i don't need to make copies, for some reason i thought i had to when testing things out 🤔 changing that now.

> also is there a reason for `mathutils.Vector(b) @ mtx` instead of the more common `mtx @ mathutils.Vector(b)`

yesss i want to rotate the vector from the matrix, the other way around is, translating the matrix from the vector? not sure exactly but the other way jacks up the bounds

fast64_internal/utility.py
thecozies: i was originally going to store it somewhere but the shared references are a big scary since the first variable `mtx` needs to be modified. i could make a lambda that returns a new one instead?

image

fast64_internal/utility.py
thecozies: yeah thats definitely a good point, I'm avoiding it for now because I want this PR to be based in GeoLayout transformations (which are only uniform unfortunately). sometime in the future that'd be great shared functionality between oot/sm64

fast64_internal/utility.py
thecozies: https://github.com/Fast-64/fast64/pull/16#discussion_r702184244

I'm turning this into a lambda that returns this matrix b/c shared references wont work here

thecozies:
@Dragorn421 I resolved all of the suggestions where I did the exact thing you suggested, and then left the suggestions open that you had questions about

fast64_internal/sm64/sm64_geolayout_classes.py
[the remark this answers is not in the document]
Dragorn421: `isnumeric` allows many more codepoints than just 0-9 characters
> Numeric characters include digit characters, and all characters that have the Unicode numeric value property, e.g. U+2155, VULGAR FRACTION ONE FIFTH.
for example with U+2155
```
>>> int("\u2155")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '⅕'
```

it's quite common to use `except ValueError` to detect if something can't be converted
maybe you would prefer something like `all(c >= "0" and c <= "9" for c in drawLayer)`

fast64_internal/sm64/sm64_objects.py
Dragorn421: PEP-8 recommends indenting with 4-spaces instead of tabs and I'm definitely on that ship since various text editors can't handle tabs consistently... I'd rather have all indentation changed in a cleanup PR though so if it looks fine on your end then good

fast64_internal/utility.py
```diff
@@ -523,6 +531,115 @@ def deleteIfFound(filePath, stringValue):
...
+			yield from yield_children(o)
+
+def store_original_mtx():
+	active_obj = bpy.context.view_layer.objects.active
+	for obj in yield_children(active_obj):
+		obj['original_mtx'] = obj.matrix_local
+
+def rotate_bounds(bounds, mtx: mathutils.Matrix):
+	return [
+		(mathutils.Vector(b) @ mtx).to_tuple()
```
Dragorn421, line 547: Are you aware this means the rotation is inverted? `v @ mtx` is the same as `mtx.transposed() @ v`, and transpose on a rotation is the same as inverting it
If you're only using this to ditch the translation part you should use `mtx.decompose()` similarly to your `translation_rotation_from_mtx` function

fast64_internal/utility.py
```diff
@@ -991,4 +1163,20 @@ def read16bitRGBA(data):
...
+
+def translate_xyz_to_xzy(translate: mathutils.Vector):
+	xzy_translate = translate.xzy
+	xzy_translate[2] = -xzy_translate[2]
+	return xzy_translate
+
+def rotation_xyz_quat_to_xzy_quat(rotation: mathutils.Quaternion):
+    euler_rot = rotation.to_matrix().inverted().to_euler('ZXY')
+    new_rot = mathutils.Euler([-euler_rot.x, -euler_rot.z, euler_rot.y], 'ZXY')
+    return new_rot.to_quaternion()
```
Dragorn421, line 1179: these look very sketchy, I feel like using the blender_to_n64 matrix instead could clean it up, if that's what is meant to be done here

fast64_internal/sm64/sm64_objects.py
thecozies: awesome, yeah it'd be good to enable PEP-8 auto formatting sometime soon

fast64_internal/sm64/sm64_geolayout_classes.py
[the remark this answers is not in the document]
thecozies: ahhh yeah, good point. i'll go with the ValueError check then b/c its probably a bit cheaper

fast64_internal/utility.py
thecozies: do you know how i should use that here? I can't get any other variation of this function to work, I probably spent like 4 hours on this function alone lmao it was killing me

thecozies: i'm messing with variations of this right now and cant find anything that produces the correct output
```python
def rotation_xyz_quat_to_xzy_quat(rotation: mathutils.Quaternion):
    new_rot = (rotation.to_matrix().to_4x4() @ transform_mtx_blender_to_n64())
    return new_rot.to_quaternion()
```

Dragorn421: maybe
```py
def rotation_xyz_quat_to_xzy_quat(rotation: mathutils.Quaternion):
    new_rot = (transform_mtx_blender_to_n64() @ rotation.to_matrix().to_4x4() @ transform_mtx_blender_to_n64().invert())
    return new_rot.to_quaternion()
```
assuming `rotation` is in "blender coordinates" and `new_rot` in "n64 coordinates"

thecozies: awesome!! yeah that works perfectly, identical output. `def rotate_quat_blender_to_n64`?

Dragorn421: sure

fast64_internal/utility.py
thecozies: using `translation_rotation_from_mtx` isn't working here, but there shouldn't be any translation in the matrix anyways. i can change it to `mtx.transposed() @ v` if you think its clearer, i just don't understand exactly why that produces the same output in this scenario

fast64_internal/utility.py
```diff
@@ -991,4 +1162,19 @@ def read16bitRGBA(data):
...
\ No newline at end of file
+	return [r,g,b,a]
+
+def join_c_args(args: 'list[str]'):
+	return ', '.join(args)
+
+def translate_xyz_to_xzy(translate: mathutils.Vector):
+	xzy_translate = translate.xzy
+	xzy_translate[2] = -xzy_translate[2]
+	return xzy_translate
```
Dragorn421, line 1173:
```suggestion
	return transform_mtx_blender_to_n64() @ xzy_translate
```

thecozies: lmao all my matrix math is very suspicious now.

just pushed this changed and renamed it to `translate_blender_to_n64`

Dragorn421 (approved):

PR #17, opened 2021-09-03
Have all exports use global 'save textures as PNG'

thecozies:
Like basic F3D settings, all `save textures as PNG` toggles are just in the F3D menu as a global setting (helps prevent people from accidentally pressing it)

### DIFF
--- a/fast64_internal/f3d/f3d_writer.py
+++ b/fast64_internal/f3d/f3d_writer.py
@@ -2348,7 +2348,7 @@ class F3D_ExportDL(bpy.types.Operator):
 			f3dType = context.scene.f3d_type
 			isHWv1 = context.scene.isHWv1
 			texDir = bpy.context.scene.DLTexDir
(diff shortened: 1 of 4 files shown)


Zelllll:
This looks good so far! One thing I'm noticing is that none of the oot code seems to be affected by this, though I might be wrong. Can you look into that? (This is the sort of annoying issues though that I've been talking about with how each game has duplicate code)

thecozies:
> This looks good so far! One thing I'm noticing is that none of the oot code seems to be affected by this, though I might be wrong. Can you look into that? (This is the sort of annoying issues though that I've been talking about with how each game has duplicate code)

OOT already defaults to the global settings!

thecozies:
@Zelllll 
image
all instances of saveTextures in oot look for `scene.saveTextures`

Zelllll:
Alright!

Zelllll (approved):

Dragorn421 (approved):

PR #18, opened 2021-09-05
OoT cutscenes as first-class objects

sauraen:
This PR adds "cutscene" as a new OoT empty type. This enables several features:
- Export of cutscenes to actor / object C files, not only scene C files
- Cutscenes can be shared across multiple scene headers even if the headers are otherwise different
- Scenes can have arbitrary named references to cutscene data
- Better integration with zcamedit, as the same empty can (and should) be both a fast64 cutscene and a zcamedit cutscene

"Embedded" cutscenes (cutscene data is a property of the scene header) is supported as a deprecated feature, as I didn't want to break all of my existing scenes which have cutscene data in this format.

Most of the line changes in this PR are moving code to new files to resolve circular import issues. As is the import situation in the fast64 codebase was very fragile; adding a new file and including reasonably needed other files often led to circular imports. So I moved classes and other generic code out of the files which used them, so the classes could be included by other files. Generally, the solution is to not `from .foo import *` except for things like utilities which don't depend on anything else. However, fixing this would require a lot more changes.


sauraen:
Also Python choked on inconsistent tab/spaces in my changes, so I converted everything to tabs as that was by far the more common indentation used throughout the codebase. That change is responsible for a lot of the changed lines.

Zelllll (approved):
Awesome I'm glad you took my advice with how to set this up.

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+
+	collectionType : bpy.props.StringProperty()
+	textboxType : bpy.props.EnumProperty(items = ootEnumCSTextboxType)
+	listIndex : bpy.props.IntProperty()
+	objName : bpy.props.StringProperty()
+
+	def execute(self, context):
+		collection = getCollection(self.objName, self.collectionType, self.listIndex)
+		collection.add()
+		collection[len(collection)-1].textboxType = self.textboxType
```
Dragorn421, line 80:
```suggestion
		newTextboxElement = collection.add()
		newTextboxElement.textboxType = self.textboxType
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+		#self.report({'INFO'}, 'Success!')
```
Dragorn421, line 81: there are a bunch of commented out lines starting from this line, I think most of them can go?

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+		attrName = "bgm"
+	elif listProp.listType == "Misc":
+		attrName = "misc"
+	elif listProp.listType == "0x09":
+		attrName = "nine"
+	elif listProp.listType == "Unk":
+		prop_split(box, listProp, 'unkType', 'Unk List Type')
+		attrName = "unk"
+	else:
+		raise PluginError("Invalid listType")
```
Dragorn421, line 238:
```suggestion
		raise PluginError("Invalid listType: " + listProp.listType)
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+	bl_options = {'REGISTER', 'UNDO'} 
+
+	collectionType : bpy.props.StringProperty()
+	listType : bpy.props.EnumProperty(items = ootEnumCSListType)
+	objName : bpy.props.StringProperty()
+
+	def execute(self, context):
+		collection = getCollection(self.objName, self.collectionType, None)
+		collection.add()
+		collection[len(collection)-1].listType = self.listType
```
Dragorn421, line 273: same as previously

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+		#self.report({'INFO'}, 'Success!')
+		return {'FINISHED'} 
+
+def drawCSAddButtons(layout, objName, collectionType):
+	def addButton(row):
+		nonlocal l
```
Dragorn421, line 279:
```suggestion
	def addButton(row, l):
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+			addButton(row)
+	box.label(text = 'Install zcamedit for camera/actor motion.')
+
+
+class OOTCutsceneProperty(bpy.types.PropertyGroup):
+	csEndFrame : bpy.props.IntProperty(name = "End Frame", min = 0, default = 100)
+	csWriteTerminator : bpy.props.BoolProperty(name = "Write Terminator (Code Execution)")
+	csTermIdx : bpy.props.IntProperty(name = "Index", min = 0)
+	csTermStart : bpy.props.IntProperty(name = "Start Frm", min = 0, default = 99)
+	csTermEnd : bpy.props.IntProperty(name = "End Frm", min = 0, default = 100)
```
Dragorn421, line 302: What is "Frm" ?

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+	csLists : bpy.props.CollectionProperty(type = OOTCSListProperty, name = 'Cutscene Lists')
+
+def drawCutsceneProperty(box, obj):
+	prop = obj.ootCutsceneProperty
+	box.prop(prop, "csEndFrame", text = "End Frame")
```
Dragorn421, line 307: the property name automatically gets used as text, no need to duplicate the name here (same with others)
```suggestion
	box.prop(prop, "csEndFrame")
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+				entryOut.startFrame = entryIn.startFrame
+				listOut.entries.append(entryOut)
+		elif listOut.listType == 'Time':
+			for entryIn in listIn.time:
+				entryOut = OOTCSTime()
+				entryOut.startFrame = entryIn.startFrame
+				entryOut.hour = entryIn.hour
+				entryOut.minute = entryIn.minute
+				listOut.entries.append(entryOut)
+		elif listOut.listType in ['PlayBGM', 'StopBGM', 'FadeBGM']:
```
Dragorn421, line 356:
```suggestion
		elif listOut.listType in {'PlayBGM', 'StopBGM', 'FadeBGM'}:
```
speeeeed

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+
+
+################################################################################
+# Level classes to C
+################################################################################
+
+def ootCutsceneIncludes(headerfilename):
+	ret = CData()
+	ret.source = "#include \"ultra64.h\"\n" + \
+		"#include \"z64.h\"\n" + \
```
Dragorn421, line 416:
```suggestion
	ret.source = (
		'#include "ultra64.h"\n'
		+ '#include "z64.h"\n'
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+			elif list.listType == "0x09":
+				data.source += "0, " + str(e.startFrame) + ", " + str(e.startFrame + 1) + ", " \
+					+ e.unk2 + ", " + e.unk3 + ", " + e.unk4 + ", 0, 0"
+			elif list.listType == "Unk":
+				data.source += e.unk1 + ", " + e.unk2 + ", " + e.unk3 + ", " \
+					+ e.unk4 + ", " + e.unk5 + ", " + e.unk6 + ", " \
+					+ e.unk7 + ", " + e.unk8 + ", " + e.unk9 + ", " \
+					+ e.unk10 + ", " + e.unk11 + ", " + e.unk12
+			else:
+				raise PluginError("Invalid cutscene list type")
```
Dragorn421, line 482:
```suggestion
				raise PluginError("Invalid cutscene list type: " + list.listType)
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+################################################################################
+# Operators and panel
+################################################################################
+
+def checkGetFilePaths(context):
+	cpath = bpy.path.abspath(context.scene.ootCutsceneExportPath)
+	if not cpath.endswith('.c'):
+		raise PluginError('Output file must end with .c')
+	hpath = cpath[:-1] + 'h'
+	headerfilename = hpath.split('/')[-1]
```
Dragorn421, line 498:
```suggestion
	headerfilename = os.path.basename(hpath)
```

fast64_internal/oot/oot_cutscene.py
```diff
@@ -0,0 +1,604 @@
...
+	bl_space_type = 'VIEW_3D'
+	bl_region_type = 'UI'
+	bl_category = 'OOT'
+	
+	@classmethod
+	def poll(cls, context):
+		return True
+
+	# called every frame
+	def draw(self, context):
```
Dragorn421, line 569:
```suggestion
	def draw(self, context):
```
that comment is wrong

fast64_internal/oot/oot_operators.py
```diff
@@ -130,13 +130,35 @@ def execute(self, context):
...
-			parentObject(bpy.context.scene.ootSceneExportObj, roomObj)
+			parentObject(sceneObj, roomObj)
 		
 		bpy.ops.object.select_all(action = "DESELECT")
 		roomObj.select_set(True)
 		context.view_layer.objects.active = roomObj
 		return {"FINISHED"}
 
+class OOT_AddCutscene(bpy.types.Operator):
+	# set bl_ properties
```
Dragorn421, line 141:
```suggestion
```

fast64_internal/oot/oot_scene_room.py
```diff
@@ -593,6 +310,9 @@ class OOTSceneHeaderProperty(bpy.types.PropertyGroup):
 	exitList : bpy.props.CollectionProperty(type = OOTExitProperty, name = "Exit List")
 
 	writeCutscene : bpy.props.BoolProperty(name = "Write Cutscene")
+	csWriteType : bpy.props.EnumProperty(name = "Cutscene Data Type", items = ootEnumCSWriteType, default = "Embedded")
+	csWriteCustom : bpy.props.StringProperty(name = "CS hdr var:", default = "")
+	csWriteObject : bpy.props.PointerProperty(name = "Cutscene Object", type = bpy.types.Object)
 	csEndFrame : bpy.props.IntProperty(name = "End Frame", min = 0, default = 100)
```
Dragorn421, line 316: Isn't a lot (all?) of the stuff here duplicate of what `oot_cutscene.py` contains?

Dragorn421 (changes requested):

fast64_internal/oot/oot_cutscene.py
sauraen: Changed

fast64_internal/oot/oot_cutscene.py
sauraen: Changed. The commented out lines below that were for unknown or unused properties. These are properties that are in the cutscene data, and sometimes set to nonzero values that seem to have meaning, but they have been confirmed by decomp (at least in OoT) to never be read by the game. I left them as commented out because if you wanted to create arbitrary *cutscene data*--not merely create a specific behavior in a cutscene in the game--these could be uncommented and used. However, I didn't actually have commented-out support for them through the exporting process, and I doubt anyone is going to care what's in the unused data, so I just removed them.

fast64_internal/oot/oot_cutscene.py
sauraen: Changed

fast64_internal/oot/oot_cutscene.py
sauraen: Changed

fast64_internal/oot/oot_cutscene.py
sauraen: The change as suggested is incorrect, `l` is modified by the function. I could take `l` as a parameter and also return a modified value for it, but this is also awkward. I wouldn't use `nonlocal` for an "API-level" function, but since this is just for a function within another function, I think this is the most elegant way to do it.

fast64_internal/oot/oot_cutscene.py
sauraen: "Frame". This has to fit within the background of one of those value scroll boxes, next to the value itself, so the characters are very limited.

fast64_internal/oot/oot_cutscene.py
sauraen: You are correct and I have changed it for these. However, there are many places throughout the codebase where the name is specified in both places, often with different names.

fast64_internal/oot/oot_cutscene.py
sauraen: That is extremely pedantic but I suppose strictly speaking superior. Changed.

fast64_internal/oot/oot_cutscene.py
Dragorn421: Then what's the difference between "end frame" and "end frm" 🤔

fast64_internal/oot/oot_cutscene.py
Dragorn421: It's 3 times faster! Speeeed

fast64_internal/oot/oot_cutscene.py
sauraen: Changed, also changed to single quoted strings so the double quotes could be used without escape sequences. Tried a multiline string approach but there was no good way to get it to look good and also not put a bunch of tabs/spaces in the output.

fast64_internal/oot/oot_cutscene.py
sauraen: Changed

fast64_internal/oot/oot_cutscene.py
sauraen: Changed

fast64_internal/oot/oot_cutscene.py
sauraen: You are correct, it is called when there are mouse movements over the panel area. I think I copied that from some Panel tutorial. (This is code which was already in fast64, just moved to this file, by the way.)

fast64_internal/oot/oot_operators.py
sauraen: I included this as it appeared in other Operator classes in fast64 and it seemed to be the project's standard practice. Removed.

fast64_internal/oot/oot_scene_room.py
sauraen: Yes, that is intentional. This is the deprecated "embedded" cutscene system, where the cutscene data is stored in the scene header itself. Removing this would break any cutscenes which anyone (me!) had already made with that system. However, please note that it's just these 8 properties which are duplicated. Everything within `OOTCSListProperty`, including all the GUI for editing this property, is shared between the "embedded" system and the new first-class cutscene object system.

fast64_internal/oot/oot_cutscene.py
sauraen: `csEndFrame` is the end frame of the cutscene. `csTermStart` and `csTermEnd` are the start and end frames of the terminator. This is clear from the GUI (at least I think it is, if you think there should be an extra box or something around them then I can add that).
Also the `csEndFrame` value scroll box extends across the whole panel, whereas `csTermIdx`, `csTermStart`, and `csTermEnd` share a row. That's why there is enough room to fit more text for "End Frame".

sauraen: Screenshot_20210910_234609

Zelllll: The terminator shouldn't actually need to be more than one frame to begin with

sauraen: @Zelllll For a normal terminator which loads a new scene, that is correct. However:
1. The game checks both start and end frames, so end frame is not unused. fast64 + zcamedit together are supposed to be able to edit any cutscene data besides unused data.
2. Some vanilla terminators contain custom code. Someone could make a custom terminator which, for example, checks for some condition in some actor doing something in the cutscene before triggering the transition. Then, they would want that terminator to run for multiple frames.

fast64_internal/oot/oot_scene_room.py
Dragorn421: Could you make it obvious what is legacy code (kept for compatibility) by adding comments wherever needed

Dragorn421: Then if you can confirm the addon isn't broken with master merged (since you moved stuff around, but there hasn't been activity on the oot side recently so it should be fine) we can merge this

Dragorn421:
Just in case you missed my comments above, I would appreciate it if you highlighted the "legacy" code (required for not breaking compatibility), to make maintaining the code easier in the future

Then if you can confirm your changes work with main merged we can merge this

sauraen:
I've merged main and added comments to the deprecated stuff. Scene and cutscene editing and exporting works correctly. I have not tried the other parts of the plugin (e.g. skeletons, SM64 stuff...)

Dragorn421 (approved):
Thanks! Surely it will be fine ☺️

PR #19, opened 2021-09-07
Added ability to ignore blocks of code in level scripts

thecozies:
## SM64 Level Exporting
this change adds various areas in a levelscript where you can add any sort of code that will stay in that spot

```c
/* FAST64KEEP INCLUDES */
put whatever you want in here
/* FAST64END INCLUDES */
```

This makes it so that anything thats not covered by Fast64 (or just custom in general) doesn't get overwritten when exporting levels. A more generic solution could be cool to eventually implement... (e.g. keep anything between `/* FAST64KEEP BLOCK */` to `/* FAST64END BLOCK */`, but a lot of conflicts could come out of this, so I think its better to be specific for now)

### DIFF
--- a/fast64_internal/sm64/sm64_level_writer.py
+++ b/fast64_internal/sm64/sm64_level_writer.py
@@ -482,7 +482,7 @@ def parseLevelPersistentBlocks(scriptData: str, levelScript: LevelScript):
 			if areaIndex:
 				levelScript.persistentBlocks[curBlock][areaIndex].append(line)
 			else:
-				raise PluginError(f'script.c: "{PersistentBlocks.beginMagic} [area commands]" found outside of area block')
+				raise PluginError(f'script.c: "{PersistentBlocks.beginMagic} {PersistentBlocks.areaCommands}" found outside of area block')
 
 		elif curBlock:
 			levelScript.persistentBlocks[curBlock].append(line)
--- a/fast64_internal/sm64/sm64_level_writer.py
+++ b/fast64_internal/sm64/sm64_level_writer.py
@@ -217,13 +217,13 @@ class LevelDefines:
 		self.defineMacros.append(macroCmd)
 		return macroCmd
 
-FAST64KEEP = 'FAST64KEEP'
-FAST64END = 'FAST64END'
-FAST64_KEEPS = {
-	'INCLUDES': 'INCLUDES',
-	'SCRIPTS': 'SCRIPTS',
-	'LEVEL_COMMANDS': 'LEVEL_COMMANDS',
-	'AREA_COMMANDS': 'AREA_COMMANDS',
+beginPersistentBlockMagic = 'Fast64 begin persistent block'
+endPersistentBlockMagic = 'Fast64 end persistent block'
+persistentBlocksNames = {
+	'[includes]': '[includes]',
+	'[scripts]': '[scripts]',
+	'[level commands]': '[level commands]',
+	'[area commands]': '[area commands]',
 }
 
 class LevelScript:
@@ -235,7 +235,12 @@ class LevelScript:
 		self.modelLoads = []
 		self.actorIncludes = []
 		self.marioStart = None
-		self.keeps = dict(INCLUDES=[], SCRIPTS=[], LEVEL_COMMANDS=[], AREA_COMMANDS=defaultdict(list))
+		self.persistentBlocks = {
+			'[includes]': [],
+			'[scripts]': [],
+			'[level commands]': [],
+			'[area commands]': defaultdict(list)
+		}
 
 	def to_c(self, areaString):
 		result = '#include <ultra64.h>\n' +\
@@ -252,12 +257,12 @@ class LevelScript:
 		for actorInclude in self.actorIncludes:
 			result += actorInclude + '\n'
 		
-		result += f'\n{self.get_keeps(FAST64_KEEPS["INCLUDES"])}\n\n'
+		result += f'\n{self.get_persistent_block(persistentBlocksNames["[includes]"])}\n\n'
 
 		result += '#include "make_const_nonconst.h"\n'
 		result += '#include "levels/' + self.name + '/header.h"\n\n'
 
-		result += f'{self.get_keeps(FAST64_KEEPS["SCRIPTS"])}\n\n'
+		result += f'{self.get_persistent_block(persistentBlocksNames["[scripts]"])}\n\n'
 
 		result += 'const LevelScript level_' + self.name + '_entry[] = {\n'
 		result += '\tINIT_LEVEL(),\n'
@@ -271,7 +276,7 @@ class LevelScript:
 			result += '\t' + macroToString(modelLoad, True) + '\n'
 		result += '\n'
 
-		result += f'{self.get_keeps(FAST64_KEEPS["LEVEL_COMMANDS"], nTabs=1)}\n\n'
+		result += f'{self.get_persistent_block(persistentBlocksNames["[level commands]"], nTabs=1)}\n\n'
 
 		result += areaString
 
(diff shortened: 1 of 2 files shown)


Zelllll (approved):
Good idea! I think this would also be useful for oot if you want to include a specific header.

Dragorn421 (commented):
Good job me not submitting my review, it has been waiting in limbo since yesterday

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -421,6 +448,40 @@ def removeActSelectorIgnore(basePath, levelEnum):
...
 		saveDataToFile(filepath, newData)
 
+def parseLevelFast64Ignores(scriptData: str, levelScript: LevelScript):
+	keep: str = None
+	areaIndex: str = None
+
+	i = 0
+	for line in scriptData.splitlines():
+		print(line, i)
+		i += 1
```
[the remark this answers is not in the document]
thecozies: i dont think i need that print either lol that was a debug slip up

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -217,6 +217,15 @@ def getOrMakeMacroByLevelName(self, levelName):
...
 		return macroCmd
 
+FAST64KEEP = 'FAST64KEEP'
+FAST64END = 'FAST64END'
+FAST64_KEEPS = {
+	'INCLUDES': 'INCLUDES',
+	'SCRIPTS': 'SCRIPTS',
+	'LEVEL_COMMANDS': 'LEVEL_COMMANDS',
+	'AREA_COMMANDS': 'AREA_COMMANDS',
+}
```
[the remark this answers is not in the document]
thecozies: > it's open to debate

nah i like all the suggestions here. the only thing is that the `persistentBlocksNames` keys/values should be the same so everything lines up.

sidenote, i really wanted to use `TypedDict` but its only standard in python 3.8. i considered installing a package for it but idk if its worth it.

thecozies:
Did all the renamings and regex optimizations!
See changes here: https://github.com/Fast-64/fast64/pull/19/commits/f140d1423161be585e4ec9e9780f4dc927881608

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -448,38 +453,44 @@ def removeActSelectorIgnore(basePath, levelEnum):
 	if data != newData:
 		saveDataToFile(filepath, newData)
 
-def parseLevelFast64Ignores(scriptData: str, levelScript: LevelScript):
-	keep: str = None
+areaNumReg = re.compile(r'.*AREA\(([0-9]+),.+\),')
+persistentBlockReg = re.compile(r'.*Fast64 begin persistent block (\[[\w ]+\]).*')
+persistentBlockEndReg = re.compile(r'.*Fast64 end persistent block (\[[\w ]+\]).*')
```
Dragorn421, line 458:
```suggestion
persistentBlockBeginReg = re.compile(r'.*' + beginPersistentBlockMagic + ' (\[[\w ]+\]).*')
persistentBlockEndReg = re.compile(r'.*' + endPersistentBlockMagic + ' (\[[\w ]+\]).*')
```
the point of defining a magic variable for a value is to not have it hardcoded in several places 😅

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -448,38 +453,44 @@ def removeActSelectorIgnore(basePath, levelEnum):
...
+			areaIndex = areaIndexMatch.group(1)
 
-		elif keep and keep == FAST64_KEEPS['AREA_COMMANDS']:
+		elif curBlock and curBlock == persistentBlocksNames['[area commands]']:
 			if areaIndex:
-				levelScript.keeps[keep][areaIndex].append(line)
+				levelScript.persistentBlocks[curBlock][areaIndex].append(line)
 			else:
-				raise PluginError('FAST64KEEP AREA_COMMANDS found outside of area block')
+				raise PluginError('script.c: "Fast64 begin persistent block [area commands]" found outside of area block')
```
Dragorn421, line 480:
```suggestion
				raise PluginError(f'script.c: "{beginPersistentBlockMagic} {persistentBlocksNames["[area commands]"]}" found outside of area block')
```

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -448,38 +453,44 @@ def removeActSelectorIgnore(basePath, levelEnum):
...
+			blockNameMatch = persistentBlockReg.match(line)
+			if curBlock and blockNameMatch:
+				raise PluginError(f'script.c: Found new persistent block ({blockNameMatch.group(1)}) while looking for an end block for {curBlock}')
+			elif blockNameMatch:
+				curBlock = blockNameMatch.group(1)
 	
-	if keep:
-		raise PluginError(f'"FAST64KEEP {keep}" never found a "FAST64END"')
+	if curBlock:
+		raise PluginError(f'script.c: "Fast64 begin persistent block {curBlock}" never found a "Fast64 end persistent block"')
```
Dragorn421, line 493: here too

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -217,13 +217,13 @@ def getOrMakeMacroByLevelName(self, levelName):
...
-	'SCRIPTS': 'SCRIPTS',
-	'LEVEL_COMMANDS': 'LEVEL_COMMANDS',
-	'AREA_COMMANDS': 'AREA_COMMANDS',
+beginPersistentBlockMagic = 'Fast64 begin persistent block'
+endPersistentBlockMagic = 'Fast64 end persistent block'
+persistentBlocksNames = {
+	'[includes]': '[includes]',
+	'[scripts]': '[scripts]',
+	'[level commands]': '[level commands]',
+	'[area commands]': '[area commands]',
```
Dragorn421, line 226: I don't understand why this needs to have key==value, if it really does then there's no point to this dictionary, right?

If it doesn't need key==value having less magic keys would be better
(or maybe don't use a dict for this, eg `includesPersistentBlockName = "[includes]"`)

fast64_internal/sm64/sm64_level_writer.py
thecozies: 😔 

yeahhhhhhh makes sense lol changing it

fast64_internal/sm64/sm64_level_writer.py
thecozies: yeah i'm fine with that, i'm mostly just doing it this way because i wanted to group the blocks together under the same structure, i really actually dont need the keys to equal the values. what do you think about a class for these constants? that way these constants can be accessed from `PersistentBlocks.includes`

thecozies:
```python
class PersistentBlocks:
	includes = '[includes]'
	scripts = '[scripts]'
	levelCommands = '[level commands]'
	areaCommands = '[area commands]'

	@classmethod
	def new(cls) -> dict:
		return {
			[cls.includes]: [],
			[cls.scripts]: [],
			[cls.levelCommands]: [],
			[cls.areaCommands]: defaultdict(list)
		}
```

what about something like this? then the LevelScript class can do `self.persistentBlocks = PersistentBlocks.new()`

thecozies: i could even keep the magic strings or regexes in here too, so that the data is all together under a class. but maybe thats overkill lol

Dragorn421: doesn't look very python-y (more java-y) but I like it

thecozies: yeah it definitely isn't the most python-y implementation, but having loose constants can get hard to manage so i do like to get things like this grouped together

thecozies: okay, updated again!

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -421,6 +460,44 @@ def removeActSelectorIgnore(basePath, levelEnum):
...
+		
+		areaIndexMatch = areaNumReg.match(line)
+		if areaIndexMatch:
+			areaIndex = areaIndexMatch.group(1)
+
+		elif curBlock and curBlock == PersistentBlocks.areaCommands:
+			if areaIndex:
+				levelScript.persistentBlocks[curBlock][areaIndex].append(line)
+			else:
+				raise PluginError(f'script.c: "{PersistentBlocks.beginMagic} [area commands]" found outside of area block')
```
Dragorn421, line 485:
```suggestion
				raise PluginError(f'script.c: "{PersistentBlocks.beginMagic} {PersistentBlocks.areaCommands}" found outside of area block')
```

Dragorn421 (commented):
This looks much cleaner with `PersistentBlocks`, nice

fast64_internal/sm64/sm64_level_writer.py
thecozies: good catch!!

Dragorn421 (approved):

PR #20, opened 2021-09-08
SM64 check for rotation via isZeroRotation

thecozies:
before, extra geolayout nodes were added if rotations had an axis angle of `>0.00001`, resulting in useless rotate bones if the rotation would be computed to 0, 0, 0 anyways. so instead i do the same rotation check as object exports, and use the function `isZeroRotation`


Zelllll (approved):
This seems like a good solution to this problem 👍

Dragorn421 (approved):

PR #23, opened 2021-09-15
refactor and cleanup store_original_meshes

thecozies:
the logic is a lot more "positive" now. i check if something _should_ be instanced instead of shouldn't be instanced


Dragorn421 (approved):
Much better 😅

PR #32, opened 2021-10-08
Fixed skeleton regex

sauraen:
fast64 uses a regex to detect and parse a skeleton header definition from C. The existing regex matches only old decomp formatting which is missing braces from the struct definition. The new regex matches both the old style and the new style.


Zelllll (approved):
Commit history looks good now :)

PR #33, opened 2021-10-18
Add constraint model DL export

sauraen:
This PR adds support for the use case of creating meshes to be attached to another entity, usually to a skeletal mesh. This would be for creating things a character holds or wears, such as a sword, mask, or hat.

In Blender:
- Add a constraint to your mesh of type ChildOf, with the source being the appropriate bone of the skeleton of the base mesh. (Note that the ChildOf constraint is not the same as actually making the mesh a child object of the skeleton.)
- Your mesh should move with the bone as the skeleton is animated.
- Move the mesh to the correct position relative to the base mesh, while the constraint is active. For example, move the sword so it is visually in the character's hand.
- Simply export the mesh's DL with fast64's Export DL tool and compile it into the object file or actor file. The constraint **must be left enabled** while exporting.

In your custom actor:
- In `YourActor_PostLimbDraw`, add code like this:
```
if(limbIndex == YOURACTOR_LIMB_LHAND){
	gSPDisplayList(POLY_OPA_DISP++, YourMeshDL);
}
```
- Don't forget that limb indexes count the root position and root rotation as separate limbs. So, 0 is root position, 1 is root rotation, 2 is rotation of the next limb after root...

### DIFF
--- a/fast64_internal/oot/oot_utility.py
+++ b/fast64_internal/oot/oot_utility.py
@@ -237,7 +237,33 @@ def ootDuplicateHierarchy(obj, ignoreAttr, includeEmpties, objectCategorizer):
 						selectedObj.parent.select_set(True)
 						bpy.ops.object.parent_set(keep_transform = True)
 					selectedObj.parent = None
-		applyRotation([tempObj], math.radians(90), 'X')
+		
+		# Assume objects with these types of constraints are parented, and are
+		# intended to be parented in-game, i.e. rendered as an extra DL alongside
+		# a skeletal mesh, e.g. for a character to be wearing or holding it.
+		# In this case we purely want the transformation of the object relative
+		# to whatever it's parented to. Getting rid of the constraint and then
+		# doing transform_apply() sets up this transformation.
+		hasConstraint = False
+		for constraint in tempObj.constraints:
+			if constraint.type in ['COPY_LOCATION', 'COPY_ROTATION', 'COPY_SCALE', \
+				'COPY_TRANSFORMS', 'TRANSFORM', 'CHILD_OF', 'CLAMP_TO', 'DAMPED_TRACK', \
+				'LOCKED_TRACK', 'TRACK_TO'] and not constraint.mute:
+				hasConstraint = True
+				tempObj.constraints.remove(constraint)
+		if not hasConstraint:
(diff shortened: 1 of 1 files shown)


fast64_internal/oot/oot_utility.py
```diff
@@ -237,7 +237,33 @@ def ootDuplicateHierarchy(obj, ignoreAttr, includeEmpties, objectCategorizer):
...
+		# Assume objects with these types of constraints are parented, and are
+		# intended to be parented in-game, i.e. rendered as an extra DL alongside
+		# a skeletal mesh, e.g. for a character to be wearing or holding it.
+		# In this case we purely want the transformation of the object relative
+		# to whatever it's parented to. Getting rid of the constraint and then
+		# doing transform_apply() sets up this transformation.
+		hasConstraint = False
+		for constraint in tempObj.constraints:
+			if constraint.type in ['COPY_LOCATION', 'COPY_ROTATION', 'COPY_SCALE', \
+				'COPY_TRANSFORMS', 'TRANSFORM', 'CHILD_OF', 'CLAMP_TO', 'DAMPED_TRACK', \
```
Dragorn421, line 250: You don't need the \

Also, using a set is faster but no big deal

fast64_internal/oot/oot_utility.py
```diff
@@ -237,7 +237,33 @@ def ootDuplicateHierarchy(obj, ignoreAttr, includeEmpties, objectCategorizer):
...
+				'LOCKED_TRACK', 'TRACK_TO'] and not constraint.mute:
+				hasConstraint = True
+				tempObj.constraints.remove(constraint)
+		if not hasConstraint:
+			# For normal objects, the game's coordinate system is 90 degrees
+			# away from Blender's.
+			applyRotation([tempObj], math.radians(90), 'X')
+		else:
+			# This is a relative transform we care about so the 90 degrees
+			# doesn't matter (since they're both right-handed).
```
Dragorn421, line 260: I don't understand how the override-limb-draw-rigged mesh doesn't need the transform from blender coords to oot coords,
if skeletons get that transform then the mesh "attached" should too or its blender coords would not match the skeleton's (limbs') oot coords

I know "it works so it must be ok" but it's better to make sure everything doesn't work out of luck 😅

PR #37, opened 2021-10-24
Don't separate OoT limbs into normal and skinned DLs

sauraen:
`ootProcessVertexGroup` had code which separated the faces of a limb into one group for normal tris (all verts from the tri's vertex group), and separate groups for each different set of skinned tris based on which vertex groups were involved. Then, it would create separate DLs for each of these groups. This is superfluous for several reasons:
- It looks like it's a holdover from SM64 code which treats non-flex and flex/skinned portions differently. The sets of vertex groups which form these groups were not actually used in the OoT code. (For example, for limb 2, non-flex triangles use vertex groups `(2,)` and flex triangles use vertex groups `(1,2)`, but these actual tuples were never used after being created.)
- Often leads to vertices being loaded (and transformed by the RSP) twice.
- If there are multiple materials used by the non-flex tris and the same set of materials being used by flex tris, runs each of those material definitions twice, including loading the textures twice.
- Even in the best case, still produces a couple extra `gsSPDisplayList` and `gsSPEndDisplayList` commands.

### DIFF
--- a/fast64_internal/f3d/f3d_writer.py
+++ b/fast64_internal/f3d/f3d_writer.py
@@ -827,7 +827,7 @@ class TriangleConverter:
 
 	def addFace(self, face):
 		triIndices = []
-		existingVerts = []
+		#existingVerts = [] # set but never read
 		addedVerts = [] # verts added to existing vertexBuffer
 		allVerts = [] # all verts not in 'untouched' buffer region
 
@@ -841,10 +841,11 @@ class TriangleConverter:
 				addedVerts.append(bufferVert)
 
 			if bufferVert in self.vertBuffer[:self.bufferStart]:
-				existingVerts.append(bufferVert)
+				#existingVerts.append(bufferVert)
+				pass
 			else:
 				allVerts.append(bufferVert)
-				existingVerts.append(None)
+				#existingVerts.append(None)
 
 		# We care only about load size, since loading is what takes up time.
 		# Even if vert_buffer is larger, its still another load to fill it.
--- a/fast64_internal/f3d/f3d_writer.py
+++ b/fast64_internal/f3d/f3d_writer.py
@@ -827,7 +827,6 @@ class TriangleConverter:
 
 	def addFace(self, face):
 		triIndices = []
-		#existingVerts = [] # set but never read
 		addedVerts = [] # verts added to existing vertexBuffer
 		allVerts = [] # all verts not in 'untouched' buffer region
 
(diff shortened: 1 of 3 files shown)


fast64_internal/f3d/f3d_writer.py
```diff
@@ -827,7 +827,6 @@ def processGeometry(self):
 
 	def addFace(self, face):
 		triIndices = []
-		existingVerts = []
```
sauraen, line 830: This variable was set but not used.

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -45,41 +44,36 @@ def ootProcessVertexGroup(fModel, meshObj, vertexGroup, convertTransformMatrix,
...
 				continue
-			groupTuple = tuple(sorted(groupTuple))
-
-			if groupTuple == tuple([currentGroupIndex]):
-				if face.material_index not in groupFaces:
-					groupFaces[face.material_index] = []
-				groupFaces[face.material_index].append(face)
-			else:
-				if groupTuple not in skinnedFaces:
-					skinnedFaces[groupTuple] = {}
```
sauraen, line 74: The tuples were stored into skinnedFaces...

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -94,65 +88,30 @@ def ootProcessVertexGroup(fModel, meshObj, vertexGroup, convertTransformMatrix,
...
 		if fMaterial.useLargeTextures:
-			currentGroupIndex = saveMeshWithLargeTexturesByFaces(material, faces, fModel, fMeshes[drawLayer],
-				meshObj, drawLayer, convertTextureData, currentGroupIndex, triConverterInfo, None, None)
+			currentGroupIndex = saveMeshWithLargeTexturesByFaces(material, faces, fModel, fMesh,
+				meshObj, drawLayerOverride, convertTextureData, currentGroupIndex, triConverterInfo, None, None)
 		else:
-			currentGroupIndex = saveMeshByFaces(material, faces, fModel, fMeshes[drawLayer], 
-				meshObj, drawLayer, convertTextureData, currentGroupIndex, triConverterInfo, None, None)
-
-	for groupTuple, materialFaces in skinnedFaces.items():
```
sauraen, line 128: Here is where the tuples were not used.

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -94,65 +88,30 @@ def ootProcessVertexGroup(fModel, meshObj, vertexGroup, convertTransformMatrix,
 			return None, False
 	
 	meshInfo.vertexGroupInfo.vertexGroupToMatrixIndex[currentGroupIndex] = nextDLIndex
-	fMeshes = {}
```
sauraen, line 97: Since there is only one draw layer, we only need one fMesh, not an array.

Zelllll (approved):
I was not aware how unoptimized this was before, great job!

PR #39, opened 2021-10-24
Begin OoT skeleton optimizer

sauraen:
This begins a set of features for optimizing OoT skeletons. The optimizer can be turned on or off at skeleton export time; if it is off, the export process is like it has been in the past, and the assumptions below don't apply.

The optimizing features rely on maintaining state in the RSP and/or RDP between limbs of a skeleton. If optimization is enabled, drawing other DLs within the SkelAnime callbacks (OverrideLimbDraw or PostLimbDraw) will corrupt the graphics state. Limbs also can't be culled in software, though since there is no built-in culling mechanism, the user would have to write the culling themselves. (Actually, they can be culled, just very carefully.)

This PR provides one very basic optimization feature: not running the material setup DL when the material was already set up correctly by the last limb. In a basic case with N limbs all using the same material, this saves N-1 superfluous material executions, including N-1 superfluous texture loads (or 2N-2 superfluous loads if using CI textures). If a limb uses multiple materials but one of them is the last material, it will draw those tris first so it still saves one material setup.


sauraen:
Since this now has two approvals from people other than me (the original requester), I'm going to go ahead and merge it.

PR #42, opened 2021-11-06
Fix bug that made some materials not be possible to import

Zelllll:
This mainly affected Majora's Mask displaylists


Dragorn421 (approved):
thonk

EDIt: oh right that's the implementation for `gsSPLoadGeometryMode`. that command is rarely used so i understand why it slipped through

Zelllll:
Yeah that's correct, Nintendo updated their converter to use it a lot in mm though

PR #44, opened 2021-11-09
Make OoT scene exporter include externs for room rom segments

Zelllll:
This change is being made since eventually `segment_symbols.h` will ideally be removed from decomp, and it makes more sense to just extern these within the scene file.

### DIFF
--- a/fast64_internal/oot/c_writer/oot_level_c.py
+++ b/fast64_internal/oot/c_writer/oot_level_c.py
@@ -444,13 +444,24 @@ def ootTransitionActorListToC(scene, headerIndex):
 	data.source += '};\n\n'
 	return data
 
+def ootRoomExternToC(room):
+	return "extern u8 " + room.roomName() + "SegmentRomStart[];\n" + \
+		"extern u8 " + room.roomName() + "SegmentRomEnd[];\n"
+
 def ootRoomListEntryToC(room):
 	return "{ (u32)_" + room.roomName() + "SegmentRomStart, (u32)_" + room.roomName() + "SegmentRomEnd },\n"
 
 def ootRoomListHeaderToC(scene):
 	data = CData()
-	data.header = "extern RomFile " + scene.roomListName() + "[];\n"
-	data.source = "RomFile " + scene.roomListName() + "[] = {\n"
+
+	# Write externs for rom segments
+	for i in range(len(scene.rooms)):
+		data.source += ootRoomExternToC(scene.rooms[i])
+	data.source += "\n"
+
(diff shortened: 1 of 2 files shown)


fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -444,13 +444,24 @@ def ootTransitionActorListToC(scene, headerIndex):
...
 
 def ootRoomListHeaderToC(scene):
 	data = CData()
-	data.header = "extern RomFile " + scene.roomListName() + "[];\n"
-	data.source = "RomFile " + scene.roomListName() + "[] = {\n"
+
+	# Write externs for rom segments
+	for i in range(len(scene.rooms)):
+		data.source += ootRoomExternToC(scene.rooms[i])
+	data.source += "\n"
```
Dragorn421, line 460: wouldn't this fit better in the header?

fast64_internal/oot/c_writer/oot_segment_symbols_c.py
```diff
@@ -62,8 +62,4 @@ def modifySegmentSymbols(scene, exportInfo):
 		segmentSymbols.insert(firstIndex, SegmentSymbolEntry("DECLARE_ROM_SEGMENT", scene.name + "_scene"))
 		firstIndex += 1
 
-		for i in range(len(scene.rooms)):
-			segmentSymbols.insert(firstIndex, SegmentSymbolEntry("DECLARE_ROM_SEGMENT", scene.name + "_room_" + str(i)))
-			firstIndex += 1
```
Dragorn421, line 67: what was this for and why can it be removed now?

Zelllll: It adds the segment define to segment_symbols.h, but now it is just extern'd in the c file.

fast64_internal/oot/c_writer/oot_level_c.py
Zelllll: Not really, it isn't needed by any other files

Dragorn421: I don't think the main point of the header is to have symbols for use by other files though,
for example the next line `data.header += "extern RomFile " + scene.roomListName() + "[];\n"` declares the room list in the header

Zelllll: Well, eventually I planned on making another pr that merges the headers for each room and the scene into one file, which in my mind would make sense as something for external access rather than internal access. So since these symbols are intended to be used "internally" I thought this was more fitting.

thecozies (approved):
looks good imo

PR #45, opened 2021-11-11
Make OoT scenes be split into multiple, modular files

Zelllll:
This PR splits OoT scene and rooms files into multiple files based on their "data type." The types are as follows:

- Scene Main File (includes scene header, transition actor list, exit list, etc.)
- Scene Collision File (includes scene collision)
- Scene Texture File (includes material setup displaylists and textures that are shared between multiple rooms)
- Scene Cutscene File(s) (includes cutscene data, one file for each cutscene)
- Room Main File (includes room header, actor lists, etc.)
- Room Model Info File (contains the list of displaylists in a room)
- Room Model File (contains the room model and textures)

This is designed to make it easier to both read and modify scene or room files generated by Fast64, as well as be more consistent with how the vanilla scenes were split up. This also aligns with how SM64 handles files (apparently). See here: https://github.com/n64decomp/sm64/tree/master/levels/bowser_3/areas/1

This feature is optional though and can be enabled with a checkbox. If disabled it will work as before. Also, this PR combines all the room headers into one single header.

### DIFF
--- a/fast64_internal/oot/c_writer/oot_level_c.py
+++ b/fast64_internal/oot/c_writer/oot_level_c.py
@@ -10,77 +10,65 @@ def cmdName(name, header, index):
 # Scene Commands
 def cmdSoundSettings(scene, header, cmdCount):
 	cmd = CData()
-	# cmd.header = "extern " + "SCmdSoundSettings " + cmdName(scene.sceneName(), header, cmdCount) + ';\n'
 	cmd.source = "\tSCENE_CMD_SOUND_SETTINGS(" + str(scene.audioSessionPreset) + ", " + str(scene.nightSeq) + \
 		", " + str(scene.musicSeq) + "),\n"
 	return cmd
 
 def cmdRoomList(scene, header, cmdCount):
 	cmd = CData()
-	# cmd.header = "extern " + "SCmdRoomList " + cmdName(scene.sceneName(), header, cmdCount) + ';\n'
 	cmd.source = "\tSCENE_CMD_ROOM_LIST(" + str(len(scene.rooms)) + ", &" + scene.roomListName() + "),\n"
 	return cmd
 
 def cmdTransiActorList(scene, header, cmdCount):
 	cmd = CData()
-	# cmd.header = "extern " + "SCmdTransiActorList " + cmdName(scene.sceneName(), header, cmdCount) + ';\n'
 	cmd.source = "\tSCENE_CMD_TRANSITION_ACTOR_LIST(" + str(len(scene.transitionActorList)) + ", &" + \
 		scene.transitionActorListName(header) + "),\n"
 	return cmd
 
 def cmdMiscSettings(scene, header, cmdCount):
 	cmd = CData()
-	# cmd.header = "extern " + "SCmdMiscSettings " + cmdName(scene.sceneName(), header, cmdCount) + ';\n'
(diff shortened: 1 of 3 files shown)


sauraen (approved):
I personally don't think that breaking up scenes and rooms into multiple files is useful, for several reasons:
- Clutters already-cluttered scene directories
- Makes the build process slightly harder, especially for those less experienced
- Modularity doesn't really make sense since one would rarely want to include only certain pieces of a scene or room or swap the pieces around between different scenes / rooms
- Doesn't match OoT decomp style

However, it's an option and it's in the normal, single-file mode by default, so no problem, hopefully this new mode helps someone.

thecozies (approved):

thecozies:
> I personally don't think that breaking up scenes and rooms into multiple files is useful, for several reasons:
> 
> * Clutters already-cluttered scene directories
> * Makes the build process slightly harder, especially for those less experienced
> * Modularity doesn't really make sense since one would rarely want to include only certain pieces of a scene or room or swap the pieces around between different scenes / rooms
> * Doesn't match OoT decomp style
> 
> However, it's an option and it's in the normal, single-file mode by default, so no problem, hopefully this new mode helps someone.

It's a good idea to keep things as split up as possible in Fast64 exports. Like we want Fast64 to have every feature that somebody could want but we have to expect that people will want to make manual changes to exports. And multiple files will keep that data much more organized and easier to find what you're looking for.

> Makes the build process slightly harder, especially for those less experienced

If it makes the build process harder, I could see that being frustrating. In SM64, Fast64 uses `inc.c` and `inc.h` files then includes those files into a single .c/.h file, so multiple files essentially behave as one. They aren't protected (you can't import from the individual broken apart files) but you can at least import the grouped file and still get the organizational benefits. Something to consider.

Zelllll:
Most people that I have shown this to have found it to be very useful actually, (CDi-Fails did and he uses Fast64 a lot), but that is why I made the feature optional, specifically for Sauraen lol

thecozies:
if it's useful for most people then I'd say make it checked by default

Zelllll:
In that case I'll make it the default, and Sauraen can turn it off

sauraen:
> And multiple files will keep that data much more organized and easier to find what you're looking for.
I do not agree. I can only see people manually editing things like actor and object lists, which are at the beginning of the single file. Nobody's going to manually edit collision data or vertex lists. And even if they edit everything, I'm not sure how breaking up a file into 3 parts would make editing it easier. If we were talking about something like an object file which could contain dozens of different unrelated DLs and associated textures etc., that would make more sense. But the individual files here cannot be used on their own, so we're really just adding two file breaks because it is believed that makes manual editing easier. Again, I'm fine with this being an option, but I see no reason the default behavior should change.

> Most people that I have shown this to have found it to be very useful actually
Zel has asked three people about this feature: me, thecozies, and CDi. Two said it was good and one said it was bad. It is a bit disingenuous to cast this as "most people... have found it to be very useful actually".

> that is why I made the feature optional, specifically for Sauraen lol
I do not need it myself, I can change my build toolchain to handle multiple files, because I know how to use make. I'm arguing against the change for the reasons I stated above, which I don't think have been addressed.

> In that case I'll make it the default, and Sauraen can turn it off
My approval was specifically because it was in normal, single-file mode by default.

Zelllll:
The issue isn't editing specific things necessarily (i.e. collision), it's making it so the scene file isn't cluttered with giant textures or displaylists etc

Zelllll:
Two approvals so I'm going to merge this

PR #47, opened 2021-11-13
Add support for multitextures where both texels fill tmem

Zelllll:
This pull request adds support for a "hack" that Majora's Mask does which allows you to have a texture that fills all of tmem (for example 32x64 rgba16) and load it into both Texel 0 and Texel 1. It is somewhat experimental but from the testing that I have done it seems to work. Essentially it just loads the texture into the same location in tmem for both texels.


Zelllll:
Since there hasn't been any comments on this I am going to merge it.

PR #48, opened 2021-11-23
Instant Warp offset objects

Bitlytic:
Let users select two objects to use for instant warp offset instead of requiring manual entry.

### DIFF
--- a/fast64_internal/sm64/sm64_objects.py
+++ b/fast64_internal/sm64/sm64_objects.py
@@ -1148,9 +1148,9 @@ class WarpNodeProperty(bpy.types.PropertyGroup):
 
 		# Convert from Blender space to SM64 space
 		ret = Vector()
-		ret.x = int(difference.x * bpy.context.scene.blenderF3DScale)
-		ret.y = int(difference.z * bpy.context.scene.blenderF3DScale)
-		ret.z = int(-difference.y * bpy.context.scene.blenderF3DScale)
+		ret.x = int(round(difference.x * bpy.context.scene.blenderF3DScale))
+		ret.y = int(round(difference.z * bpy.context.scene.blenderF3DScale))
+		ret.z = int(round(-difference.y * bpy.context.scene.blenderF3DScale))
 		return ret
 
 	def to_c(self):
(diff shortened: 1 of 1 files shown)


thecozies:
This is awesome! The one question I have when using it, is that I think the difference could be flipped.

So in this screenshot, I'm on Area 2 and I'm warping to Area1 and Area 3 respectively. When I put in my Area objects as the offset objects, the values were reversed from what they should be.

I should note that in my blend, my area roots are in different places. So maybe the offset is correct when instant warping from the same origin? Also, it looks like my second warp may have a rounding error.

image

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1135,14 +1135,38 @@ class WarpNodeProperty(bpy.types.PropertyGroup):
...
+	def calc_offsets_from_objects(self):
+		if self.instantWarpObject1 is None or self.instantWarpObject2 is None:
+			raise PluginError(f'Warp Start and Warp End in Warp Node {self.warpID} must have objects selected.')
+		difference = self.instantWarpObject2.location - self.instantWarpObject1.location
+
+		# Convert from Blender space to SM64 space
+		ret = Vector()
+		ret.x = int(difference.x * bpy.context.scene.blenderF3DScale)
+		ret.y = int(difference.z * bpy.context.scene.blenderF3DScale)
+		ret.z = int(-difference.y * bpy.context.scene.blenderF3DScale)
```
thecozies, line 1153: prevent rounding errors this way
```suggestion
		ret.x = int(round(difference.x * bpy.context.scene.blenderF3DScale))
		ret.y = int(round(difference.z * bpy.context.scene.blenderF3DScale))
		ret.z = int(round(-difference.y * bpy.context.scene.blenderF3DScale))
```

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1194,7 +1218,22 @@ def drawWarpNodeProperty(layout, warpNode, index):
...
+				writeBox = box.box()
+				if warpNode.instantWarpObject1 is None or warpNode.instantWarpObject2 is None:
+					writeBox.label(text='Both Objects must be selected for offset')
+				else:
+					writeBox.label(text='Current Offset: ')
+					difference = warpNode.calc_offsets_from_objects()
+					
+					writeBox.label(text='X: ' + str(int(difference.x)))
+					writeBox.label(text='Y: ' + str(int(difference.y)))
+					writeBox.label(text='Z: ' + str(int(difference.z)))
```
thecozies, line 1234: non blocking suggestion, this is just my personal preference to use f strings instead of concatenation, though you can skip the cast to `int` here since you're doing that in `calc_offsets_from_objects`

```suggestion
					writeBox.label(text=f"X: {difference.x}")
					writeBox.label(text=f"Y: {difference.y}")
					writeBox.label(text=f"Z: {difference.z}")
```

thecozies (approved):
##### well done

PR #51, opened 2021-12-04
Cross compatibility for Blender 3.0

thecozies:
Hey Zelda people can any of you test some OOT exports with this branch?


Zelllll (approved):
I've tested this with OoT and it seems to work great. Make sure @sauraen tests if zcamedit works though.

PR #52, opened 2021-12-05
Fix OoT skeleton export

CDi-Fails:
Fixes the "expected 3, got 2" error when exporting skeletons and fixes the "no such file: dmadata.s" error when exporting scenes. Since decomp now automatically generates dmadata from spec, I've just commented modifyDmaMgrFileNames and modifyDmaTableEntries. It may be a good idea to fully oot_dma_manager_c later on down the road.

### DIFF
--- a/fast64_internal/oot/oot_level_writer.py
+++ b/fast64_internal/oot/oot_level_writer.py
@@ -130,8 +130,8 @@ def writeOtherSceneProperties(scene, exportInfo):
 	modifySceneTable(scene, exportInfo)
 	modifySegmentSymbols(scene, exportInfo)
 	modifySceneIDs(scene, exportInfo)
-	# modifyDmaMgrFileNames(scene, exportInfo)
-	# modifyDmaTableEntries(scene, exportInfo)
+	modifyDmaMgrFileNames(scene, exportInfo)
+	modifyDmaTableEntries(scene, exportInfo)
 	modifySegmentDefinition(scene, exportInfo)
 	modifySceneFiles(scene, exportInfo)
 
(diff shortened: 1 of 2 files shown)


thecozies (commented):
looks file but did you mean to have modifyDmaMgrFileNames commented out?

CDi-Fails:
> looks file but did you mean to have modifyDmaMgrFileNames commented out?

Yes, sDmaMgrFileNames[] in z_std_dma.c just includes an auto-generated file full of filenames now, so there's no need manually edit anything there.

fast64_internal/oot/oot_level_writer.py
```diff
@@ -130,8 +130,8 @@ def writeOtherSceneProperties(scene, exportInfo):
 	modifySceneTable(scene, exportInfo)
 	modifySegmentSymbols(scene, exportInfo)
 	modifySceneIDs(scene, exportInfo)
-	modifyDmaMgrFileNames(scene, exportInfo)
-	modifyDmaTableEntries(scene, exportInfo)
+	# modifyDmaMgrFileNames(scene, exportInfo)
```
Zelllll, line 133: I'd say either remove this change or actually properly remove this feature rather than just commenting it out. I can make a pr later to do that so for now I'd say just revert these changes.

Zelllll (changes requested):

fast64_internal/oot/oot_level_writer.py
CDi-Fails: Sure.

Dragorn421 (approved):

Zelllll (approved):
Looks good now

PR #54, opened 2021-12-09
[OoT] addAtStartPosIndex -> addStartPosAtIndex

Yanis42:
From what I understand this is a typo (the parameters match addStartPosAtIndex), randomly found this while looking at the OoT exporter

### DIFF
--- a/fast64_internal/oot/oot_level_classes.py
+++ b/fast64_internal/oot/oot_level_classes.py
@@ -457,7 +457,7 @@ def addStartPosition(scene, index, actor, actorProp, actorObjName):
 		for cutsceneHeader in sceneSetup.cutsceneHeaders:
 			if cutsceneHeader.headerIndex >= len(scene.cutsceneHeaders) + 4:
 				raise PluginError(actorObjName + " uses a cutscene header index that is outside the range of the current number of cutscene headers.")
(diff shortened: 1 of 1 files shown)


Zelllll:
I think this might have been related to a bug that existed a long time ago where start positions would be exported incorrectly, what does that function do exactly?

Yanis42:
I don't really know what this is for, I randomly saw that typo while checking the purpose of  ``actorProp`` in ``addStartPosition`` (I'm working on the exporter for the detailed parameter panel thing I'm doing)

Apparently it fills a dictionnary of start positions, I'm not sure why though, I haven't focused a lot of what Fast64 actually does outside drawing things

Zelllll:
Yeah that makes sense. I'm kind of confused why it appeared to work with the typo existing though.

Zelllll (approved):

Yanis42:
I guess no one ever tried to export a scene with the scene setup preset set to custom, but yeah I'm really surprised Blender never complained about that

Zelllll:
Should we close this since you added it to your other PR?

Yanis42:
I'll close this yeah, it's not an important change anyway so this can wait until the #56 is merged

PR #55, opened 2021-12-09
[OoT] Delete the DMA data manager as it's no longer needed

Yanis42:
The exporter broke because of a recent change in the OoT Decomp: the dmadata stuff is generated from the spec file if the rom is non-matching (as defined in the makefile or by doing NON_MATCHING=1 with the make command), so Fast64 just need to edit the spec file (building the rom will update the dma table)


Yanis42:
I just noticed Zel did the same thing in #53, closing this

PR #56, opened 2021-12-10
[OoT] Detailed Actor Panel (Actors, Transition Actors, Entrance Actors)

Yanis42:
### List of changes/features:
* You can use that as a calculator to get an Actor's parameters
* _**It's beginner friendly**_, you just have to put whatever you want and it'll to the maths for you
* The rotation override thing is gone from the panel, it's not really useful to keep because...
* ... only rotations that are tied to a parameter overrides Blender's rotations, for instance, EN_BOX's Switch Flag override Blender's Y Rotation (which is the Z-Rot in OoT) but X and Y Rots will not be overridden, I made it that way because I thought it made more sense, also this changes is why I got rid of the Override rotation display thing (note that it's like it was before this PR only for custom actors)
* I'm generating the actor list list from the data of the XML file
* I changed a bit the entrance display so you can see the Player detailed properties, same thing for transition actors
* In the XML you'll find ``<List>`` nodes, it's used to avoid hardcoding the lists in ``oot_constants.py``
* I made some changes to the exporter and the functions that create a scene, room, ...
* Data from older blend files should update to the new data format when Fast64 loads, it won't show the actor panel UI until it's properly updated (older data should never be lost)
* Added a button that adds an actor in the OoT Tools box
* Added a button that adds an entrance in the OoT Tools box
* Raise a Plugin Error if the user sets X and Y rotations for transition actors
* Fixed a typo (#54)
* The XML gives a lot of flexibility, you can add nodes to existing actors (or even add a new actor) and it'll generate the options on the UI automatically (if you remove the added options in the XML file everything *should* still be saved in the blend, so if you add it back later you're not going to lose any data), the imaged at the end of this comment will show you what you can do with the UI

XML nodes explanations:
* ``<Enum>`` is meant to be used to generate an ``EnumProperty`` on the UI
* ``<Property>`` is meant to be used to generate a textbox
* ``<Flag>`` does the same as ``<Property>`` but it tells you it's a flag (Switch/Collectible/Chest)
* ``<Collectible>`` adds an enum with Item00 drop types (if you need a custom enum you can use the ``<Enum>`` node instead, see ``ACTOR_EN_WONDER_ITEM``)
* ``<Bool>`` draws a checkbox
* ``<Parameter>`` draws an enum with the actor types (the value in Params is already shifted, if there's no Mask it'll consider it to be ``0xFFFF``)
* ``<Message>`` draws the navi message id searchbox
* ``<ChestContent>`` draws the chest content searchbox
* ``<List>`` is not meant to be used to generate lists (at least for now), use ``<Enum>`` instead

And I think that's everything, I may have forgot some changes, in that case I apologize I should have noted every change I made.

__Also you should make a backup of your existing blend before updating just in case something goes wrong__ (it shoud work properly, you'll just have to set every actor's value again)

Note that I got random crashes; it's rare but it happened once or twice during the 2 months of developpment (the last one happened when I added a scene, which I could do 20 times in a row without crashing)

And I think that's all I have to say on this feature, it was really instructive as a person that doesn't do a lot of Python per year, if you have any questions, ask them!

(Shoutouts again to Dragorn that helped me getting started on this feature!)

### Adding options to an actor:
<details closed>
<summary>Default</summary>

</details>
<details closed>
<summary>Modified</summary>

</details>

### Adding a new actor
<details closed>
<summary>Custom actor</summary>

</details>

### Actor list searchbox
<details closed>
<summary>Readable names</summary>

</details>


Yanis42:
PR #54 change are included in this PR

Zelllll:
My first question before I review this more thoroughly is how does this react to scenes that are already made? Will params that are already set up be detected with this new system and be properly filled out?

Zelllll:
> Nothing is hardcoded outside things that should never be changed like the entrance actor ID, editing the XML is enough to add or remove props on the panel

This is not exactly true, because you may have an alternate player actor that you want to spawn instead (i.e. something kind of like Kafei).

> The rotation override thing is gone from the panel

I don't really approve of this change because it was there for custom actors as well. I don't really think people should have to worry about editing the XML when they have their own custom actors. This is also going to break people's existing scenes, which is something that really needs to be avoided.

> I got rid of transition actors in the actor list, instead if you wanna add a door you'll have to add a transition actor and you'll have the list

I can't say this change is good either. You could always edit an existing actor to become a transition actor.

> The other advantage with this method is that if OoT decomp people decides to change names you'll just have to make the changes in the XML. To avoid data loss because of decomp updates I decided to use Actor's hexadecimal IDs (i.e. 0009 for chests) to save the data in the .blend file.

I think you should elaborate on this a bit. The exported data from Fast64 MUST use the enums. If anyone ever shifts the actor table the IDs will all change. Furthermore, Fast64 needs to be updated anyways whenever the enum changes in decomp, so this change doesn't really add anything of value IMO, since the enum being exported is very important.

> Also you should make a backup of your existing blend before updating just in case something goes wrong (it shoud work properly, you'll just have to set every actor's value again)

This really needs solved before this can be merged. Having to redo the actors of an entire dungeon etc. isn't something that we would want users to have to do.

fast64_internal/oot/oot_constants.py
```diff
@@ -1494,3 +1070,183 @@
 	('12', 'To Dim Unk', 'Fade gSaveContext.unk_1410 255>100, works with scene xn 11/17'),
 	('13', 'From Dim', 'Alpha 100>255')
 ]
+
+ootEnWonderItemDrop = [
```
Zelllll, line 1074: Is there not some way to make all of this be handled within the XML rather than hardcoded within the plugin itself?

fast64_internal/oot/oot_actor.py
```diff
@@ -1,187 +1,529 @@
...
-		adultNightRow.prop(headerProp, 'adultNightHeader', text = "Adult Night")
-
-		childNightRow.enabled = not altProp.childNightHeader.usePreviousHeader if altProp is not None else True
-		adultDayRow.enabled = not altProp.adultDayHeader.usePreviousHeader if altProp is not None else True
-		adultNightRow.enabled = not altProp.adultNightHeader.usePreviousHeader if altProp is not None else True
-
-		for i in range(len(headerProp.cutsceneHeaders)):
-			drawActorHeaderItemProperty(headerSetup, propUser, headerProp.cutsceneHeaders[i], i, altProp, objName)
-		drawAddButton(headerSetup, len(headerProp.cutsceneHeaders), propUser, None, objName)
+    headerSetup = layout.column()
```
Zelllll, line 359: So a sort of recurring theme that I am noticing is that your editor seems to like spaces instead of tabs. I would suggest trying to convert all of the spaces back to tabs since it is making there be tons of edits in this pr that I don't think you actually meant to change.

Yanis42:
For the first two points you can still choose a custom actor, you just have to choose 'custom' in the actor list and there's a checkbox for entrances, and with the custom mode you still have the possibility to set the override for rotations, it's only automatic for the XML data (also I always thought ACTOR_PLAYER was the only entrance actor, if there's more tell me)

> I can't say this change is good either. You could always edit an existing actor to become a transition actor

I understand, but you just have to change the empty type, which makes more sense since doors are transition actors, imo it's a bit confusing to mix it with other actors, but if that's what people want I'll make it like it was before

> I think you should elaborate on this a bit. The exported data from Fast64 MUST use the enums. If anyone ever shifts the actor table the IDs will all change. Furthermore, Fast64 needs to be updated anyways whenever the enum changes in decomp, so this change doesn't really add anything of value IMO, since the enum being exported is very important.

It's exporting decomp actor IDs, hex IDs are only used for the blend file, if I use decomp names you'd lose data every time they decide to change an ID, that's why there's both in the XML

And for the last thing you said I just haven't thought of that, I'll work on that tomorrow

Zelllll:
Ah so it's like an internal storage thing. And that doesn't affect if you have a custom manually set ID for an actor? Also ACTOR_PLAYER is the only one used in the final game but I think it's technically possible to use a different one (hence the option of being able to set it to begin with)

fast64_internal/oot/oot_constants.py
Yanis42: Yeah that's a good idea, I'll move the enums I added to the XML

fast64_internal/oot/oot_actor.py
Yanis42: Gonna switch back to tabs, I should have done that before PR-ing, my bad

Yanis42:
Yes it's just for internal use, also when you choose the custom actor in the search box the console will tell you it's the ID 'ACTOR_CUSTOM' but it'll export the name you typed on the panel, I tried to be faithful to how it worked before

fast64_internal/oot/ActorList.xml
```diff
@@ -0,0 +1,3266 @@
...
+    <Parameter Params="0700"> Spawn: Thrown Out</Parameter>
+    <Parameter Params="0D00"> Stand</Parameter>
+    <Parameter Params="0E00"> Slow Walk Forward</Parameter>
+    <Parameter Params="0F00"> Retain Previous Velocity //Walk speed set by last exit</Parameter>
+    <Property Mask="0x0000" Name="None"/>
+    <Property Index="1" Mask="0x00FF" Name="Camera ID (nullable)"/>
+    <Notes>Last 2 digits = Camera Initial Focus //Data defined by scene Link spawns in
++00FF No Special Positioning</Notes>
+  </Actor>
+  <Actor Category="ACTORCAT_ENNEMY" ID="ACTOR_EN_TEST" Key="0002" Name="Stalfos - EN_TEST" ObjectID="OBJECT_SK2">
```
mzxrules:
```suggestion
  <Actor Category="ACTORCAT_ENEMY" ID="ACTOR_EN_TEST" Key="0002" Name="Stalfos - EN_TEST" ObjectID="OBJECT_SK2">
```

Yanis42: Apparently I got confused with how you write it in french, gonna fix that

Yanis42:
I think I haven't forgot anything, I'll do more bug research tomorrow (everything should work but I'll do more tests just in case)

Zelllll (approved):

Yanis42:
There's a bug where the informations on the panel won't save for some reasons, I don't really know why and how, gonna look into that later today

Yanis42:
> There's a bug where the informations on the panel won't save for some reasons, I don't really know why and how, gonna look into that later today (it seems to be a precise case so I missed it when I looked into bugs the other day)

Apparently it's resetting every actor if a new one is added then the file's saved, I have no idea why this happens

Yanis42:
Ok the bug I found the other day should be fixed now, this may need more testing by other people than me to make sure everything works, it should though (I tried an older blend file of mine and new ones)

fast64_internal/oot/oot_utility.py
```diff
@@ -599,3 +630,260 @@ def oot_utility_register():
...
+	if attr != '0x00' and attr is not None:					
+		if isinstance(attr, str):
+			if shift != '0':
+				return f"(({attr} << {shift}) & {mask}) | "
+			else:
+				return f"({attr} & {mask}) | " 
+		elif isinstance(attr, bool) and attr:
+			return f"(1 << {shift}) | "
+		else:
+			return ""
```
Dragorn421, line 646: is there a case in which attr being neither str nor bool is not an error?
maybe this should raise PluginError / NotImplementedError instead of defaulting to ""

fast64_internal/oot/oot_utility.py
```diff
@@ -599,3 +630,260 @@ def oot_utility_register():
...
+	else:
+		return ""
+
+def getActorFinalParameters(detailedProp, actorKey, paramTarget, field):
+	'''Returns the full parameter string'''
+	params = ""
+	if field != None:
+		panelParams = getattr(detailedProp, field, "")
+		if panelParams == "": return "0x0"
```
Dragorn421, line 655:
```suggestion
		if panelParams == "":
			return "0x0"
```

fast64_internal/oot/oot_utility.py
```diff
@@ -599,3 +630,260 @@ def oot_utility_register():
...
+		if actorKey == actorNode.get('Key'):
+			lenProp = getMaxElemIndex(actorKey, 'Property', None)
+			lenSwitch = getMaxElemIndex(actorKey, 'Flag', 'Switch')
+			lenBool = getMaxElemIndex(actorKey, 'Bool', None)
+			lenEnum = getMaxElemIndex(actorKey, 'Enum', None)
+			for elem in actorNode:
+				actorType = getattr(detailedProp, actorKey + '.type', None)
+				if hasTiedParams(elem.get('TiedParam'), actorType):
+					params += getActorString(elem, detailedProp, actorKey, lenProp, lenSwitch, lenBool, lenEnum, paramTarget)
+	params = params.rstrip(' | ')
```
Dragorn421, line 666:
```suggestion
	params = params.rstrip().rstrip('|').rstrip()
```
may be more what you intended here. this handles any whitespace (rstrip() with no arguments), not just spaces

fast64_internal/oot/oot_utility.py
```diff
@@ -599,3 +630,260 @@ def oot_utility_register():
...
+	if field.endswith('.type'):
+		setattr(object, field, f'{param & mask:04X}')
+	else:
+		attr = getattr(object, field, '0x0')
+		if isinstance(attr, str):
+			setattr(object, field, f'0x{(param & mask) >> shift:02X}')
+		elif isinstance(attr, bool):
+			setattr(object, field, bool((param & mask) >> shift))
+		else:
+			setattr(object, field, '0x09')
```
Dragorn421, line 690: why 0x09
same as before maybe this should just error idk

fast64_internal/oot/oot_utility.py
```diff
@@ -599,3 +630,260 @@ def oot_utility_register():
...
+	for actorNode in root:
+		if actorNode.get('Key') == actorKey:
+			for elem in actorNode:
+				if elem.tag == elemTag:
+					if flagType is None or (flagType == 'Switch' and elem.get('Type') == flagType):
+						length = elem.get('Index')
+	return length
+
+def hasTiedParams(tiedParam, actorType):
+	'''Looking for parameters that depends on other parameters'''
```
Dragorn421, line 704:
```suggestion
	'''Looking for parameters that depend on other parameters'''
```

fast64_internal/oot/oot_utility.py
```diff
@@ -599,3 +630,260 @@ def oot_utility_register():
...
+	if tiedParam is None or actorType is None:
+		return True
+	else:
+		if tiedParam.find(',') != -1: tiedList = tiedParam.split(',')
+		else: tiedList = [tiedParam]
```
Dragorn421, line 709: madge

fast64_internal/oot/oot_utility.py
Yanis42: I don't really remember what I intended there, I think I thought it was useless to raise an error, I can't tell why though

I can't get a proper opinion on this right now but you're probably right

Yanis42: I did figured out why I did it that way and fixed it

fast64_internal/oot/oot_operators.py
```diff
@@ -158,7 +170,29 @@ def execute(self, context):
...
 		context.view_layer.objects.active = csObj
 		return {"FINISHED"}
-	
+
+class OOT_AddActor(bpy.types.Operator):
+	bl_idname = 'object.oot_add_actor'
+	bl_label = "Add Actor"
+	bl_options = {'REGISTER', 'UNDO', 'PRESET'}
+
+	scale : bpy.props.FloatProperty(default = 2)
```
Dragorn421, line 179: this is unused, can be removed afaik

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
 
-	actorID : bpy.props.EnumProperty(items = ootEnumActorID, default = "ACTOR_PLAYER")
-	actorUser : bpy.props.StringProperty(default = "Actor")
+class OOT_UpgradeActors(bpy.types.Operator):
+	bl_idname = 'object.oot_upgrade_actors'
+	bl_label = "Upgrade Data Now!"
+	bl_options = {'REGISTER', 'UNDO', 'PRESET'}
+
+	version: bpy.props.IntProperty(name="OOT_ObjectProperties Version", default=0)
+	cur_version = 1 # version after property migration
```
Dragorn421, line 17: `version` is unused and I think this shouldn't duplicate `cur_version = 1` but rather reuse `OOT_ObjectProperties.cur_version`
(even simpler just change this operator's execute to simply call `OOT_ObjectProperties.upgrade_changed_props`)

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+		if obj.data is None:
+			if obj.ootEmptyType == 'Scene':
+				print(f"Processing Scene '{obj.name}'...")
+				for roomObj in obj.children:
+					if roomObj.data is None:
+						if roomObj.ootEmptyType == 'Room':
+							# Actor is parented to a room
+							for obj in roomObj.children:
+								upgradeActorInit(obj)
+						elif (roomObj.ootEmptyType == 'Actor' or \
```
Dragorn421, line 76:
```suggestion
						elif (roomObj.ootEmptyType == 'Actor' or
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+							roomObj.ootEmptyType == 'Transition Actor' or roomObj.ootEmptyType == 'Entrance'):
+							# Actor is parented to a scene
+							upgradeActorInit(roomObj)
```
Dragorn421, line 79: calling `roomObj` something that may be a room or something else is confusing, I can suggest the uninspired `sceneChildObj` but idk

alternatively you could turn `upgrade_object` into a recursive function which may approximately look like
```py
if obj.data is None:
	if obj.ootEmptyType == 'Scene' (or Room):
		for child in obj.children:
			upgrade_object(child) # recursion
	elif obj is actor stuff:
		upgradeActorInit(obj)
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+			elif (obj.ootEmptyType == 'Actor' or obj.ootEmptyType == 'Transition Actor' or obj.ootEmptyType == 'Entrance'):
+				upgradeActorInit(obj)
+
+def getValues(self, actorID, actorField, target, paramField):
+	'''Updates the actor parameter field when the user change the options'''
```
Dragorn421, line 84:
```suggestion
	'''Updates the actor parameter field when the user changes the options'''
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+	if self.isActorSynced:
```
Dragorn421, line 85: instead of `isActorSynced` shouldn't this be a `version == cur_version` check?

having an "is up-to-date" boolean is a problem because obviously it has two values so it's not friendly for reusing in a future update. and it gets obsoleted immediately

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+						paramTarget = elem.get('Target')
+						if paramTarget is None: paramTarget = 'Params'
+						if paramTarget == target:
+							tiedParam = elem.get('TiedParam')
+							actorType = getattr(self, dPKey + '.type', None)
+							if hasTiedParams(tiedParam, actorType) is True:
+								value = getActorFinalParameters(actorProp, dPKey, paramTarget, None)
+			if target == 'XRot': self.XRotBool = True
+			if target == 'YRot': self.YRotBool = True
+			if target == 'ZRot': self.ZRotBool = True
```
Dragorn421, line 102: 👺 
```suggestion
			if target == 'ZRot':
				self.ZRotBool = True
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+			for actorNode in root:
+				if actorNode.get('ID') == actorID:
+					dPKey = actorNode.get('Key')
+					for elem in actorNode:
+						paramTarget = elem.get('Target')
+						if paramTarget is None: paramTarget = 'Params'
+						if paramTarget == target:
+							tiedParam = elem.get('TiedParam')
+							actorType = getattr(self, dPKey + '.type', None)
+							if hasTiedParams(tiedParam, actorType) is True:
```
Dragorn421, line 98:
```suggestion
							if hasTiedParams(tiedParam, actorType):
```

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,266 @@ def oot_utility_register():
...
+		else:
+			tiedList = [tiedParam]
+
+		if actorType is not None:
+			if len(tiedList) == 1 and tiedList[0] == actorType:
+				return True
+			else:
+				for j in range(len(tiedList)):
+					if tiedList[j] == actorType:
+						return True
```
Dragorn421, line 726:
```suggestion
						return True
	return False
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+								value = getActorFinalParameters(actorProp, dPKey, paramTarget, None)
+			if target == 'XRot': self.XRotBool = True
+			if target == 'YRot': self.YRotBool = True
+			if target == 'ZRot': self.ZRotBool = True
+		return value
+	else:
+		if paramField == 'transActorParam': actorProp = bpy.context.object.ootTransitionActorProperty.actor
+		else: actorProp = bpy.context.object.ootActorProperty
```
Dragorn421, line 106: 👺

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+	'''This function is used to edit the OOTActorProperties class before it's registered'''
+	propAnnotations = getattr(OOTActorProperties, '__annotations__', None)
+	if propAnnotations is None:
+		propAnnotations = {}
+		OOTActorProperties.__annotations__ = propAnnotations
+
+	# Each prop has its 'custom' variant because of the get and set functions
+	# Actors/Entrance Actor
+	propAnnotations['actorID'] = bpy.props.EnumProperty(name='Actor ID', items=ootEnumActorID)
+	propAnnotations['actorIDCustom'] = bpy.props.StringProperty(name='Actor Key', default='ACTOR_CUSTOM')
```
Dragorn421, line 194: imo these properties that are not dynamic (don't depend on the xml data) can be in `OOTActorProperties` directly

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+	propAnnotations['actorKey'] = bpy.props.StringProperty(name='Actor Key', default='0000')
+	propAnnotations['actorParam'] = bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000', \
```
Dragorn421, line 196:
```suggestion
	propAnnotations['actorParam'] = bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000',
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+	propAnnotations['ZRotToSave'] = bpy.props.StringProperty(default='0x0')
+
+	# Collectible Drops List
+	itemDrops = [(elem.get('Value'), elem.get('Name'), \
+					elem.get('Name')) for listNode in root for elem in listNode if listNode.tag == 'List' \
+					and listNode.get('Name') == 'Collectibles']
+
+	# Generate the fields
+	for actorNode in root:
+		i = j = k = l = 1
```
Dragorn421, line 252: maybe i j k l should be `i_property`, `i_flag`, `i_bool`, `i_enum` since there's one for each of those. at first I though there was going to be a whole bunch of for loops monka

fast64_internal/oot/oot_actor.py
```diff
@@ -86,116 +449,134 @@ def drawActorHeaderProperty(layout, headerProp, propUser, altProp, objName):
...
-	split.label(text = getEnumName(ootEnumActorID, actorProp.actorID))
+	if detailedProp.isActorSynced:
+		drawDetailedProperties('Actor Property', actorProp, actorIDBox, objName, \
+			OOT_SearchActorIDEnumOperator, 'actorID', 'actorParam', detailedProp, detailedProp.actorKey)
+		drawActorHeaderProperty(actorIDBox, actorProp.headerSettings, "Actor", altRoomProp, objName)
+	else:
+		sceneObj = getSceneObj(bpy.context.object)
+		if sceneObj is None: sceneName = 'Unknown'
+		else: sceneName = sceneObj.name
+		actorIDBox.box().label(text=f"Scene: '{sceneName}' Actors are not synchronised!")
```
Dragorn421, line 503: I think "actors data is from an old fast64 version" would be more understandable to the end user. (especially with the "upgrade now" button below, and ideally both of these shouldn't show up at all if the blender handler works)
(also I don't like "synchronised" itself personally)

fast64_internal/oot/oot_actor.py
```diff
@@ -86,116 +449,134 @@ def drawActorHeaderProperty(layout, headerProp, propUser, altProp, objName):
...
 	else:
-		label_split(actorIDBox, "Room To Transition From", str(roomObj.ootRoomHeader.roomIndex))
-	prop_split(actorIDBox, transActorProp, "roomIndex", "Room To Transition To")
-	actorIDBox.label(text = "Y+ side of door faces toward the \"from\" room.", icon = 'ORIENTATION_NORMAL')
-	drawEnumWithCustom(actorIDBox, transActorProp, "cameraTransitionFront", "Camera Transition Front", "")
-	drawEnumWithCustom(actorIDBox, transActorProp, "cameraTransitionBack", "Camera Transition Back", "")
+		sceneObj = getSceneObj(bpy.context.object)
+		if sceneObj is None: sceneName = 'Unknown'
+		else: sceneName = sceneObj.name
+		actorIDBox.box().label(text=f"Scene: '{sceneName}' Actors are not synchronised!")
```
Dragorn421, line 562: add "upgrade data" button here?

fast64_internal/oot/oot_operators.py
```diff
@@ -82,21 +86,28 @@ def execute(self, context):
 		entranceObj = context.view_layer.objects.active
 		entranceObj.ootEmptyType = "Entrance"
 		entranceObj.name = "Entrance"
-		entranceObj.ootEntranceProperty.actor.actorParam = "0x0FFF"
+		entranceObj.fast64.oot.actor.actorID = 'ACTOR_PLAYER'
+		entranceObj.fast64.oot.actor.actorKey = '0000'
+		setattr(entranceObj.fast64.oot.actor, '0000.type', '0F00')
+		setattr(entranceObj.fast64.oot.actor, '0000.props1', '0xFF')
```
Dragorn421, line 92: I will say I'm very much not a fan of hex without the proper 0x prefix
I also saw hex without 0x in the UI but it may be up to the xml

fast64_internal/oot/oot_level.py
```diff
@@ -169,18 +169,31 @@ def onUpdateOOTEmptyType(self, context):
 
 class OOT_ObjectProperties(bpy.types.PropertyGroup):
 	version: bpy.props.IntProperty(name="OOT_ObjectProperties Version", default=0)
-	cur_version = 0 # version after property migration
+	cur_version = 1 # version after property migration
```
Dragorn421, line 172: the default of `version` should be cur_version, as new objects will be on the current version obviously. and old objects already have this property set so changing the default won't change those
this also means you could then remove setting `version = cur_version` in the `OOT_AddScene` and `OOT_AddDoor` operators, and maybe elsewhere

fast64_internal/oot/oot_operators.py
```diff
@@ -158,7 +170,29 @@ def execute(self, context):
...
+
+	scale : bpy.props.FloatProperty(default = 2)
+
+	def execute(self, context):
+		if context.mode != "OBJECT":
+			bpy.ops.object.mode_set(mode = "OBJECT")
+		bpy.ops.object.select_all(action = "DESELECT")
+
+		location = mathutils.Vector(bpy.context.scene.cursor.location)
+		bpy.ops.object.empty_add(type='CUBE', radius = 1, align='WORLD', location=location[:])
```
Dragorn421, line 187: a nice quality of life feature would be to add a property to the operator for setting the room the actor is parented to. it would show up in the redo panel. just a suggestion though

Dragorn421 (changes requested):

fast64_internal/oot/oot_actor.py
Yanis42: I tried but it gave me an error since I need to do something like ``from .oot_level import OOT_ObjectProperties`` so I figured copy pasting the code was the easiest solution (and since I can't test this it was the best thing to do tbh)

fast64_internal/oot/oot_actor.py
Yanis42: is ``Legacy data from scene: '{sceneName}' has not been upgraded!`` better?

fast64_internal/oot/oot_operators.py
Yanis42: for the actor keys it's this way to keep an easy compatibility between fast64 versions because of decomp still evolving (avoids having a headache everytime fast64 gets broken because of decomp changes)

I think I can add the 0x prefix but would it be odd to have something like: ``fast64.oot.actor.0x0000.type``?

fast64_internal/oot/oot_operators.py
Yanis42: I'm not sure what you mean there, like automatically parenting the actor to the room if there's one? if so, it could give issues if you have more than 1 room

fast64_internal/oot/oot_actor.py
Dragorn421: an import is definitely better than having to update a value in two places

and yeah imports in the whole codebase are a mess. they need cleaning up anyway, dw about adding imports that may look weird

fast64_internal/oot/oot_operators.py
Dragorn421: > like automatically parenting the actor to the room if there's one?

yes

> it could give issues if you have more than 1 room

I'm thinking the operator would have a room picker. (in the redo panel, not as a popup, and default would be to not parent to any room)

again this is just a feature suggestion, nothing mandatory to change

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
-	bl_label = "Select Actor ID"
-	bl_property = "actorID"
-	bl_options = {'REGISTER', 'UNDO'} 
+# General classes and functions
 
-	actorID : bpy.props.EnumProperty(items = ootEnumActorID, default = "ACTOR_PLAYER")
-	actorUser : bpy.props.StringProperty(default = "Actor")
+class OOT_UpgradeActors(bpy.types.Operator):
+	bl_idname = 'object.oot_upgrade_actors'
+	bl_label = "Upgrade Data Now!"
```
Dragorn421, line 13: `bl_label` should be more (very briefly) descriptive of the operator
Maybe `bl_label = "Upgrade Fast64 OoT actors data"`

and "Upgrade Data Now!" can be used for the ui like `actorIDBox.operator(OOT_UpgradeActors.bl_idname, text="Upgrade Data Now!")`

Also you can test this if you mimic blender handlers not working, which seems to be what happens for me sometimes, for example by removing https://github.com/Fast-64/fast64/blob/347d19e3fbc8449aaba91de337952fe3d738f4b9/__init__.py#L359

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,410 @@
...
+		drawParams(userLayout, detailedProp, dpKey, f'{dpKey}.switchFlag', 'Switch Flag', 'Flag', 'Switch', getMaxElemIndex(dpKey, 'Flag', 'Switch'))
+		drawParams(userLayout, detailedProp, dpKey, f'{dpKey}.enum', None, 'Enum', None, None)
+		drawParams(userLayout, detailedProp, dpKey, f'{dpKey}.props', None, 'Property', None, None)
+		drawParams(userLayout, detailedProp, dpKey, f'{dpKey}.bool', None, 'Bool', None, None)
+
+		paramBox = userLayout.box()
+		paramBox.label(text="Actor Parameter")
+		params = getattr(detailedProp, userParamField, "")
+		if params != "": paramBox.prop(detailedProp, userParamField, text="")
+		else: paramBox.label(text="That Actor doesn't have parameters.")
```
Dragorn421, line 361:
```suggestion
		else: paramBox.label(text="This Actor doesn't have parameters.")
```
what happens with custom actors btw?

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,266 @@ def oot_utility_register():
...
 	for cls in reversed(oot_utility_classes):
 		unregister_class(cls)
+
+def getActorParameter(object, field, shift, mask):
+	'''Returns the actor parameter with the shifts'''
+	attr = getattr(object, field, '0x00')
+	if attr != '0x00' and attr is not None:					
+		if isinstance(attr, str):
+			if shift != '0':
+				return f"(({attr} << {shift}) & {mask}) | "
```
Dragorn421, line 643: I think the ` | ` at the end doesn't fit the function

Removing it from here and handling `|`ing the parts in `getActorString` would be cleaner, and you could remove `params = params.rstrip().rstrip('|').rstrip()`

fast64_internal/oot/oot_actor.py
Yanis42: I have some issues when I move things like ``actorParam``

fast64_internal/oot/oot_actor.py
Yanis42: this code will execute if it's not a custom actor so it should be fine

fast64_internal/oot/oot_operators.py
Yanis42: I'll let this to someone else, I don't feel like doing this right now

fast64_internal/oot/oot_actor.py
Yanis42: I can't move a lot of annotations since some are needed for the upgrade, I think I can only move the ``ToSave`` annotations since it's only used for saving in the blend

fast64_internal/oot/oot_utility.py
Dragorn421: to get rid of `params = params.rstrip().rstrip('|').rstrip()`, I was thinking about collecting params parts into a list in `getActorString` and then using `" | ".join()`
it would also help with not having to append `" | "` when the params part is empty

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,282 @@ def oot_utility_register():
...
+					tmpParams = getActorParameter(detailedProp, (field + f'.bool{i}'), shift, strMask)
+					if tmpParams != "":
+						params += tmpParams + charOR
+		elif elem.tag == 'Enum':
+			for i in range(1, (int(lenEnum, base=10) + 1)):
+				if i == int(elem.get('Index'), base=10) and target == paramTarget:
+					tmpParams = getActorParameter(detailedProp, (field + f'.enum{i}'), shift, strMask)
+					if tmpParams != "":
+						params += tmpParams + charOR
+	return params
```
Dragorn421, line 789: `params = []`
```suggestion
					paramsPart = getActorParameter(detailedProp, (field + f'.enum{i}'), shift, strMask)
					if paramsPart != "":
						params.append(paramsPart)
	return " | ".join(params)
```
not sure about how to call `paramsPart` as a variable but yeah

I would also change `paramsPart != ""` to `paramsPart is not None` and make `getActorParameter` return `None` instead of empty strings

Also `getActorParameter` may get a more descriptive name like `getAndFormatActorParameterPart` idk

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,282 @@ def oot_utility_register():
...
+				params += tmpParams + charOR
+		elif elem.tag == 'Bool':
+			for i in range(1, (int(lenBool, base=10) + 1)):
+				if i == int(elem.get('Index'), base=10) and target == paramTarget:
+					tmpParams = getActorParameter(detailedProp, (field + f'.bool{i}'), shift, strMask)
+					if tmpParams != "":
+						params += tmpParams + charOR
+		elif elem.tag == 'Enum':
+			for i in range(1, (int(lenEnum, base=10) + 1)):
+				if i == int(elem.get('Index'), base=10) and target == paramTarget:
```
Dragorn421, line 785: is this different from
```suggestion
			i = int(elem.get('Index'), base=10)
			if i >= 1 and i <= int(lenEnum, base=10):
				if target == paramTarget:
```
🤔 ?

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,282 @@ def oot_utility_register():
...
+
+def hasTiedParams(tiedParam, actorType):
+	'''Looking for parameters that depend on other parameters'''
+	if tiedParam is None or actorType is None:
+		return True
+	else:
+		if tiedParam.find(',') != -1:
+			tiedList = tiedParam.split(',')
+		else:
+			tiedList = [tiedParam]
```
Dragorn421, line 718:
```suggestion
		tiedList = tiedParam.split(',')
```
```
>>> "abcdef".split(",")
['abcdef']
```

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,282 @@ def oot_utility_register():
...
+
+		if actorType is not None:
+			if len(tiedList) == 1 and tiedList[0] == actorType:
+				return True
+			else:
+				for j in range(len(tiedList)):
+					if tiedList[j] == actorType:
+						return True
```
Dragorn421, line 726:
```suggestion
			return actorType in tiedList
```
😅

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,282 @@ def oot_utility_register():
...
+		attr = getattr(object, field, '0x0')
+		if isinstance(attr, str):
+			setattr(object, field, f'0x{(param & mask) >> shift:02X}')
+		elif isinstance(attr, bool):
+			setattr(object, field, bool((param & mask) >> shift))
+		else:
+			raise NotImplementedError
+
+def getMaxElemIndex(actorKey, elemTag, flagType):
+	'''Looking for the highest index for the current actor'''
```
Dragorn421, line 700: I'm not sure where the "highest" part of this is implemented in this function? there's no max() or comparison

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,401 @@
...
+	propAnnotations['transActorIDCustom'] = bpy.props.StringProperty(name='Actor Key', default='ACTOR_CUSTOM')
+	propAnnotations['transActorKey'] = bpy.props.StringProperty(name='Transition Actor ID', default='0009')
+	propAnnotations['transActorParam'] = bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000',
+		get=lambda self: getValues(self, self.transActorID, 'actorParam', 'Params', 'transActorParam'),
+		set=lambda self, value: setValues(self, value, 'Params', 'transActorID'))
+	propAnnotations['transActorParamCustom'] = bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000')
+
+	# Other
+	# isActorSynced is used to check if the blend's data is from an older version of Fast64
+	propAnnotations['isActorSynced'] = bpy.props.BoolProperty(default=False)
```
Dragorn421, line 224: I think `isActorSynced` should be gotten rid of entirely, as I said it's duplicating what a version != cur_version check does,
and it's going to be useless besides this one particular version upgrade, it couldn't be reused for other version changes

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,401 @@
...
+	propAnnotations['ZRotBool'] = bpy.props.BoolProperty(default=False)
+
+	# Transition Actors
+	propAnnotations['transActorID'] = bpy.props.EnumProperty(name='Transition Actor ID', items=ootEnumTransitionActorID)
+	propAnnotations['transActorIDCustom'] = bpy.props.StringProperty(name='Actor Key', default='ACTOR_CUSTOM')
+	propAnnotations['transActorKey'] = bpy.props.StringProperty(name='Transition Actor ID', default='0009')
+	propAnnotations['transActorParam'] = bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000',
+		get=lambda self: getValues(self, self.transActorID, 'actorParam', 'Params', 'transActorParam'),
+		set=lambda self, value: setValues(self, value, 'Params', 'transActorID'))
+	propAnnotations['transActorParamCustom'] = bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000')
```
Dragorn421, line 220: so what goes wrong when putting these properties (which don't rely on data from the xml) in `OOTActorProperties` exactly?

EDIT: see another comment I think I figured out the issue

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,401 @@
...
 		return {'RUNNING_MODAL'}
 
+class OOTActorProperties(bpy.types.PropertyGroup):
+	# If you use the 'get=' option from Blender props don't actually save the data in the .blend
+	# When the get function is called we have to save the data that'll be returned
+	paramToSave = bpy.props.StringProperty()
+	transParamToSave = bpy.props.StringProperty()
+	XRotToSave = bpy.props.StringProperty(default='0x0')
+	YRotToSave = bpy.props.StringProperty(default='0x0')
+	ZRotToSave = bpy.props.StringProperty(default='0x0')
```
Dragorn421, line 56: you need to use `:` to define annotations
this is probably the issue you had when moving props to the class definition

https://docs.blender.org/api/current/bpy.props.html#propertygroup-example

Dragorn421 (changes requested):

fast64_internal/oot/oot_actor.py
Yanis42: works fine lol
I just completely forgot you need to replace ``=`` by ``:`` when I first moved everything

fast64_internal/oot/oot_utility.py
Yanis42: it returns the last index string from the XML, I should reword it this way: "Looking for the last index of the current actor"

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+			entranceProp = obj.ootEntranceProperty.actor
+			upgradeActorProcess(objType, obj, entranceProp.actorID, obj.fast64.oot.actor,
+				int(entranceProp.actorParam, base=16), 'param', 'actorID', 'actorParam', 'Params')
+			obj.fast64.oot.actor.actorParam
+
+	for childObj in obj.children:
+		upgradeActorInit(childObj)
+
+def upgradeActorProcess(user, obj, actorID, detailedProp, params, toSaveField, idField, paramField, paramTarget):
+	if obj.ootEntranceProperty.customActor == False and actorID != 'Custom':
```
Dragorn421, line 867:
```suggestion
	if not obj.ootEntranceProperty.customActor and actorID != 'Custom':
```

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+			setActorParameter(detailedProp, field + '.type', params, mask)
+		if elem.tag == 'Flag':
+			elemType = elem.get('Type')
+			if elemType == 'Chest':
+				setActorParameter(detailedProp, field + '.chestFlag', params, mask)
+			elif elemType == 'Collectible':
+				setActorParameter(detailedProp, field + '.collectibleFlag', params, mask)
+			elif elemType == 'Switch':
+				for i in range(1, (int(lenSwitch, base=10) + 1)):
+					if i == int(elem.get('Index'), base=10):
```
Dragorn421, line 805: same as `i =` + range check, like previously

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+				params.append(paramPart)
+		elif elem.tag == 'Bool':
+			i = int(elem.get('Index'), base=10)
+			if (i >= 1 and i <= int(lenBool, base=10)) and (target == paramTarget):
+					paramPart = getAndFormatActorParameterPart(detailedProp, (field + f'.bool{i}'), shift, strMask)
+					if paramPart is not None:
+						params.append(paramPart)
+		elif elem.tag == 'Enum':
+			i = int(elem.get('Index'), base=10)
+			if (i >= 1 and i <= int(lenEnum, base=10)) and (target == paramTarget):
```
Dragorn421, line 780: isn't `i >= 1 and i <= int(lenEnum, base=10)` always supposed to be true? or at least if it as false it would be an issue

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+		tiedList = tiedParam.split(',')
+		if actorType is not None:
+			return actorType in tiedList
+	return False
+
+def getActorString(elem, detailedProp, field, lenProp, lenSwitch, lenBool, lenEnum, paramTarget):
+	'''Returns the current actor's parameter string'''
+	shiftTmp = 0
+	params = []
+	paramPart = ""
```
Dragorn421, line 728:
```suggestion
```

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+	if tiedParam is None or actorType is None:
+		return True
+	else:
+		tiedList = tiedParam.split(',')
+		if actorType is not None:
+			return actorType in tiedList
+	return False
+
+def getActorString(elem, detailedProp, field, lenProp, lenSwitch, lenBool, lenEnum, paramTarget):
+	'''Returns the current actor's parameter string'''
```
Dragorn421, line 725: the function doesn't return a string anymore

it needs a better name overall

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+	'''Looking for the last index of an actor's property (from XML data)'''
+	indexes = []
+	for actorNode in root:
+		if actorNode.get('Key') == actorKey:
+			for elem in actorNode:
+				if elem.tag == elemTag:
+					if flagType is None or (flagType == 'Switch' and elem.get('Type') == flagType):
+						indexes.append(elem.get('Index'))
+	indexes.sort()
+	return f"{indexes[len(indexes) - 1]}" if len(indexes) != 0 else "0"
```
Dragorn421, line 712: why convert to string here and later convert back to int? it probably should stay int and only convert to string when needed

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,279 @@ def oot_utility_register():
...
+		if isinstance(attr, str):
+			setattr(object, field, f'0x{(param & mask) >> shift:02X}')
+		elif isinstance(attr, bool):
+			setattr(object, field, bool((param & mask) >> shift))
+		else:
+			raise NotImplementedError
+
+def getLastElemIndex(actorKey, elemTag, flagType):
+	'''Looking for the last index of an actor's property (from XML data)'''
+	indexes = []
```
Dragorn421, line 704:
```suggestion
	indices = []
```
the proper plural of index is indices
like vertex -> vertices

fast64_internal/oot/oot_utility.py
Dragorn421, line 712: `indexes[len(indexes) - 1]` can be just `indexes[-1]` in python, but moreover:
```suggestion
	return max(indices) if indices else 0
```
also idk about 0 being the default
also idk if the function should check all indices are >0 ?
(I didn't entirely understand what this is for yet)

Dragorn421 (changes requested):

fast64_internal/oot/oot_utility.py
Yanis42: yeah you're right, it's always true since it's in the actorNode/elem loop

fast64_internal/oot/oot_level_writer.py
```diff
@@ -460,34 +460,66 @@ def ootProcessEmpties(scene, room, sceneObj, obj, transformMatrix):
...
-				translation, rotation[1], # TODO: Correct axis?
-				transActorProp.actor.actorParam), transActorProp.actor, "transitionActorList", obj.name)
-			#scene.transitionActorList.append(OOTTransitionActor(
-			#	getCustomProperty(transActorProp.actor, "actorID"),
-			#	room.roomIndex, transActorProp.roomIndex,
-			#	getCustomProperty(transActorProp, "cameraTransitionFront"),
-			#	getCustomProperty(transActorProp, "cameraTransitionBack"),
-			#	translation, rotation[1], # TODO: Correct axis?
-			#	transActorProp.actor.actorParam))
+				translation, rotY, # TODO: Correct axis?
```
Dragorn421, line 510:
```suggestion
				translation, rotY,
```
I assume this was about "is +y the upwards axis" and yes it is

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +633,278 @@ def oot_utility_register():
...
+			setActorParameterPart(detailedProp, field + '.type', params, mask)
+		if elem.tag == 'Flag':
+			elemType = elem.get('Type')
+			if elemType == 'Chest':
+				setActorParameterPart(detailedProp, field + '.chestFlag', params, mask)
+			elif elemType == 'Collectible':
+				setActorParameterPart(detailedProp, field + '.collectibleFlag', params, mask)
+			elif elemType == 'Switch':
+				i = int(elem.get('Index'), base=10)
+				if lenSwitch is not None and (i >= 1 and i <= lenSwitch) and target == paramTarget:
```
Dragorn421, line 804: did you mean to keep the range check in there

fast64_internal/oot/oot_utility.py
```diff
@@ -701,15 +701,15 @@ def setActorParameter(object, field, param, mask):
...
 	for actorNode in root:
 		if actorNode.get('Key') == actorKey:
 			for elem in actorNode:
 				if elem.tag == elemTag:
 					if flagType is None or (flagType == 'Switch' and elem.get('Type') == flagType):
-						indexes.append(elem.get('Index'))
-	indexes.sort()
-	return f"{indexes[len(indexes) - 1]}" if len(indexes) != 0 else "0"
+						indices.append(int(elem.get('Index'), base=10))
+	indices.sort()
```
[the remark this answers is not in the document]
Yanis42: is ``indices.sort()`` not needed at all? I used that because I thought it would be the fix to the issue where the indices from the XML aren't in the correct order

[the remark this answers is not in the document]
Dragorn421: max() gets the maximum, it's the same as sort() and take the last argument

fast64_internal/oot/oot_utility.py
```diff
@@ -379,6 +380,36 @@ def getCustomProperty(data, prop):
...
+				return getActorParameter(detailedProp, detailedProp.actorKey, 'Params', field)
+			elif field == 'transActorParam':
+				return getActorParameter(detailedProp, detailedProp.transActorKey, 'Params', field)
+			else:
+				for actorNode in root:
+					dpKey = detailedProp.actorKey
+					if dpKey == actorNode.get('Key'):
+						for elem in actorNode:
+							target = elem.get('Target')
+							actorType = getattr(detailedProp, dpKey + '.type', None)
```
Dragorn421, line 396: moving stuff outside the loops
```py
				dpKey = detailedProp.actorKey
				actorType = getattr(detailedProp, dpKey + '.type', None)
				for actorNode in root:
					if dpKey == actorNode.get('Key'):
						for elem in actorNode:
							target = elem.get('Target')
```

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,393 @@
...
+		set=lambda self, value: setValues(self, value, 'ZRot', 'actorID'))
+	rotOverrideCustom : bpy.props.BoolProperty(name = 'Override Rotation', default = False)
+	rotOverrideXCustom : bpy.props.StringProperty(name = 'Rot X', default = '0x0')
+	rotOverrideYCustom : bpy.props.StringProperty(name = 'Rot Y', default = '0x0')
+	rotOverrideZCustom : bpy.props.StringProperty(name = 'Rot Z', default = '0x0')
+
+	# We have to use a bool to know what's needed to be exported
+	XRotBool : bpy.props.BoolProperty(default=False)
+	YRotBool : bpy.props.BoolProperty(default=False)
+	ZRotBool : bpy.props.BoolProperty(default=False)
```
Dragorn421, line 79: why are these named "Bool"? I'm not sure I understand what these or the `rotxbool` variables are for

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,393 @@
...
+	rotOverrideY : bpy.props.StringProperty(name = 'Rot Y', default = '0x0',
+		get=lambda self: getValues(self, self.actorID, None, 'YRot', None),
+		set=lambda self, value: setValues(self, value, 'YRot', 'actorID'))
+	rotOverrideZ : bpy.props.StringProperty(name = 'Rot Z', default = '0x0',
+		get=lambda self: getValues(self, self.actorID, None, 'ZRot', None),
+		set=lambda self, value: setValues(self, value, 'ZRot', 'actorID'))
+	rotOverrideCustom : bpy.props.BoolProperty(name = 'Override Rotation', default = False)
+	rotOverrideXCustom : bpy.props.StringProperty(name = 'Rot X', default = '0x0')
+	rotOverrideYCustom : bpy.props.StringProperty(name = 'Rot Y', default = '0x0')
+	rotOverrideZCustom : bpy.props.StringProperty(name = 'Rot Z', default = '0x0')
```
Dragorn421, line 74: why do custom actors need to use the "Custom" properties?

fast64_internal/oot/oot_actor.py
Yanis42: it's used to let the export know which rotation it should export, the blender's one or the parameter one, but I guess I could just check if the value is not 0

fast64_internal/oot/oot_actor.py
Yanis42: iirc I used that because I had some issues when using the non "Custom" one

fast64_internal/oot/oot_actor.py
Dragorn421: well checking if it's 0 isn't perfect because you may want to have 0
couldn't this be determined by looking in the xml data if the actor stores any data in its rotations instead?

fast64_internal/oot/oot_actor.py
```diff
@@ -360,15 +368,17 @@ def drawDetailedProperties(user, userProp, userLayout, userObj, userSearchOp, us
...
 		if user != userEntrance:
-			prop_split(userLayout, detailedProp, userIDField + 'Custom', currentActor)
-			prop_split(userLayout, detailedProp, userParamField + 'Custom', 'Actor Parameter')
-			userLayout.prop(detailedProp, 'rotOverrideCustom', text = 'Override Rotation (ignore Blender rot)')
-			if detailedProp.rotOverrideCustom:
-				if user == userActor: prop_split(userLayout, detailedProp, 'rotOverrideXCustom', 'Rot X')
-				prop_split(userLayout, detailedProp, 'rotOverrideYCustom', 'Rot Y')
-				if user == userActor: prop_split(userLayout, detailedProp, 'rotOverrideZCustom', 'Rot Z')
+			prop_split(userLayout, detailedProp, getCustomPropName(userIDField), currentActor)
+			prop_split(userLayout, detailedProp, userParamField, 'Actor Parameter')
```
Dragorn421, line 372:
```suggestion
			prop_split(userLayout, detailedProp, getCustomPropName(userParamField), 'Actor Parameter')
```
?
but also I don't think you need `getCustomPropName` anywhere actually, using the non-custom prop with a custom actor will use the custom prop thanks to getValues/setValues

fast64_internal/oot/oot_constants.py
```diff
@@ -1,3 +1,11 @@
+import os, xml.etree.ElementTree as ET
+from ..utility import PluginError
+
+# Read the XML file, throws an error if the file is missing
+try: tree = ET.parse(os.path.dirname(os.path.abspath(__file__)) + '/ActorList.xml')
+except: PluginError("ERROR: File 'fast64_internal/oot/ActorList.xml' is missing.")
```
Dragorn421, line 6:
```suggestion
except: raise PluginError("ERROR: File 'fast64_internal/oot/ActorList.xml' is missing or malformed.")
```
I'm not sure what effect raising an error during module loading has, I'd expect prevent fast64 from loading but idk

fast64_internal/oot/oot_constants.py
```diff
@@ -1,3 +1,11 @@
+import os, xml.etree.ElementTree as ET
```
Dragorn421, line 1:
```suggestion
import os
import xml.etree.ElementTree as ET
```

fast64_internal/oot/oot_constants.py
```diff
@@ -44,449 +52,24 @@
...
-	("ACTOR_EN_DOOR", "EN_DOOR", "EN_DOOR"),
-	("ACTOR_DOOR_SHUTTER", "DOOR_SHUTTER", "DOOR_SHUTTER"),
-	("ACTOR_DOOR_WARP1", "DOOR_WARP1", "DOOR_WARP1"),
-	("ACTOR_DOOR_TOKI", "DOOR_TOKI", "DOOR_TOKI"),
-	("ACTOR_DOOR_ANA", "DOOR_ANA", "DOOR_ANA"),
-	("ACTOR_DOOR_GERUDO", "DOOR_GERUDO", "DOOR_GERUDO"),
-	("ACTOR_DOOR_KILLER", "DOOR_KILLER", "DOOR_KILLER"),
-]
+# Add every actor that is a transition actor (defined by the category 'ACTORCAT_DOOR')
+ootEnumTransitionActorID = [(actorNode.get('ID'), actorNode.get('Name'), \
```
Dragorn421, line 56: `actorNode.get('ID')` is the enum name (like `ACTOR_EN_DOOR`)
if you look at the blender api for EnumProperty, this as the first element of the tuple is what is actually saved in the blend
we can't change it as we need to keep compatibility with old blends, but it's something to keep in mind for when the enum names may change.

this means the xml should probably use "ID" or something else as something that will never ever change, and have another "EnumName" thing for the enum name, which may change. and currently both would be the same

fast64_internal/oot/oot_level.py
```diff
@@ -50,6 +50,24 @@ def handleRefreshDiffModelIDs(modelID):
...
+class OOT_UpgradeActors(bpy.types.Operator):
+	bl_idname = 'object.oot_upgrade_actors'
+	bl_label = "Upgrade Fast64 OoT actors data"
+	bl_options = {'REGISTER', 'UNDO', 'PRESET'}
+
+	def execute(self, context):
+		OOT_ObjectProperties.upgrade_changed_props()
+		return {'FINISHED'}
+
+def drawUpgrade(obj, layout):
```
Dragorn421, line 62:
```suggestion
def drawUpgradeActorData(obj, layout):
```

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +642,274 @@ def oot_utility_register():
...
+	indices = []
+	for actorNode in root:
+		if actorNode.get('Key') == actorKey:
+			for elem in actorNode:
+				if elem.tag == elemTag:
+					if flagType is None or (flagType == 'Switch' and elem.get('Type') == flagType):
+						indices.append(int(elem.get('Index'), base=10))
+	return max(indices) if indices else None
+
+def hasTiedParams(tiedParam, actorType):
```
Dragorn421, line 721:
```suggestion
def hasActorTiedParams(tiedParam, actorType):
```

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +642,274 @@ def oot_utility_register():
...
+	else:
+		attr = getattr(object, field, '0x0')
+		if isinstance(attr, str):
+			setattr(object, field, f'0x{(param & mask) >> shift:02X}')
+		elif isinstance(attr, bool):
+			setattr(object, field, bool((param & mask) >> shift))
+		else:
+			raise NotImplementedError
+
+def getLastElemIndex(actorKey, elemTag, flagType):
```
Dragorn421, line 710:
```suggestion
def getActorLastElemIndex(actorKey, elemTag, flagType):
```

fast64_internal/oot/oot_actor.py
Yanis42: yeah I'm already doing that (the get/set thing), that's why I didn't use ``getCustomPropName``

Dragorn421: it's needed for `userIDField` and in other spots though?

fast64_internal/oot/oot_constants.py
Yanis42: I don't get it, like older blends shouldn't have like ``ACTOR_EN_BOX`` as a transition actor right? letting the user choose from the whole list was a mistake as it's just odd to be able to choose actors that don't fall under the actor category "doors"

and the value only matters when you upgrade, so I don't really understand

fast64_internal/oot/oot_actor.py
Yanis42: yeah because it's (blender) properties that doesn't use get/set at all as it's not needed

Dragorn421: so why do these have a Custom variant at all then?

fast64_internal/oot/oot_constants.py
Dragorn421: in practice, imagine decomp changes `ACTOR_EN_DOOR` to `ACTOR_DOOR`
the xml would change `ID="ACTOR_EN_DOOR"` to `ID="ACTOR_DOOR"` right
but this ID value is being used as the "identifier" https://docs.blender.org/api/current/bpy.props.html#bpy.props.EnumProperty which is used to save the property in the .blend
so changing the ID would break old blends

Yanis42: how about using the actor keys (like 0x000A for en_box) instead? this should never change

Dragorn421: what about the existing blends though
and ideally we wouldn't want to hardcode the hex values, they may change in decomp mods

Yanis42: ok then keeping ``ID`` as the current decomp IDs and replace ``Key`` by what's been used as "ID" currently and never changing that ever (and remove the ``ACTOR_`` part of course)

Yanis42: (I just don't feel like finding a proper name for all of the ~430 actors in the game lol)

fast64_internal/oot/oot_constants.py
```diff
@@ -44,449 +55,29 @@
...
+# Old actor list, we can't delete this (for now) as it'd create compatibility issues
+ootEnumActorIDLegacy = [("ACTOR_" + (actorNode.get('Key').upper()), actorNode.get('Name'),
+					actorNode.get('ID').replace('ACTOR_','')) for actorNode in root if actorNode.get('ID') is not None]
+ootEnumActorIDLegacy.insert(0, ("Custom", "Custom Actor", "Custom"))
+
+ootChestContent = [(elem.get('Value'), elem.get('Name'),
+					elem.get('Name')) for listNode in root for elem in listNode if listNode.tag == 'List'
+					and listNode.get('Name') == 'Chest Content']
+
+ootNaviMsgID = [(elem.get('Value'), elem.get('Name'),
```
Dragorn421, line 78: the identifier in hex works but as for actors it isn't great if a mod ever shifts the values

------------

also why is `itemDrops`, same as this but with the `Collectibles` `<List>`, in `editOOTActorProperties` instead of here (they probably should all be together but idk)

actually why is Collectibles at all different from ChestContent

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,401 @@
...
+				elif elem.get('Type') == 'Switch':
+					genString(propAnnotations, actorKey, f'.switchFlag{i_flag}', 'Switch Flag')
+					i_flag += 1
+			elif elem.tag == 'Collectible':
+				genEnum(propAnnotations, actorKey, '.collectibleDrop', itemDrops, 'Collectible Drop')
+			elif elem.tag == 'Parameter':
+				actorTypeList = [(elem2.get('Params'), elem2.text, elem2.get('Params'))
+								for actorNode2 in root for elem2 in actorNode2
+								if actorNode2.get('Key') == actorKey and elem2.tag == 'Parameter']
+				genEnum(propAnnotations, actorKey, '.type', actorTypeList, 'Actor Type')
```
Dragorn421, line 254: so this and Collectible (and maybe others, unsure) assume only one of these can be used by actor? it's an okay assumption to make I guess but it should be enforced by checking (or even better allow several of these)

fast64_internal/oot/oot_constants.py
Yanis42: "collectibles" is whatever ``en_item00`` can give you, "ChestContent" is whatever items a chest can have, both are not the same

about ``itemDrops`` I guess I put it in ``editOOTActorProperties`` because it's only used there

Dragorn421: why is itemDrops only used in `editOOTActorProperties` with `genEnum`, while ChestContent is used by `OOT_SearchChestContentEnumOperator`?

fast64_internal/oot/oot_actor.py
Yanis42: I think this is the "added numbers to more props" part of the commit so I'm gonna resolve this, feel free to unresolve if I'm wrong

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,426 @@
...
+	naviMsgID : bpy.props.StringProperty(name='Navi Message ID', default = '0x00')
+
+	@staticmethod
+	def upgrade_object(obj):
+		if obj.data is None:
+			if (obj.ootEmptyType == 'Scene') or (obj.ootEmptyType == 'Room'):
+				print(f"Processing '{obj.name}'...")
+				for child in obj.children:
+					OOTActorProperties.upgrade_object(child)
+			elif ((obj.fast64.oot.version != obj.fast64.oot.cur_version) and
```
Dragorn421, line 101:
```suggestion
			elif ((obj.fast64.oot.version < obj.fast64.oot.cur_version) and
```

fast64_internal/oot/oot_utility.py
```diff
@@ -379,6 +380,35 @@ def getCustomProperty(data, prop):
...
+					if dpKey == actorNode.get('Key'):
+						for elem in actorNode:
+							target = elem.get('Target')
+							if hasActorTiedParams(elem.get('TiedParam'), actorType):
+								if target == field == 'XRot':
+									return getActorParameter(detailedProp, dpKey, 'XRot', None)
+								elif target == field == 'YRot':
+									return getActorParameter(detailedProp, dpKey, 'YRot', None)
+								elif target == field == 'ZRot':
+									return getActorParameter(detailedProp, dpKey, 'ZRot', None)
```
Dragorn421, line 406: afaict this is equivalent:
```suggestion
			elif field in {"XRot", "YRot", "ZRot"}:
				dpKey = detailedProp.actorKey
				actorType = getattr(detailedProp, dpKey + '.type', None)
				for actorNode in root:
					if dpKey == actorNode.get('Key'):
						for elem in actorNode:
							target = elem.get('Target')
							if hasActorTiedParams(elem.get('TiedParam'), actorType):
								if target == field:
									return getActorParameter(detailedProp, dpKey, field, None)
								
									
								
									
```
which makes me wonder why `getActorExportValue` is a single function since it splits most of its behavior by an argument, so it should in theory be better split into several functions?

Dragorn421:
error when clicking "Upgrade Data Now!" on a en_box actor with params 0x5020 added with `main`:
```
Error: Python: Traceback (most recent call last):
  File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_level.py", line 59, in execute
    OOT_ObjectProperties.upgrade_changed_props()
  File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_level.py", line 201, in upgrade_changed_props
    OOTActorProperties.upgrade_object(obj)
  File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_actor.py", line 103, in upgrade_object
    upgradeActorInit(obj)
  File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_utility.py", line 866, in upgradeActorInit
    upgradeActorProcess(objType, obj, entranceProp.actorID, obj.fast64.oot.actor,
  File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_utility.py", line 878, in upgradeActorProcess
    if ("ACTOR_" + dPKey.upper()) == actorID:
AttributeError: 'NoneType' object has no attribute 'upper'
```

fast64_internal/oot/oot_utility.py
Yanis42: I did the split, it makes more sense that way

Yanis42:
> error when clicking "Upgrade Data Now!" on a en_box actor with params 0x5020 added with `main`:
> 
> ```
> Error: Python: Traceback (most recent call last):
>   File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_level.py", line 59, in execute
>     OOT_ObjectProperties.upgrade_changed_props()
>   File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_level.py", line 201, in upgrade_changed_props
>     OOTActorProperties.upgrade_object(obj)
>   File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_actor.py", line 103, in upgrade_object
>     upgradeActorInit(obj)
>   File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_utility.py", line 866, in upgradeActorInit
>     upgradeActorProcess(objType, obj, entranceProp.actorID, obj.fast64.oot.actor,
>   File "%appdata%\Blender Foundation\Blender\2.93\scripts\addons\fast64\fast64_internal\oot\oot_utility.py", line 878, in upgradeActorProcess
>     if ("ACTOR_" + dPKey.upper()) == actorID:
> AttributeError: 'NoneType' object has no attribute 'upper'
> ```

This should be fixed now

fast64_internal/oot/oot_utility.py
```diff
@@ -602,3 +627,308 @@ def oot_utility_register():
...
+
+	for childObj in obj.children:
+		upgradeActorInit(childObj)
+
+def upgradeActorProcess(user, obj, actorID, detailedProp, params, paramField, paramTarget):
+	if not obj.ootEntranceProperty.customActor and actorID != 'Custom':
+		actorParams = 0
+		for actorNode in root:
+			dPKey = actorNode.get('Key')
+			if dPKey is not None and (("ACTOR_" + dPKey.upper()) == actorID):
```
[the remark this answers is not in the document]
Yanis42: I added this check to fix the bug you mentionned yesterday, ``actorNode.get('Key')`` can actually be none when it's going through ``<List>``

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,424 @@
...
+	transActorID : bpy.props.StringProperty(name='Transition Actor ID', default='ACTOR_EN_DOOR')
+	transActorKey : bpy.props.EnumProperty(name='Transition Actor ID', items=ootEnumTransitionActorID)
+	transActorParam : bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000',
+		get=lambda self: getValues(self, self.transActorKey, 'actorParam', 'Params', 'transActorParam'),
+		set=lambda self, value: setValues(self, value, 'Params', 'transActorID'))
+	transActorParamCustom : bpy.props.StringProperty(name = 'Actor Parameter', default = '0x0000')
+
+	# Other
+	itemChest : bpy.props.StringProperty(name='Chest Content', default = '0x48')
+	naviMsgID : bpy.props.StringProperty(name='Navi Message ID', default = '0x00')
```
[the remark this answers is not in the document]
Yanis42: I think the default being an integer is a mistake, when you use the search prop it sets the key

and using enums would be painful because of how large these are, it's faster to just type "heart" or "wake up!" (especially for navi messages)

fast64_internal/oot/oot_actor.py
```diff
@@ -1,47 +1,424 @@
...
 
 	def invoke(self, context, event):
 		context.window_manager.invoke_search_popup(self)
 		return {'RUNNING_MODAL'}
 
+class OOTActorProperties(bpy.types.PropertyGroup):
+	# Each prop has its 'custom' variant because of the get and set functions
+	# Actors/Entrance Actor
+	actorID : bpy.props.StringProperty(name='Actor ID', default='ACTOR_PLAYER')
+	actorKey : bpy.props.EnumProperty(name='Actor Key', items=ootEnumActorID)
```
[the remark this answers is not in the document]
Yanis42: I'll investigate later but iirc we actually need ``actorID`` and ``transActorID``

fast64_internal/oot/oot_utility.py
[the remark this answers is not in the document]
Dragorn421: then the better thing to do would be to skip `<List>` elements I guess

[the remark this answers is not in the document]
Yanis42: I'll change that then

Yanis42:
I think I'm done with adding/documenting/testing things in this PR, I used a blend with basically every actors to help me documenting some actors in the XML (used code to create it)

From now I won't work on this PR except for reviews, I've been working on that for 10 months now I want to move on something else

Zelllll (approved):
I think this is all very good now, I don’t have any objections to merging.

Yanis42:
I'm closing this PR for now, I want improve the code a lot so it doesn't make sense to keep this opened, also I made a [WIP tool](https://github.com/Yanis42/Cole) to get a standalone version of everything this PR adds (and more!)

also with the recent changes made to #95 I think it's fine to delay this for next weeks/months (again)

krm01:
this is a tragedy D:

Yanis42:
it will come back better and stronger 💪

PR #61, opened 2021-12-23
Support non folded skeletons

sauraen:
fast64 previously produced corrupted SkelAnime results when the mesh in Blender didn't have all the bones pointing the same direction in "rest position" (edit mode position). This was known as a "folded" skeleton because of the way imported meshes looked.

This issue was caused by the animation export code being overcomplicated and wrong. There is an explanation in the comments below of why the new version should work, so it's actually theoretically based, not just it happens to work.

This is tested and works with a complicated custom flex skel mesh which has many bones not pointing the same direction as well as with translation offsets. This has NOT yet been tested with creating custom animations on a mesh imported from the game. However, the changes made only impact the animation export codepath, not the skeleton / dlist export codepath, so I think they should work with vanilla skeletons.

I will mention in particular that there were several issues with `setArmatureToNonRotatedPose`, which is used on the skeleton / dlist export codepath, and which has been removed. So you might think this would affect those results. However, this code only changed pose bone rotations, and pose bones are always ignored by the skeleton / dlist export codepath. The latter also ignored the returned `restPoseRotations`. So, there are no changes to this codepath. The other issues were:
- the new pose set up is immediately overwritten when the scene animation frame is changed
- the rotations returned are just the edit-mode / "rest position" rotations of the bones, which are always accessible regardless of the pose. They are not, as it appears they might be, some sort of rotations built up incrementally from the parent bones.
- this function removed all animation drivers and constraints from the bones, and obviously if the user has applied these, they are needed for the results to be what the user created.

### DIFF
--- a/fast64_internal/oot/oot_anim.py
+++ b/fast64_internal/oot/oot_anim.py
@@ -55,7 +55,68 @@ class OOTAnimation:
 
 		return data
 
-def ootConvertAnimationData(anim, armatureObj, frameInterval, restPoseRotations, convertTransformMatrix):
+def ootGetAnimBoneRot(bone, poseBone, convertTransformMatrix, isRoot, frame, boneIndex):
+	# frame and boneIndex are just for debugging
+	# 
+	# OoT draws limbs like this:
+	# limbMatrix = parentLimbMatrix @ limbFixedTranslationMatrix @ animRotMatrix
+	# There is no separate rest position rotation; an animation rotation of 0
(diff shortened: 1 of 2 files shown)


sauraen:
I have tested this with custom animations on a vanilla skeleton and it appears to work.

Dragorn421 (approved):

thecozies (approved):
hopefully this can easily apply to sm64. if theres any parity i'll make some common functions for both to use

Zelllll (approved):

PR #63, opened 2021-12-24
Add limb name defines

sauraen:
Add defines for the limbs of a skeleton to the header, for example:
```
#define YOURMESH_ROOT_POS_LIMB 0
#define YOURMESH_ROOT_ROT_LIMB 1
#define YOURMESH_SOMETHING_LIMB 2
#define YOURMESH_FOO_LIMB 3
#define YOURMESH_NUM_LIMBS 4
```

### DIFF
--- a/fast64_internal/oot/oot_skeleton.py
+++ b/fast64_internal/oot/oot_skeleton.py
@@ -81,7 +81,6 @@ class OOTSkeleton():
 				str(self.getNumLimbs()) + " };\n\n" 
 			data.header = "extern SkeletonHeader " + self.name + ";\n"
 
-		index = 0
 		for limb in limbList:
 			name = (self.name + "_" + toAlnum(limb.boneName)).upper()
 			if limb.index == 0:
@@ -89,8 +88,7 @@ class OOTSkeleton():
 				data.header += "#define " + name + "_ROT_LIMB 1\n"
 			else:
 				data.header += "#define " + name + "_LIMB " + str(limb.index+1) + "\n"
-			index += 1
-		data.header += "#define " + self.name.upper() + "_NUM_LIMBS " + str(index+1) + "\n"
+		data.header += "#define " + self.name.upper() + "_NUM_LIMBS " + str(len(limbList)+1) + "\n"
 
 		limbData.append(data)
 
--- a/fast64_internal/oot/oot_skeleton.py
+++ b/fast64_internal/oot/oot_skeleton.py
@@ -84,7 +84,6 @@ class OOTSkeleton():
 		index = 0
 		for limb in limbList:
 			name = (self.name + "_" + toAlnum(limb.boneName)).upper()
-			assert limb.index == index
 			if limb.index == 0:
 				data.header += "#define " + name + "_POS_LIMB 0\n"
 				data.header += "#define " + name + "_ROT_LIMB 1\n"
@@ -163,8 +162,11 @@ class OOTLimb():
 					return True
 			return False
 
-	# should be same order as ootProcessBone
 	def getList(self, limbList):
+		# Like ootProcessBone, this must be in depth-first order to match the
+		# OoT SkelAnime draw code, so the bones are listed in the file in the
+		# same order as they are drawn. This is needed to enable the programmer
+		# to get the limb indices and to enable optimization between limbs.
 		limbList.append(self)
 		for child in self.children:
 			child.getList(limbList)
@@ -363,6 +365,10 @@ def ootProcessBone(fModel, boneName, parentLimb, nextIndex, meshObj, armatureObj
 	limb.isFlex = hasSkinnedFaces
 	nextIndex += 1
 
+	# This must be in depth-first order to match the OoT SkelAnime draw code, so
+	# the bones are listed in the file in the same order as they are drawn. This
+	# is needed to enable the programmer to get the limb indices and to enable
+	# optimization between limbs.
 	childrenNames = getSortedChildren(armatureObj, bone)
 	for childName in childrenNames:
 		nextIndex, lastMaterialName = ootProcessBone(fModel, childName, limb, nextIndex, meshObj, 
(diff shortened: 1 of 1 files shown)


fast64_internal/oot/oot_skeleton.py
```diff
@@ -81,13 +81,26 @@ def toC(self):
...
+
 		limbData.append(data)
 
 		return limbData
 
 class OOTLimb():
-	def __init__(self, skeletonName, index, translation, DL, lodDL):
+	def __init__(self, skeletonName, boneName, index, translation, DL, lodDL):
 		self.skeletonName = skeletonName
+		self.boneName = boneName
```
Dragorn421, line 103: I think `boneName` can contain invalid characters for a C identifier at that point, like a leading digit, spaces...

sauraen:
Good catch, fixed.

fast64_internal/oot/oot_skeleton.py
```diff
@@ -81,13 +81,26 @@ def toC(self):
 				str(self.getNumLimbs()) + " };\n\n" 
 			data.header = "extern SkeletonHeader " + self.name + ";\n"
 
+		index = 0
+		for limb in limbList:
+			name = (self.name + "_" + toAlnum(limb.boneName)).upper()
+			assert limb.index == index
```
Dragorn421, line 87: So, the defines need to follow the order the limbs draw in, which is not necessarily the order they have to be (for the game to work) in `limbList` (aka the `limb.index` value) (`SkeletonHeader#segment`), right?

In this assert, both `index` (order in `limbList`) and `limb.index` (order of exploration by `ootProcessBone`) could in theory not be in the order limbs are drawn in (for example, if `createLimbList` and `ootProcessBone` did breadth-first-search instead of depth-first)

So if this assert is supposed to ensure the indices correspond to the drawing order, it's flawed, I would just remove it, and add a comment to `OOTSkeleton#createLimbList`/`OOTLimb#getList` instead, along the lines of "this needs to be depth-first like the oot skelanime draw code, so the bones are in the order they're drawn in. this is needed for the limb indices defines"

A similar comment may be added to `ootProcessBone` since its traversal should be the same, as "documented" in `OOTLimb#getList`

sauraen:
Sorry for taking so long to get to this.

There was already a comment mentioning the order requirement on `OOTLimb.getList`. `OOTSkeleton.createLimbList` does not itself do any ordering. Depth-first order is not merely required for this define feature, but so that the programmer can get the limb index at all without manually doing the depth-first traversal in their head. It's also required for the materials optimizer which is already merged, as well as the large-scale flex skeleton optimizer I'm working on.

I have updated the comment on `OOTLimb.getList`, added a similar comment to `ootProcessBone`, and removed the assertion as you have requested.

While more documentation is always a good thing, I disagree that the assertion was flawed. The assertion checked that a condition which the code was assuming was in fact true. If someone decided to change `ootProcessBone` and `OOTLimb.getList` to breadth-first or something else for some reason, they would hit this assertion and have to either abandon their changes, or fix the defines. The only "flaw" with this approach is that there should be another assertion in the code for the materials optimizer checking the same condition, because this also depends on it.

fast64_internal/oot/oot_skeleton.py
```diff
@@ -81,13 +81,25 @@ def toC(self):
...
+		index = 0
+		for limb in limbList:
+			name = (self.name + "_" + toAlnum(limb.boneName)).upper()
+			if limb.index == 0:
+				data.header += "#define " + name + "_POS_LIMB 0\n"
+				data.header += "#define " + name + "_ROT_LIMB 1\n"
+			else:
+				data.header += "#define " + name + "_LIMB " + str(limb.index+1) + "\n"
+			index += 1
+		data.header += "#define " + self.name.upper() + "_NUM_LIMBS " + str(index+1) + "\n"
```
Dragorn421, line 93: you can remove `index` and
```suggestion
		data.header += "#define " + self.name.upper() + "_NUM_LIMBS " + str(len(limbList)+1) + "\n"
```

then is this really _num_limbs or _frame_table_length, what do you intend this define as

Dragorn421 (commented):
We're on the same page about the assert I refered to I think, I do think it makes sense to check basically `limbList[limb.index] == limb`, just that it isn't the proper check to ensure the indices are correct

Like, if you wanted to add the asserts back I would put it in its own loop right after `limbList = self.createLimbList()`

sauraen:
I still don't understand why the assertion was incorrect, but it's no longer included so it should not be considered a blocker to merging these changes.

I have removed the index counter.

The define actually should just be `NUM_LIMBS`, not `SKELNAME_NUM_LIMBS`, but for clarity I am using the latter. `NUM_LIMBS` is a define in actors which represents the frame table length, or the number of "animation channels", where root is two. Instead of having this be defined within the actor C file, I thought it would be good to move it to the object header.

Zelllll (approved):

Dragorn421 (approved):
so the num_limbs define is meant as `SkelAnime_Init`'s `limbCount` argument alright
(the decomp seems to be wrong about having that argument named `limbCount` then, looks more like `frameTableLength`, I'll investigate eventually)

sauraen:
> so the num_limbs define is meant as `SkelAnime_Init`'s `limbCount` argument alright (the decomp seems to be wrong about having that argument named `limbCount` then, looks more like `frameTableLength`, I'll investigate eventually)

It looks like the decomp currently calls this `SKELNAME_LIMB_MAX`, for example:
```
typedef enum {
    /* 0x00 */ IMPA_LIMB_NONE,
    /* 0x01 */ IMPA_LIMB_ROOT,
    /* 0x02 */ IMPA_LIMB_TORSO,
    ...
    /* 0x10 */ IMPA_LIMB_HEAD,
    /* 0x11 */ IMPA_LIMB_MAX
} ImpaLimb;

typedef struct DemoIm {
    /* 0x0000 */ Actor actor;
    /* 0x014C */ SkelAnime skelAnime;
    /* 0x0190 */ Vec3s jointTable[IMPA_LIMB_MAX];
    /* 0x01F6 */ Vec3s morphTable[IMPA_LIMB_MAX];
```

which is slightly better than `SKELNAME_NUM_LIMBS`, but not by too much, since the first two are actually not "none" and "root", but "root translation" and "root rotation". Also, this information is still stored in the *actor* header, whereas it's actually a property of the skeleton in the object file.

There's two acceptances so I'm gonna merge this now; when these issues are corrected in decomp, then we can align fast64 to match it.

PR #65, opened 2021-12-31
Fix scene header confusion bug introduced in multi-file export changes

sauraen:
The code for making a list of all the cutscenes to be included in a scene was wrong, it ignored the main header and then exported duplicates of the main header cutscene for all other headers.

In the process I also realized that the "extra cutscenes" feature--to enable cutscene data to be included in a scene file, but not referenced from any scene header--could be used from any of the alternate scene headers. This is nonsense because the whole point is that these cutscenes are not associated with any header. So I've added code to ensure this feature is only usable from the main scene.


Zelllll (approved):
thanks for fixing my dumb mistakes :)

PR #68, opened 2022-01-10
Make Fast64 use LoadGeometryMode when Write All material mode is used

Zelllll:
This PR aims to simplify material output when the Write All material method is used.
```
gsSPSetGeometryMode(G_SHADE | G_CULL_FRONT | G_SHADING_SMOOTH),
gsSPClearGeometryMode(G_ZBUFFER | G_CULL_BACK | G_FOG | G_LIGHTING | G_TEXTURE_GEN | G_TEXTURE_GEN_LINEAR),
```
now becomes
```
gsSPLoadGeometryMode(G_SHADE | G_CULL_FRONT | G_SHADING_SMOOTH),
```

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -3768,6 +3768,37 @@ def geoFlagListToWord(flagList, f3d):
 	
 	return word
 
+class SPGeometryMode:
+	# Note: SPGeometryMode can only be used with GBI 2
+	def __init__(self, clearFlagList, setFlagList):
+		self.clearFlagList = clearFlagList
+		self.setFlagList = setFlagList
+    
+	def to_binary(self, f3d, segments):
+		wordClear = geoFlagListToWord(self.clearFlagList, f3d)
+		wordSet = geoFlagListToWord(self.setFlagList, f3d)
+
+		return gsSPGeometryMode_F3DEX_GBI_2(wordClear, wordSet, f3d)
+
+	def to_c(self, static = True):
+		data = 'gsSPGeometryMode(' if static else \
+			'gSPGeometryMode(glistp++, '
+		data += ((' | '.join(self.clearFlaglist)) if len(self.clearFlagList) > 0 else '0') + ', '
+		data += ((' | '.join(self.setFlagList)) if len(self.setFlagList) > 0 else '0') + ')'
+		return data
+
+	def to_sm64_decomp_s(self):
+		data = 'gsSPGeometryMode '
+		for flag in self.clearFlagList:
+			data += flag + ' | '
+		data = data[:-3] + ', '
+		for flag in self.setFlagList:
+			data += flag + ' | '
+		return data[:-3]
+	
+	def size(self, f3d):
+		return GFX_SIZE
+
 class SPSetGeometryMode:
 	def __init__(self, flagList):
 		self.flagList = flagList
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -3769,16 +3769,18 @@ def geoFlagListToWord(flagList, f3d):
 	return word
 
 class SPGeometryMode:
-	# Note: SPGeometryMode can only be used with GBI 2
 	def __init__(self, clearFlagList, setFlagList):
 		self.clearFlagList = clearFlagList
 		self.setFlagList = setFlagList
     
 	def to_binary(self, f3d, segments):
-		wordClear = geoFlagListToWord(self.clearFlagList, f3d)
-		wordSet = geoFlagListToWord(self.setFlagList, f3d)
+		if f3d.F3DEX_GBI_2:
+			wordClear = geoFlagListToWord(self.clearFlagList, f3d)
+			wordSet = geoFlagListToWord(self.setFlagList, f3d)
 
-		return gsSPGeometryMode_F3DEX_GBI_2(wordClear, wordSet, f3d)
+			return gsSPGeometryMode_F3DEX_GBI_2(wordClear, wordSet, f3d)
+		else:
+			raise PluginError("LoadGeometryMode only available in F3DEX_GBI_2.")
 
 	def to_c(self, static = True):
 		data = 'gsSPGeometryMode(' if static else \
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -3780,7 +3780,7 @@ class SPGeometryMode:
 
 			return gsSPGeometryMode_F3DEX_GBI_2(wordClear, wordSet, f3d)
 		else:
-			raise PluginError("LoadGeometryMode only available in F3DEX_GBI_2.")
+			raise PluginError("GeometryMode only available in F3DEX_GBI_2.")
 
 	def to_c(self, static = True):
 		data = 'gsSPGeometryMode(' if static else \
--- a/fast64_internal/f3d/f3d_writer.py
+++ b/fast64_internal/f3d/f3d_writer.py
@@ -1248,7 +1248,7 @@ def saveOrGetF3DMaterial(material, fModel, obj, drawLayer, convertTextureData):
 
 	defaults = bpy.context.scene.world.rdp_defaults
 	if bpy.context.scene.f3d_type == 'F3DEX_GBI_2':
-		saveGeoModeDefinitionGBI2(fMaterial, f3dMat.rdp_settings, defaults, fModel.matWriteMethod)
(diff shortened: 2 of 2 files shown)


fast64_internal/f3d/f3d_writer.py
```diff
@@ -1247,7 +1247,10 @@ def saveOrGetF3DMaterial(material, fModel, obj, drawLayer, convertTextureData):
 		defaultRM = None
 
 	defaults = bpy.context.scene.world.rdp_defaults
-	saveGeoModeDefinition(fMaterial, f3dMat.rdp_settings, defaults, fModel.matWriteMethod)
+	if (fModel.matWriteMethod == GfxMatWriteMethod.WriteAll):
+		saveGeoModeDefinition(fMaterial, f3dMat.rdp_settings, defaults, fModel.matWriteMethod)
+	else:
+		saveGeoModeDefinition(fMaterial, f3dMat.rdp_settings, defaults, fModel.matWriteMethod)
```
Dragorn421, line 1253: afaict there's no point to this conditional

fast64_internal/f3d/f3d_writer.py
```diff
@@ -2028,11 +2031,14 @@ def saveBitGeo(value, defaultValue, flagName, setGeo, clearGeo, matWriteMethod):
...
 		if value:
 			setGeo.flagList.append(flagName)
-		else:
+		elif matWriteMethod != GfxMatWriteMethod.WriteAll:
 			clearGeo.flagList.append(flagName)
 
 def saveGeoModeDefinition(fMaterial, settings, defaults, matWriteMethod):
-	setGeo = SPSetGeometryMode([])
+	if matWriteMethod == GfxMatWriteMethod.WriteAll:
+		setGeo = SPLoadGeometryMode([])
```
Dragorn421, line 2039: `SPLoadGeometryMode` is only in f3dex2, so this should actually be made f3dex2-specific somehow

(and if using f3dex2 it would be better to use `gsSPGeometryMode(clear, set)` in both write methods, but that would be more involved with regards to fast64 code)

Zelllll:
I've now changed this PR to introduce SPGeometryMode for GBI2. SM64 people should definitely test this.

fast64_internal/f3d/f3d_writer.py
```diff
@@ -2024,6 +2027,53 @@ def normToSigned8Vector(normal):
...
+	elif len(geo.setFlagList) == 0:
+		geo.setFlagList.append('0')
+
+	if matWriteMethod == GfxMatWriteMethod.WriteAll:
+		fMaterial.material.commands.append(SPLoadGeometryMode(geo.setFlagList))
+	else:
+		fMaterial.material.commands.append(geo)
+
+	if matWriteMethod == GfxMatWriteMethod.WriteDifferingAndRevert:
+		fMaterial.revert.commands.append(SPGeometryMode(geo.setFlagList, geo.clearFlagList))
```
Dragorn421, line 2075: I'm pretty sure the only check on the mat write method that is needed is in `saveBitGeoGBI2`
```suggestion
	fMaterial.material.commands.append(geo)
```

fast64_internal/f3d/f3d_writer.py
```diff
@@ -2024,6 +2027,53 @@ def normToSigned8Vector(normal):
...
+
+	# make sure normals are saved correctly.
+	saveBitGeoGBI2(settings.g_tex_gen, defaults.g_tex_gen, 'G_TEXTURE_GEN',
+		geo, matWriteMethod)
+	saveBitGeoGBI2(settings.g_tex_gen_linear, defaults.g_tex_gen_linear,
+		'G_TEXTURE_GEN_LINEAR', geo, matWriteMethod)
+	saveBitGeoGBI2(settings.g_shade_smooth, defaults.g_shade_smooth,
+		'G_SHADING_SMOOTH', geo, matWriteMethod)
+	if bpy.context.scene.f3d_type == 'F3DEX_GBI_2' or \
+		bpy.context.scene.f3d_type == 'F3DEX_GBI':
```
Dragorn421, line 2060: this condition is always true in this function

fast64_internal/f3d/f3d_writer.py
```diff
@@ -1247,7 +1247,10 @@ def saveOrGetF3DMaterial(material, fModel, obj, drawLayer, convertTextureData):
 		defaultRM = None
 
 	defaults = bpy.context.scene.world.rdp_defaults
-	saveGeoModeDefinition(fMaterial, f3dMat.rdp_settings, defaults, fModel.matWriteMethod)
+	if bpy.context.scene.f3d_type == 'F3DEX_GBI_2':
+		saveGeoModeDefinitionGBI2(fMaterial, f3dMat.rdp_settings, defaults, fModel.matWriteMethod)
```
Dragorn421, line 1251: I would call it f3dex2 instead of gbi2 but I'm not very familiar with microcodes so not so sure

Dragorn421 (commented):
should be fine with sm64 since this only adds code in the case of f3dex2, but yeah making sure wouldn't hurt

fast64_internal/f3d/f3d_writer.py
Zelllll: Yeah I'm not actually sure why it is like that, this function is just a modified version of another function, and it was like that originally.

Dragorn421: well `saveGeoModeDefinition` deals with many microcodes, `saveGeoModeDefinitionGBI2` only with f3dex2

according to oot decomp's gbi.h `G_CLIPPING` only exists in f3dex/f3dex2 so that's why

fast64_internal/f3d/f3d_writer.py
Zelllll: I don't think this would be equivalent. I'll explain:

I check if WriteAll is used first, since that means that all flags will either be set or cleared, so LoadGeometryMode can be used in all cases for that scenario, however if Differing and Revert is used then that is not the case. I am now realizing that I can still clean this code up a little bit here though.

fast64_internal/f3d/f3d_writer.py
```diff
@@ -2024,6 +2027,49 @@ def normToSigned8Vector(normal):
...
+	if len(geo.clearFlagList) == 0:
+		geo.clearFlagList.append('0')
+	elif len(geo.setFlagList) == 0:
+		geo.setFlagList.append('0')
+
+	if matWriteMethod == GfxMatWriteMethod.WriteAll:
+		fMaterial.material.commands.append(SPLoadGeometryMode(geo.setFlagList))
+	else:
+		fMaterial.material.commands.append(geo)
+		fMaterial.revert.commands.append(SPGeometryMode(geo.setFlagList, geo.clearFlagList))
```
Dragorn421, line 2071: yeah that's more understandable, I missed the revert before
still I'm pretty sure this can just be:
```suggestion
	fMaterial.material.commands.append(geo)
	if matWriteMethod == GfxMatWriteMethod.WriteDifferingAndRevert:
		fMaterial.revert.commands.append(SPGeometryMode(geo.setFlagList, geo.clearFlagList))
```

also shouldn't the last line be
```py
		fMaterial.revert.commands.append(SPGeometryMode(geo.clearFlagList, geo.setFlagList))
```
(notice the swap of clear/set)

Zelllll: the swap is intentional, since its supposed to be the inverse

Zelllll: also while what you are saying is correct, I want to use LoadGeometryMode when possible

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3768,6 +3768,39 @@ def geoFlagListToWord(flagList, f3d):
...
+		self.setFlagList = setFlagList
+    
+	def to_binary(self, f3d, segments):
+		if f3d.F3DEX_GBI_2:
+			wordClear = geoFlagListToWord(self.clearFlagList, f3d)
+			wordSet = geoFlagListToWord(self.setFlagList, f3d)
+
+			return gsSPGeometryMode_F3DEX_GBI_2(wordClear, wordSet, f3d)
+		else:
+			raise PluginError("LoadGeometryMode only available in F3DEX_GBI_2.")
```
Dragorn421, line 3783:
```suggestion
			raise PluginError("GeometryMode only available in F3DEX_GBI_2.")
```

Dragorn421 (changes requested):

fast64_internal/f3d/f3d_writer.py
Zelllll: I guess someone else can weigh in, personally I think it's fine. Thoughts? @thecozies

Dragorn421 (approved):

PR #69, opened 2022-01-16
deprecate combined_params - upgrade all objects in blend

thecozies:
[issue 1: only objects in the first loaded scene were getting updated](https://canary.discord.com/channels/874816709855440926/874816710639767585/931351721232437249)
i was originally under the impression that changing scenes would run the persistent handlers, but its only when first loading a blend. so i changed iterating `bpy.context.scene.objects` to `bpy.data.objects`

[issue 2: combined bparams not matching the individuals](https://canary.discord.com/channels/874816709855440926/874816710639767585/931348767792332830)
My solution is to just _always_ derive combined bparams instead of storing it to another field

### DIFF
--- a/fast64_internal/sm64/sm64_objects.py
+++ b/fast64_internal/sm64/sm64_objects.py
@@ -1509,14 +1509,6 @@ class SM64_GameObjectProperties(bpy.types.PropertyGroup):
 		if (game_object.bparams != DEFAULT_BEHAVIOR_PARAMS):
 			game_object.use_individual_params = False
 
-	@staticmethod
-	def remove_unused_props(obj: bpy.types.Object):
-		game_object: SM64_GameObjectProperties = obj.fast64.sm64.game_object
-
-  		# replaced "combined_bparams" with method `get_combined_bparams`
-		if "combined_bparams" in game_object:
-			del game_object["combined_bparams"]
-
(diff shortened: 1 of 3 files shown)


fast64_internal/sm64/sm64_objects.py
```diff
@@ -1507,14 +1501,22 @@ def upgrade_object(obj):
...
 		if (game_object.bparams != DEFAULT_BEHAVIOR_PARAMS):
 			game_object.use_individual_params = False
 
+	@staticmethod
+	def remove_unused_props(obj: bpy.types.Object):
+		game_object: SM64_GameObjectProperties = obj.fast64.sm64.game_object
+
+  		# replaced "combined_bparams" with method `get_combined_bparams`
+		if "combined_bparams" in game_object:
+			del game_object["combined_bparams"]
```
Dragorn421, line 1518: imo there is no need to worry about such cleanup. code for upgrading is already a burden to maintain, I don't think we need code dealing with removed props

I think simpler upgrade code is worth the slight blend file bloat, thoughts?

thecozies: yeah thats a good point! dropping it

Dragorn421 (approved):
fingers crossed no more issues

PR #71, opened 2022-01-29
[OoT] Add option to write a dummy room list

Dragorn421:
This adds an option to OoT scene objects:
<details>

image

</details>

Diff between exporting with the option disabled and with the option enabled:
```diff
 	{ ACTOR_PLAYER, 0, -120, 0, 0, 0, 0, 0x0FFF },
 };
 
-extern u8 _spot03_room_0SegmentRomStart[];
-extern u8 _spot03_room_0SegmentRomEnd[];
-
+// Dummy room list
 RomFile spot03_scene_roomList[] = {
-	{ (u32)_spot03_room_0SegmentRomStart, (u32)_spot03_room_0SegmentRomEnd },
+	{0, 0},
 };
 
 EntranceEntry spot03_scene_header00_entranceList[] = {
```

The intent is to allow compiling the files separately (without linking) and then using the compiled files in tools such as zzrtl.

--------

The PR also introduces `Object.fast64.oot` and `Object.fast64.oot.scene`. Hopefully it doesn't break anything, I don't have much testing material

### DIFF
--- a/fast64_internal/oot/oot_level.py
+++ b/fast64_internal/oot/oot_level.py
@@ -89,6 +89,7 @@ class OOTObjectPanel(bpy.types.Panel):
 			drawSceneHeaderProperty(box, obj.ootSceneHeader, None, None, objName)
 			if obj.ootSceneHeader.menuTab == 'Alternate':
 				drawAlternateSceneHeaderProperty(box, obj.ootAlternateSceneHeaders, objName)
+			box.prop(obj.fast64.oot.scene, "write_dummy_room_list")
 
 		elif obj.ootEmptyType == 'Room':
 			drawRoomHeaderProperty(box, obj.ootRoomHeader, None, None, objName)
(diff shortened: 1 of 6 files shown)


fast64_internal/oot/oot_level.py
```diff
@@ -166,7 +167,16 @@ def onUpdateOOTEmptyType(self, context):
...
+class OOT_ObjectProperties(bpy.types.PropertyGroup):
+	version: bpy.props.IntProperty(name="OOT_ObjectProperties Version", default=0)
+	cur_version = 0 # version after property migration
+
+	scene: bpy.props.PointerProperty(type=OOTSceneProperties)
+
 oot_obj_classes = (
+	OOTSceneProperties,
+	OOT_ObjectProperties,
+
```
Zelllll, line 179: extra new line

Dragorn421: that's on purpose to separate those property groups from whatever is below, it looks somewhat organized somehow

Zelllll: fair enough, how come it's OOT_ though but then only OOT for the new one? Might be better to make it consistent

Dragorn421: There is `Fast64_Properties`, `SM64_Properties`, `OOT_Properties`, `Fast64_ObjectProperties`, `SM64_ObjectProperties`, ...
So for consistency with the SM64 counterpart at least (`SM64_ObjectProperties`) I chose `OOT_ObjectProperties`

Ideally `OOTSceneProperties` would be `OOT_SceneProperties` but then per this whole naming scheme one may mistakenly think "Scene" is `bpy.types.Scene`, not a OoT scene
Simply removing the underscore doesn't entirely address it :/

Zelllll (approved):

fast64_internal/oot/oot_level.py
Dragorn421: classes can be renamed without compatibility issues for blend files so let's just go with this and rename later if someone has a better idea

PR #76, opened 2022-02-18
scene table error fix

AriaHiro64:
decomp no longer touches src/code/z_scene_table.c anymore


Dragorn421:
Do not merge

This branch could be used as a temporary workaround for anyone having issues with oot's scene table, but it's not a proper fix

PR #77, opened 2022-02-20
Added import all DLs feature

sauraen:
This adds an option to the OoT DL importer which will import all DLs in the input file. This is a hacky alternative to proper scene import support, but that is much larger in scope. However, this is much better than previously, when importing a scene meant grepping for DLs in the scene file and one by one entering their names into Blender. (Importing the DLs from a scene is needed so that you can see the scene when creating new cutscenes in vanilla scenes.)

### DIFF
--- a/fast64_internal/f3d/f3d_parser.py
+++ b/fast64_internal/f3d/f3d_parser.py
@@ -1285,7 +1285,7 @@ class F3DContext:
 			else:
 				image.pixels[4*i : 4*(i+1)] = newValues
 
-	def processCommands(self, dlData, dlName, dlCommands):
+	def processCommands(self, dlData, dlName, dlCommands, ignoreCalls):
 		callStack = [F3DParsedCommands(dlName, dlCommands, 0)]
 		while len(callStack) > 0:
 			currentCommandList = callStack[-1]
@@ -1309,7 +1309,7 @@ class F3DContext:
 				self.addTriangle(command.params[0:3], dlData)
 			elif command.name == 'gsSP2Triangles':
 				self.addTriangle(command.params[0:3] + command.params[4:7], dlData)
-			elif command.name == 'gsSPDisplayList' or command.name[:10] == 'gsSPBranch':
+			elif (command.name == 'gsSPDisplayList' or command.name[:10] == 'gsSPBranch') and not ignoreCalls:
 				newDLName = self.processDLName(command.params[0])
 				if newDLName is not None:
 					newDLCommands = parseDLData(dlData, newDLName)
@@ -1573,7 +1573,7 @@ class ParsedMacro:
 # we distinguish these because there is no guarantee of bone order in blender,
 # so we usually rely on alphabetical naming.
 # This means changing the c variable names.
-def parseF3D(dlData, dlName, obj, transformMatrix, limbName, boneName, drawLayerPropName, drawLayer, f3dContext):
+def parseF3D(dlData, dlName, obj, transformMatrix, limbName, boneName, drawLayerPropName, drawLayer, f3dContext, ignoreCalls):
 
 	f3dContext.matrixData[limbName] = transformMatrix
 	f3dContext.setCurrentTransform(limbName)
@@ -1584,10 +1584,10 @@ def parseF3D(dlData, dlName, obj, transformMatrix, limbName, boneName, drawLayer
 	#groupIndex = vertexGroup.index
 
 	dlCommands = parseDLData(dlData, dlName)
-	f3dContext.processCommands(dlData, dlName, dlCommands)
+	f3dContext.processCommands(dlData, dlName, dlCommands, ignoreCalls)
 
 def parseDLData(dlData, dlName):
-	matchResult = re.search("Gfx\s*" + re.escape(dlName) + "\s*\[\s*\w*\s*\]\s*=\s*\{([^\}]*)\}", dlData)
+	matchResult = re.search("Gfx\s+" + re.escape(dlName) + "\s*\[\s*\w*\s*\]\s*=\s*\{([^\}]*)\}", dlData)
 	if matchResult is None:
 		raise PluginError("Cannot find display list named " + dlName)
 
@@ -1848,9 +1848,7 @@ def getImportData(filepaths):
 
 	return data
 
-def importMeshC(filepaths, name, scale, removeDoubles, importNormals, drawLayer, f3dContext):
-	data = getImportData(filepaths)
-
+def importMeshC(data, name, scale, removeDoubles, importNormals, drawLayer, f3dContext, ignoreCalls):
 	# Create new skinned mesh
 	mesh = bpy.data.meshes.new(name + '_mesh')
 	obj = bpy.data.objects.new(name + '_mesh', mesh)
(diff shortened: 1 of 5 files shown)


fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -227,16 +227,23 @@ def execute(self, context):
...
 				filepaths.append(os.path.join(bpy.context.scene.ootDecompPath, "assets/objects/gameplay_keep/gameplay_keep.c"))
 
-			importMeshC(filepaths, name, scale, removeDoubles, importNormals, drawLayer,
-				OOTF3DContext(F3D("F3DEX2/LX2", False), [name], basePath))
+			data = getImportData(filepaths)
+			dlNames = findAllDLs(data) if importAll else [name]
+			for n in dlNames:
+				print('Importing DL ' + n)
+				importMeshC(data, n, scale, removeDoubles, importNormals, drawLayer,
+					OOTF3DContext(F3D("F3DEX2/LX2", False), [n], basePath), importAll)
```
Dragorn421, line 246:
```suggestion
					OOTF3DContext(F3D("F3DEX2/LX2", False), [n], basePath), ignoreCalls)
```
with `ignoreCalls = importAll` somewhere, would be less confusing imo

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -227,16 +227,23 @@ def execute(self, context):
...
 			basePath = bpy.path.abspath(context.scene.ootDecompPath)
+			importAll = context.scene.ootDLImportAll
 			removeDoubles = context.scene.ootDLRemoveDoubles
 			importNormals = context.scene.ootDLImportNormals
 			drawLayer = bpy.context.scene.ootDLImportDrawLayer
 
 			filepaths = [ootGetObjectPath(isCustomImport, importPath, folderName)]
 			if not isCustomImport:
+				if importAll:
+					raise PluginError("Import All DLs requires custom import path!")
```
Dragorn421, line 238: I can't make sense of `ootGetObjectPath` 😅 why does importAll need isCustomImport

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -316,14 +323,18 @@ def draw(self, context):
...
 		if context.scene.ootDLImportUseCustomPath:
 			prop_split(col, context.scene, 'ootDLImportCustomPath', "File")
 		else:		
 			prop_split(col, context.scene, 'ootDLImportFolderName', "Object")
 		prop_split(col, context.scene, "ootDLImportDrawLayer", "Import Draw Layer")
+		if context.scene.ootDLImportAll:
+			col.label(icon = 'INFO', text = 'May have to manually fix layers after.')
 
 		col.prop(context.scene, "ootDLImportUseCustomPath")
+		col.prop(context.scene, "ootDLImportAll")
```
Dragorn421, line 337: it's not obvious that info line has to do with ootDLImportAll imo, I'd suggest
```suggestion
		col.prop(context.scene, "ootDLImportUseCustomPath")
		if context.scene.ootDLImportAll:
			box = col.box()
			box.prop(context.scene, "ootDLImportAll")
			box.label(icon = 'INFO', text = 'May have to manually fix layers after.')
		else:
			col.prop(context.scene, "ootDLImportAll")
```

fast64_internal/oot/oot_f3d_writer.py
```diff
@@ -495,6 +506,7 @@ def oot_dl_writer_register():
 	bpy.types.Scene.ootDLImportUseCustomPath = bpy.props.BoolProperty(
 		name = "Use Custom Path")
 
+	bpy.types.Scene.ootDLImportAll = bpy.props.BoolProperty(name = "Import All DLs", default = False)
```
Dragorn421, line 509: add this to `OOT_Properties` instead, and access it like `scene.fast64.oot.importAllDL` (or whatever name you like for `importAllDL`)

Dragorn421 (commented):
Nice

fast64_internal/oot/oot_f3d_writer.py
sauraen: Why? This is set up exactly the same way as all the other controls for importing and exporting DLs (and the other import/export tools too). Why should this one be handled differently?

fast64_internal/oot/oot_f3d_writer.py
sauraen: A *non*-custom import means fast64 will look for the DL name in `gameplay_keep` (and possibly `gameplay_field_keep` or `gameplay_dangeon_keep`) as well as in object files. So first of all, using non-custom import for scene import won't work because the scene DLs aren't in the object files. And second, import all with a non-custom import will import every DL in `gameplay_keep` (and possibly more)!

fast64_internal/oot/oot_f3d_writer.py
sauraen: I agree, however I'll try to do the hierarchical importing instead, so there won't be a need for ignoreCalls at all.

fast64_internal/f3d/f3d_parser.py
```diff
@@ -1309,7 +1309,7 @@ def processCommands(self, dlData, dlName, dlCommands):
 				self.addTriangle(command.params[0:3], dlData)
 			elif command.name == 'gsSP2Triangles':
 				self.addTriangle(command.params[0:3] + command.params[4:7], dlData)
-			elif command.name == 'gsSPDisplayList' or command.name[:10] == 'gsSPBranch':
+			elif (command.name == 'gsSPDisplayList' or command.name[:10] == 'gsSPBranch') and not ignoreCalls:
```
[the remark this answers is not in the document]
sauraen: You are right, I did this because in some scenes it was importing the parent objects and then also the child objects, so there were duplicates. Instead of doing a full hierarchy, I'll try to have it take the current list of DLs and then exclude any DLs which are called by another DL in the same file.

fast64_internal/oot/oot_f3d_writer.py
sauraen: I disagree, the warning is a warning about the layer selection box (that there's no way to set that box which will be fully correct), that's why I put it below that box. I think the fact that that warning only appears when you check the import all box is enough indication to the user that it's a result of that. The text could be better but it'll get cut off depending on how wide the user keeps the pane.

fast64_internal/oot/oot_f3d_writer.py
Dragorn421: We only introduced the .fast64 property groups recently so old ones haven't been migrated (yet), but new ones should go there

Why did we introduce these property groups, because it's better practice to group custom properties, less chance for name clashes and more organized


Not that I like having what's essentially an operator property as a scene property at all anyway but that's probably too late to change

fast64_internal/oot/oot_f3d_writer.py
Dragorn421: You could expand the explanation in the description for the `ootDLImportAll` property, and redirect to it here maybe

fast64_internal/oot/oot_f3d_writer.py
Dragorn421: I may be a bit lost about the intent, does this PR only allow importing all dlists from a scene or all dlists from any .c file?

sauraen: What it actually does, assuming custom import, is all DLs from any C file. This is probably only useful for scenes (meaning scenes + rooms).

fast64_internal/oot/oot_f3d_writer.py
sauraen: Sure.

fast64_internal/oot/oot_f3d_writer.py
sauraen: I mean sure, that's a better place to put them, but I think it's crazy to put one of the controls for an operator in one place and leave the rest of them in a different place. This isn't a new category of control, this is exactly like the other DL import / export controls. When the other ones get "migrated", whoever does that can migrate this one too.

krm01:
I have been testing out this branch (great work by the way), it is extremely helpful. Some scenes import beautifully, with everything split up nicely and perfect textures - `SCENE_YDAN (Inside the Deku Tree)` is a great example. Some others have issues, but unrelated to this work, just problems around importing DLs from scenes in general (because some use "sceneTex" textures I think)

sauraen:
Closing in favor of upcoming scene importer by @kurethedead .

PR #79, opened 2022-02-20
[OoT] Fixed Scene Export and removed unwanted changes in ``spec``

Yanis42:
### Remove unwanted changes:
With the old code it was changing stuff that triggered a change in git, and I find it really annoying when it's only spaces replaced by tabs

Things that changed:
- Avoid using tabs when exporting ``spec`` data (indent defined by the ``indent`` variable in ``oot_utility.py``, currently it is 4 spaces)
- Keep ``compress`` in spec only if there's at least an other one, otherwise it won't add this flag

### Fix OoT Scene Export:
[Recently, OoT Decomp changed how scenes are handled in the project.](https://github.com/zeldaret/oot/pull/1128)

Things that changed:
- Removed ``oot_scene_id_c.py`` and ``oot_segment_symbols_c.py`` as it is no longer needed
- Complete rewrite of ``oot_scene_table_c.py`` as it was better than spending hours to understand how it was working
- Fixed scene deletion option and added a safety check in ``modifySegmentDefinition()`` for the ``levelC`` parameter
- Keep the same formatting when exporting the scene table data in ``scene_table.h``, note that every parameters in from the ``DEFINE_SCENE`` macro can be handled whenever it is needed, currently there's only the draw config that may need changes
- Added a comment to tell the user when new scenes start (there's already a comment like this for debug-only scenes in the table)
- Removed unused code in [``OOT_SearchSceneEnumOperator``](https://github.com/Fast-64/fast64/blob/ef56b01d7d76312b8b0640d868836f1d1dd99737/fast64_internal/oot/oot_scene_room.py#L89) but this will need confirmation, for me it is unused but I want to be sure before this gets merged

The code I wrote for ``oot_scene_table_c.py`` can be enhanced with ``re`` but I really don't understand how to use it and I didn't want to spend a lot of time trying as this is one of the main features of Fast64 that broke (also it makes the code easier to understand since I used basic things like lists)

As always, if you have any suggestions I'll take them!


fast64_internal/oot/oot_scene_room.py
```diff
@@ -93,33 +93,9 @@ class OOT_SearchSceneEnumOperator(bpy.types.Operator):
...
-			sceneHeader = obj.ootAlternateSceneHeaders.adultDayHeader
-		elif self.headerIndex == 3:
-			sceneHeader = obj.ootAlternateSceneHeaders.adultNightHeader
-		else:
-			sceneHeader = obj.ootAlternateSceneHeaders.cutsceneHeaders[self.headerIndex - 4]
-
-		if self.headerIndex == -1:
-			context.scene.ootSceneOption = self.ootSceneID
-		else:
-			sceneHeader.exitList[self.index].sceneID = self.sceneID
```
Dragorn421, line 122: I agree with removing this, it doesn't make sense 👍

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,123 @@
...
+	try:
+		with open(os.path.join(exportPath, 'include/tables/scene_table.h')) as fileData:
+			# keep the relevant data and do some formatting
+			for i, line in enumerate(fileData):
+				if not line.startswith("// "):
+					if not (line.startswith("/**") or line.startswith(" *")):
+						dataList.append(line[(line.find("(") + 1):].rstrip(")\n").replace(" ", "").split(','))
+					else: fileHeader += line
+				elif not line.startswith("// Added scenes"): debugLine = line
+	except: raise PluginError("ERROR: Can't find scene_table.h!")
```
Dragorn421, line 21: monkaSameLine (and everywhere else in the file) (this hurts readability really bad imo)
```suggestion
					else:
						fileHeader += line
				elif not line.startswith("// Added scenes"):
					debugLine = line
	except:
		raise PluginError("ERROR: Can't find scene_table.h!")
```

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,123 @@
...
+
+	# return the parsed data, the header comment and the comment mentionning debug scenes
+	return dataList, fileHeader, debugLine
+
+def getSceneIndex():
+	'''Returns the index (int) of the chosen scene, returns None if ``Custom`` is chose'''
```
Dragorn421, line 27:
```suggestion
	'''Returns the index (int) of the chosen scene, returns None if ``Custom`` is chosen'''
```

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,123 @@
...
+
+	raise PluginError("ERROR: Scene Index not found!")
+
+def getSceneParams(scene, exportInfo):
+	'''Returns the parameters that needs to be set in ``DEFINE_SCENE()``'''
+	sceneIndex = getSceneIndex()
+	sceneName = sceneTitle = sceneID = sceneUnk10 = sceneUnk12 = None
+
+	# if the index is None then this is a custom scene
+	if sceneIndex == None and scene != None:
```
Dragorn421, line 44: `is None`, `is not None` and wherever else

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,123 @@
...
+		if isEndOfNonDebugScenes:
+			fileData += debugLine
+			isEndOfNonDebugScenes = False
+
+		fileData += f"/* 0x{i:02X} */ DEFINE_SCENE("
+		for j in range(len(data[i])):
+			fileData += f"{data[i][j]}"
+			if j < 5: fileData += ", "
+			if data[i][0] == "testroom_scene": isEndOfExistingScenes = True
+			if data[i][0] == "ganon_tou_scene": isEndOfNonDebugScenes = True
```
Dragorn421, line 78: I think this can be outside the `for j` loop
```suggestion
		if data[i][0] == "testroom_scene": isEndOfExistingScenes = True
		if data[i][0] == "ganon_tou_scene": isEndOfNonDebugScenes = True
```

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,123 @@
...
+		
+		# adds the "// Debug-only scenes" comment after SCENE_GANON_TOU
+		if isEndOfNonDebugScenes:
+			fileData += debugLine
+			isEndOfNonDebugScenes = False
+
+		fileData += f"/* 0x{i:02X} */ DEFINE_SCENE("
+		for j in range(len(data[i])):
+			fileData += f"{data[i][j]}"
+			if j < 5: fileData += ", "
```
Dragorn421, line 76:
```suggestion
		fileData += ", ".join(str(d) for d in data[i])
```

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,123 @@
...
+	if scene == None: sceneDrawConfig = None
+	else: sceneDrawConfig = scene.sceneTableEntry.drawConfig
+	sceneParams = [sceneName, sceneTitle, sceneID, sceneDrawConfig, sceneUnk10, sceneUnk12]
+
+	# check if it's a custom scene name
+	if sceneIndex == None: isCustom = True
+	else: isCustom = False
+
+	# if so, check if the custom scene already exists in the data
+	# if it already exist set isCustom to false to consider it like a normal scene
```
Dragorn421, line 99:
```suggestion
	# if it already exists set isCustom to False to consider it like a normal scene
```

fast64_internal/oot/c_writer/oot_spec.py
```diff
@@ -12,23 +12,31 @@ def getSegmentDefinitionEntryBySceneName(segmentDefinition, sceneName):
...
 	table = []
+	compressFlag = ""
 	for match in re.finditer('beginseg(((?!endseg).)*)endseg', data, re.DOTALL):
 		segData = match.group(1)
 		table.append(segData)
 
-	return table
+		# avoid deleting compress flag if the user is using it
+		# (defined by whether it is present at least once in spec or not)
+		if segData.find("compress") != -1:
```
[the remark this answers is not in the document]
Yanis42: I didn't know you could do that, but now that I think about it it makes sense

fast64_internal/oot/c_writer/oot_spec.py
```diff
@@ -57,34 +65,38 @@ def modifySegmentDefinition(scene, exportInfo, levelC):
...
+				indent + 'include "' + includeDir + '_scene_main.o"\n' +\
+				indent + 'include "' + includeDir + '_scene_col.o"\n'
 			
-			if (levelC.sceneTexturesIsUsed()):
-				sceneSegInclude += '\tinclude "' + includeDir + '_scene_tex.o"\n'
-			
-			if (levelC.sceneCutscenesIsUsed()):
-				for i in range(len(levelC.sceneCutscenesC)):
-					sceneSegInclude += includeDir + '_cs_' + str(i) + '.o'
+			if levelC != None:
```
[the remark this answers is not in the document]
Yanis42: Once I got issues with ``is/is not`` so out of habit I'm using ``!=``, I'll change that

fast64_internal/oot/c_writer/oot_spec.py
```diff
@@ -57,34 +65,38 @@ def modifySegmentDefinition(scene, exportInfo, levelC):
...
+				if (levelC.sceneTexturesIsUsed()):
+					sceneSegInclude += indent + 'include "' + includeDir + '_scene_tex.o"\n'
+				
+				if (levelC.sceneCutscenesIsUsed()):
+					for i in range(len(levelC.sceneCutscenesC)):
+						sceneSegInclude += includeDir + '_cs_' + str(i) + '.o'
```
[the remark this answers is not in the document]
Yanis42: Good question, I don't even know why cutscenes are written in spec (unless it's the non-single file option, which makes sense in this context)

I never really used that so idk if it's working or not

fast64_internal/oot/oot_scene_room.py
Yanis42: yeah ok that's what I thought lol

I wonder how it ended up there

fast64_internal/oot/c_writer/oot_scene_table_c.py
Yanis42: I was a bit confused because of the indent but I understand, when I only have one instruction I like writing it on the same line but I get why it makes it harder to read 👍

fast64_internal/oot/c_writer/oot_scene_table_c.py
Yanis42: I really don't like how I named ``isEndOfExistingScenes`` and ``isEndOfNonDebugScenes`` there btw, I just couldn't find a better descriptive name

fast64_internal/oot/c_writer/oot_spec.py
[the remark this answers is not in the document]
Yanis42: Turns out this was broken, it is now fixed in this PR

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,144 @@
...
+							line[(line.find("(") + 1) :]
+							.rstrip(")\n")
+							.replace(" ", "")
+							.split(",")
+						)
+					else:
+						fileHeader += line
+				elif not line.startswith("// Added scenes"):
+					debugLine = line
+	except:
```
Dragorn421, line 28:
```suggestion
	except FileNotFoundError:
```
to avoid catching all kinds of exceptions

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,144 @@
...
+			fileData.append(sceneParams)
+			sceneIndex = len(fileData) - 1
+			break
+
+	# remove the scene data if scene is None (`Remove Scene` button)
+	try:
+		if scene is None:
+			fileData.remove(fileData[sceneIndex])
+	except:
+		raise PluginError("ERROR: Scene not found in ``scene_table.h``!")
```
Dragorn421, line 141:
```suggestion
	if scene is None:
		if sceneIndex < len(fileData):
			fileData.pop(sceneIndex)
		else:
			raise PluginError("ERROR: Scene not found in ``scene_table.h``!")
```

Yanis42: I had to add a check to prevent the removing bug you mentioned earlier and I'm not convinced by the way I added that suggestion

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,145 @@
...
+		if isCustom == False and sceneParams[i] is not None and fileData[sceneIndex][i] != sceneParams[i]:
+			fileData[sceneIndex][i] = sceneParams[i]
+		elif isCustom:
+			fileData.append(sceneParams)
+			sceneIndex = len(fileData) - 1
+			break
+
+	# remove the scene data if scene is None (`Remove Scene` button)
+	if scene is None:
+		if sceneIndex < len(fileData):
```
Dragorn421, line 138: mb on the suggested change here also need
```suggestion
		if sceneIndex is not None and sceneIndex < len(fileData):
```

fast64_internal/oot/oot_constants.py
```diff
@@ -1366,21 +1372,15 @@
...
+	"SCENE_GANON_TOU" : "ganon_tou",
+	"SCENE_TEST01" : "test01",
 	"SCENE_BESITU" : "besitu",
 	"SCENE_DEPTH_TEST" : "depth_test",
-	"SCENE_SASATEST" : "sasatest",
-	"SCENE_SUTARU" : "sutaru",
 	"SCENE_SYOTES" : "syotes",
-	"SCENE_SYOTES2" : "syotes2",
-	"SCENE_TEST01" : "test01",
+	"SCENE_SYOTES2" : "syotes2_scen",
```
Dragorn421, line 1380:
```suggestion
	"SCENE_SYOTES2" : "syotes2",
```
maybe?

Dragorn421:
Bug reports

-------

1) Choose scene -> deku tree
2) "remove scene"
works fine but then
3) Choose scene -> temple of time
4) "remove scene"
only removes the spec entries for ToT, and nothing from scene_table.h

----------

1) Choose scene -> temple of time
2) "remove scene"
works fine but then
3) "export scene"
only adds entries in the spec, nothing in scene_table.h
this may be "normal" behavior but looks weird idk

Yanis42:
> Bug reports
> 
>     1. Choose scene -> deku tree
> 
>     2. "remove scene"
>        works fine but then
> 
>     3. Choose scene -> temple of time
> 
>     4. "remove scene"
>        only removes the spec entries for ToT, and nothing from scene_table.h
> 
> 
>     1. Choose scene -> temple of time
> 
>     2. "remove scene"
>        works fine but then
> 
>     3. "export scene"
>        only adds entries in the spec, nothing in scene_table.h
>        this may be "normal" behavior but looks weird idk

Both of these bugs should be fixed now

Dragorn421:
Slight bug report:

1) Choose scene -> "Custom" -> "overworld", "mysc" (or whatever)
2) Export Scene
everything's fine for now
3) Choose scene -> "Treasure Chest Room (Testroom)"
4) Remove Scene
everything's fine except the `// Added scenes` comment line is gone

Yanis42:
> Slight bug report:
> 
>     1. Choose scene -> "Custom" -> "overworld", "mysc" (or whatever)
> 
>     2. Export Scene
>        everything's fine for now
> 
>     3. Choose scene -> "Treasure Chest Room (Testroom)"
> 
>     4. Remove Scene
>        everything's fine except the `// Added scenes` comment line is gone

Sorry for the delay, this should be fixed now (the debug comment had this bug too btw)

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,211 @@
...
-		return "\t" + self.sceneType + "(" + self.name + ", " + \
-			((self.title + ", ") if self.title is not None else "")  +\
-			self.unk10 + ", " + self.config + ", " + self.unk12 + "),\n"
-
-def getSceneTableEntryBySceneName(sceneTable, sceneName):
-	for entry in sceneTable:
-		if entry.name[:-6] == sceneName:
-			return entry
+import os
+from ...utility import bpy, PluginError, writeFile
```
Dragorn421, line 2:
```suggestion
import bpy
from ...utility import PluginError, writeFile
```

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,211 @@
...
+			i += 1
+
+	raise PluginError("ERROR: Scene Index not found!")
+
+def getInsertionIndex(sceneNames, sceneName, index, mode):
+	'''Returns the index to know where to insert data'''
+	# special case where the scene is "Inside the Great Deku Tree"
+	# since it's the first scene simply return 0
+	if sceneName == "SCENE_YDAN":
+		return 0
```
Dragorn421, line 69: Why is this special case needed? I feel like `getOriginalIndex` would work just fine with deku tree

fast64_internal/oot/c_writer/oot_scene_table_c.py
```diff
@@ -1,75 +1,211 @@
...
+			fileData.insert(index, sceneParams)
+
+	# remove the scene data if scene is None (`Remove Scene` button)
+	if scene is None:
+		if sceneIndex is not None and sceneIndex < len(fileData):
+			if (exportName + "_scene") == fileData[sceneIndex][0]:
+				sceneNames.pop(sceneIndex)
+				fileData.pop(sceneIndex)
+			else:
+				raise PluginError("ERROR: Scene not found in ``scene_table.h``!")
```
Dragorn421, line 208: so initially I suggested the `< len()` check because of the except IndexError but actually it was only meant to check for None? should probably get rid of the len check as it's misleading

Also is the `else: raise` on the right indentation level? shouldn't it also be raised if `sceneIndex is None`?

Dragorn421 (commented):
Just a few code things
also you can fix the existing "Ganandorf" typo while you're at it

fast64_internal/oot/c_writer/oot_scene_table_c.py
Yanis42: When you export after removing the deku tree scene it will act like it's a custom scene, so I figured returning ``0`` if the scene's ``SCENE_YDAN`` was an easy solution

fast64_internal/oot/c_writer/oot_scene_table_c.py
Yanis42: I'm not sure to understand, the len check makes sense since we're dealing with a list, it's just an extra safety check even though if should never be OoB

For the ``else: raise`` it makes more sense this way to me since it's with ``if (exportName + "_scene") == fileData[sceneIndex][0]:`` that we're actually looking for the scene, otherwise I'd prefer another error message as this one won't be really accurate since we never started checking the scene's existence

Dragorn421 (approved):

PR #81, opened 2022-03-10
fixed export bug by adding missing params to function call

thecozies:
more `lastMaterialName` misses 😔 i'm surprised this hasnt been encountered before yesterday


Dragorn421 (approved):
now this is for real the last time it happens 😄

PR #86, opened 2022-03-20
Applied black formatting to all files

thecozies:
important thing to note: i set black to be formatted for python 3.10 because its Blender's current version of python, and it also allows us to have 3.10 specific features formatted.

I believe its time for us to deprecate < 3.1 support. This is a very small group of people and having to maintain compatibility for older releases slows us down too much. Plus, 3.1 is going to open up a lot of possibilities for us. One big thing is UI enhancements. 2.8X has a ton of missing UI features. Also geometry nodes, we'll be attempting to simulate vertex normal based lighting and the differences between nodes prior to 3.1 aren't worth resolving. Its a good time to move forward, and make this plugin easier to develop!


sauraen:
Regarding the Blender version: My understanding is that fast64 is a Blender 2.8x plugin, which *happens* to work on later versions. I'd be okay with checking in with all the fast64 contributors if they are okay with moving up to 3.0. 3.1 has just released so I think it's far too early to mandate that version.

sauraen:
Also--if we are using an auto format tool, can we set up either a git hook to apply the formatting, or some GitHub CI thing which will auto-reject PRs which don't meet the formatting?

PR #87, opened 2022-04-02
[OoT] Fix vertex order in collision polygons

Dragorn421:
See added comments in the code for what this fixes.

"1) The vertex with the minimum y coordinate should be first." is a previous fix that I moved

I also had to change `convertNormalizedFloatToShort` to return a signed value instead of unsigned to get a useful normal value to compare to. This "utility" function is only used by the modified collision export function and it seems like a good change in general anyway.


thecozies (approved):
lgtm

PR #91, opened 2022-04-04
Updater

Dragorn421:
Uses https://github.com/CGCookie/blender-addon-updater

**Note:** this doesn't play well with git in certain cases, as it can remove part or all of the .git folder

This seems to work at least on my machine with Blender 2.93 and 3.1.2. I'd like more testing, I had both good and bad experiences with this in the past.

--------------

The UI is located in the addon preferences:

<details>

image

After clicking on "Check now for fast64 update":

image

</details>

This should allow to update to latest main whenever without using tags/releases.
It still does need a version for the code to work so fast64 is now fast64 1.0.0 🍾


__init__.py
```diff
@@ -323,6 +334,12 @@ def after_load(_a, _b):
 # register operators and panels here
 # append menu layout drawing function to an existing window
 def register():
+
```
thecozies, line 337:
```suggestion
```
smh extra whitespace

__init__.py
```diff
@@ -13,9 +13,12 @@
 import cProfile
 import pstats
 
+from . import addon_updater_ops
+
 # info about add on
 bl_info = {
 	"name": "Fast64",
+	"version": (1, 0, 0),
```
thecozies, line 21: so we'll need to update this on each merge?

Dragorn421: No, if I remember right the updater only uses this as the current version when comparing to tags (like v1.0.1 maybe idk), to determine if there is any new version. But the updater here is configured to not care about versions and only update to the latest main

thecozies (commented):
Can you add an explanation to the top of the readme (or anywhere) how to update with this new method? (you might be able to sneak in some info about 3.1 if you wanted to)

Dragorn421:
Good idea
https://github.com/Dragorn421/fast64/blob/updater/README.md#updater
also some notes on how versions work

__init__.py
Dragorn421: I was somewhat mistaken on this, the answer is still "no" but the

> the updater here is configured to not care about versions and only update to the latest main

part is wrong

Reonu:
merge this or I will literally be sad

Dragorn421:
Planning to merge this when we're sure we have time to fix issues that may arise

PR #93, opened 2022-05-16
Add `anim_range_choice` property

Dragorn421:
Add a drop-down list option in the anim export panel, to pick what range of the exported animation to use.

I exported a basic animation with all three options for both OoT and SM64, seems to work on my end.

# Screenshot

image

### DIFF
--- a/fast64_internal/oot/oot_anim.py
+++ b/fast64_internal/oot/oot_anim.py
@@ -114,7 +114,7 @@ def ootGetAnimBoneRot(bone, poseBone, convertTransformMatrix, isRoot):
 		raise RuntimeError('Animation contains non-root bones with animated translation. OoT SkelAnime only supports animated translation on the root bone.')
 	return finalRotation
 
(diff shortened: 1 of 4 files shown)


fast64_internal/oot/oot_anim.py
```diff
@@ -114,7 +114,7 @@ def ootGetAnimBoneRot(bone, poseBone, convertTransformMatrix, isRoot):
 		raise RuntimeError('Animation contains non-root bones with animated translation. OoT SkelAnime only supports animated translation on the root bone.')
 	return finalRotation
 
-def ootConvertAnimationData(anim, armatureObj, frameInterval, convertTransformMatrix):
+def ootConvertAnimationData(anim, armatureObj, convertTransformMatrix, *, frame_start, frame_count):
```
thecozies, line 117: what kind of python sorcery are you doin

Zelllll (approved):
Great job, looks good!

Dragorn421:
Default fast64 behavior unchanged with this.

PR #94, opened 2022-05-19
Add object name to "object has no materials" error

Dragorn421:
new error with object name

### DIFF
--- a/fast64_internal/f3d/f3d_writer.py
+++ b/fast64_internal/f3d/f3d_writer.py
@@ -129,7 +129,7 @@ def fixLargeUVs(obj):
 
 	texSizeDict = {}
 	if len(obj.data.materials) == 0:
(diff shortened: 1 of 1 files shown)


fast64_internal/f3d/f3d_writer.py
```diff
@@ -129,7 +129,7 @@ def fixLargeUVs(obj):
 
 	texSizeDict = {}
 	if len(obj.data.materials) == 0:
```
thecozies, line 131: should this also be checking if all applied materials are F3D? cus maybe the error message should instead say "object has zero materials" to be more accurate

thecozies (approved):
i understand!

Dragorn421:
Merging as a bandaid fix of the error not being very helpful right now.
It's unclear if this error should be moved somewhere else, reworded, or if it is strictly needed at all.

PR #95, opened 2022-05-19
[OoT] OoT Object changes & small changes/fixes/removals

Yanis42:
### List of changes:
- Added ``ObjectList.xml`` which contains every objects from OoT, you have the name, the decomp ID and something called ``ObjectKey``, it's used to identify an object, this should never change (the format is ``obj_`` + the debug name of the object)
- Added ``ActorList.xml``, currently unused outside identifying an actor, it will be used whenever PR 56 is merged though, the format of this file is more complex because there's every parameters etc, but it's a bit like the object list file
- Export defines in scenes for the lengths of actor and object lists, makes manual editing easier to do
- Fixed the scene being automatically selected in blender when you export (at least I think it's fixed)
- Replaced remaining ``\t`` in OoT files with the ``indent`` variable introduced with PR 79 (decomp uses 4 spaces)
- Deleted ``oot_parse.py`` and ``sceneDirectoryParser.py`` as it's unused
- Automatically add missing objects to the Room when you export the scene, (hence the ``ActorList.xml`` file being here), it's also adding it in the UI
- Finally, I got kinda got rid of debug names for the object list, now we're generating the list from the XML data, the key is the data saved in the blend (hence why you can't change that, it's there to be future proof for decomp changed) and the names are displayed in the list, makes choosing an object a lot better (for instance, you can type "zelda" and have every objects related to Zelda, see the preview image to see this example in action)
- Alternate room headers should work with these new features (child/adult day/night + cutscene headers)

TL;DR: new features regarding OoT objects and some fixes/removed unused stuff/enhanced the C output a bit

<details closed>
<summary>Object list preview</summary>

</details>

<details closed>
<summary>Actor/Object Lists defines preview</summary>

</details>


fast64_internal/oot/oot_utility.py
```diff
@@ -586,3 +586,57 @@ def oot_utility_register():
...
+		elif headerIndex == 1:
+			roomProp = roomObj.ootAlternateRoomHeaders.childNightHeader
+		elif headerIndex == 2:
+			roomProp = roomObj.ootAlternateRoomHeaders.adultDayHeader
+		elif headerIndex == 3:
+			roomProp = roomObj.ootAlternateRoomHeaders.adultNightHeader
+		elif csHeaderIndex is not None:
+			roomProp = roomObj.ootAlternateRoomHeaders.cutsceneHeaders[csHeaderIndex]
+		if roomProp is not None:
+			collection = getCollection(roomObj.name, "Object", headerIndex)
```
Dragorn421, line 628:
```suggestion
			collection = roomProp.objectList
```

fast64_internal/oot/oot_utility.py
```diff
@@ -586,3 +586,57 @@ def oot_utility_register():
...
+			collection.add()
+			collection.move(len(collection)-1, (headerIndex + 1))
+			roomProp.objectList[len(roomProp.objectList) - 1].objectID = objectID
```
Dragorn421, line 631: `collection` / `roomProp.objectList` ?

what does move() do here? (or in general)

```suggestion
			collection.move(len(collection)-1, (headerIndex + 1))
			roomProp.objectList[-1].objectID = objectID
```

fast64_internal/oot/oot_constants.py
```diff
@@ -488,6 +505,11 @@
 	("ACTOR_OBJ_WARP2BLOCK", "OBJ_WARP2BLOCK", "OBJ_WARP2BLOCK"),
 ]
 
+# Add every objects
+ootEnumObjectID = [(objectNode.get('Key'), objectNode.get('Name'),
```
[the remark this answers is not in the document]
Yanis42: I forgot about older blends

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -256,25 +263,25 @@ def ootAlternateRoomMainToC(scene, room):
 	altHeader.source = "SCmdBase* " + room.alternateHeadersName() + "[] = {\n"
 	
 	if room.childNightHeader is not None:
-		altHeader.source += "\t" + room.roomName() + "_header" + format(1, '02') + ",\n"
+		altHeader.source += indent + "" + room.roomName() + "_header" + format(1, '02') + ",\n"
```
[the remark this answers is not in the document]
Yanis42: I missed these since I used the multi-selection thing from vs code lol

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -519,6 +526,12 @@ def ootAlternateSceneMainToC(scene):
 def ootSceneMainToC(scene, headerIndex):
 	sceneMainC = CData()
 
+	for i in range(len(scene.rooms)):
+		room = scene.rooms[i]
+		sceneMainC.append(ootGetHeaderDefines(room, headerIndex))
+		if room.hasAlternateHeaders():
+			sceneMainC.append(ootGetAltHeaderDefines(room))
```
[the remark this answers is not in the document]
Yanis42: I guess I just got confused between room and scene, will fix it

fast64_internal/oot/ObjectList.xml
```diff
@@ -0,0 +1,382 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Table KeyType="System.UInt16" ValueType="System.String">
+  <Object ID="OBJECT_HUMAN" Key="obj_human" Name="Aria - HUMAN"/>
```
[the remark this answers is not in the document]
Yanis42: I thought you wanted debug names in the actual name 🤔

> the blender name should be constructed from Name and ID without the OBJECT_ prefix

isn't this what I did? I don't understand lol

fast64_internal/oot/oot_utility.py
Yanis42: I never worked with "collections" before so I just did exactly what the "add" button does when you're adding objects

So if I change this I'll change the other code too to be consistent

fast64_internal/oot/oot_constants.py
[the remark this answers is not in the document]
Dragorn421: madge

fast64_internal/oot/ObjectList.xml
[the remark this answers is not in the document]
Dragorn421: I agree with the result, it's the journey that I think could be better

I mean in blender instead of `get("Name")` = `"Aria - HUMAN"` it could be `get("Name") + " - " + get("ID").replace("OBJECT_", "")` = `"Aria" + " - " + "HUMAN"` = `"Aria - HUMAN"`

fast64_internal/oot/oot_utility.py
Dragorn421: collection methods have no api page so it's very cool to use YEP

fast64_internal/oot/ObjectList.xml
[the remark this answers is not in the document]
Yanis42: ahh ok I see, I'll change that then (and I'll do it in the PR too I guess, for consistency)

fast64_internal/oot/oot_utility.py
Yanis42: no documentation, love it YEP

fast64_internal/oot/data/oot_getters.py
```diff
@@ -0,0 +1,38 @@
+from xml.etree.ElementTree import parse as XMLParse, Element
```
Dragorn421, line 1:
```suggestion
from xml.etree.ElementTree import parse as parseXML, Element
```
I thought XMLParse was a fancy class :sweat_smile:

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,43 @@
...
+
+    from dataclasses import dataclass
+
+    @dataclass
+    class OoT_Actor:
+        id: str
+        key: str
+        name: str
+        category: str
+        tiedObjects: str
```
Dragorn421, line 12: this should be moved out of the class imo, it's unusual to define stuff inside classes, and this fits the module (oot_actor_data) anyway so okay to have at the global level

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,43 @@
...
+        self.actorXML: str
+
+        # general actor list
+        self.actorList: list[self.OoT_Actor] = []
+
+        # list of tuples used by Blender's enum properties
+        self.ootEnumActorID: list[tuple] = []
+        self.ootEnumActorIDLegacy: list[tuple] = []  # for old blends
+
+        self.initActorLists()
```
Dragorn421, line 25: `initActorLists` is unconditionally called on `__init__` and unused elsewhere, so its code should just be inside `__init__` directly

fast64_internal/oot/data/oot_data.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+from .oot_object_data import OoT_ObjectData
+from .oot_actor_data import OoT_ActorData
+from .oot_getters import OoT_Getters
+
+
+class OoT_Common:
+    """Common Data used by other functions or classes"""
+
+    getters = OoT_Getters()
```
Dragorn421, line 10: I feel like you're reinventing the import system here,
currently you're doing
```py
# oot_getters.py
class OoT_Getters:
    """Common 'get' functions"""

    def getRoot(self, xmlPath: str) -> Element:
        ...

# somewhere else
from .oot_data import OoT_Common
OoT_Common.getters.getRoot(...)
```
which is afaict the same as the more pythonic/common:
```py
# oot_getters.py
def getRoot(self, xmlPath: str) -> Element:
    ...

# somewhere else
from . import oot_getters
oot_getters.getRoot(...)
```

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -839,3 +872,21 @@ def __init__(self):
...
         self.roomMeshC = {}
+
+
+def ootGetHeaderDefines(room, headerIndex):
+    """Returns CData containing defines for actor and object lists lengths"""
+    data = CData()
+
+    if len(room.objectList) > 0:
+        data.header += (
+            "#define LENGTH_" + str(room.objectListName(headerIndex)).upper() + " " + str(len(room.objectList))
```
Dragorn421, line 885: I think instead of having the `LENGTH_` name generated in two places (and risking them to go out of sync) you could add a function to `OOTRoom` similar to `objectListName` and do like
```suggestion
            "#define " + str(room.objectListLengthDefineName(headerIndex)) + " " + str(len(room.objectList))
```
(`objectListLengthDefineName(` is only a suggestion :monkasteer:)

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -839,3 +872,21 @@ def __init__(self):
...
+        ) + "\n"
+    if len(room.actorList) > 0:
+        data.header += (
+            "#define LENGTH_" + str(room.actorListName(headerIndex)).upper() + " " + str(len(room.actorList))
+        ) + "\n"
+    if not (data.header == ""):
```
Dragorn421, line 891:
```suggestion
    if data.header != "":
```
kek

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self):
+        # Path to the ``ActorList.xml`` file
+        self.actorXML: str
+
+        # general actor list
+        self.actorList: list[OoT_Actor] = []
+
+        # list of tuples used by Blender's enum properties
+        self.ootEnumActorID: list[tuple] = []
+        self.ootEnumActorIDLegacy: list[tuple] = []  # for old blends
```
Dragorn421, line 27: this isn't needed, the members are set with `getEnumList` below

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,40 @@
...
+    category: str
+    tiedObjects: str
+
+
+class OoT_ActorData:
+    """Everything related to OoT Actors"""
+
+    def __init__(self):
+        # Path to the ``ActorList.xml`` file
+        self.actorXML: str
```
Dragorn421, line 20: I think this can be a normal variable, doesn't have to be a class member?
```suggestion
        actorXML: str
```
also this line is not really needed anyway with `actorXML` being set below

fast64_internal/oot/data/oot_getters.py
```diff
@@ -0,0 +1,37 @@
+from xml.etree.ElementTree import parse as parseXML, Element
+
+
+def getRoot(xmlPath: str) -> Element:
```
Dragorn421, line 4: I think this function should be named after "xml" or "parsing" or both or something

`parseXmlGetRoot` 4head

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,91 @@
...
+class OoT_ObjectData:
+    """Everything related to OoT objects"""
+
+    from dataclasses import dataclass
+
+    @dataclass
+    class OoT_Object:
+        id: str
+        key: str
+        name: str
```
Dragorn421, line 16: madge

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,91 @@
...
+        # general object list
+        self.objectList: list[self.OoT_Object] = []
+
+        # list of tuples used by Blender's enum properties
+        self.ootEnumObjectID: list[tuple] = []
+        self.ootEnumObjectIDLegacy: list[tuple] = []  # for old blends
+
+        self.initObjectLists()
+
+    def initObjectLists(self):
```
Dragorn421, line 31: madge x2

fast64_internal/oot/data/oot_getters.py
```diff
@@ -0,0 +1,37 @@
...
+    """Parse an XML file and return its root element"""
+    try:
+        return parseXML(xmlPath).getroot()
+    except:
+        from ...utility import PluginError
+
+        raise PluginError(f"ERROR: File '{xmlPath}' is missing or malformed.")
+
+
+def getEnumList(dataList: list, customName: str):
```
Dragorn421, line 14: it's weird that this expects dataList elements to have id, key, name

I'd either embrace supporting two different classes that happen to have id, key, name defined
```py
from .oot_actor_data import OoT_Actor
from .oot_object_data import OoT_ObjectData
def getEnumList(dataList: list[OoT_Actor | OoT_ObjectData.OoT_Object], customName: str):
```
this may have circular import issues, that could be fixable with e.g. `from . import oot_actor_data` and annotate with `"oot_actor_data.OoT_Actor"` (notice it's a string)

or create a base dataclass which OoT_Actor and OoT_Object derive from (a better idea imo)
```py
@dataclass
class OoT_DataElement: # name???
    id: str
    key: str
    name: str
def getEnumList(dataList: list[OoT_DataElement], customName: str):
```
```py
@dataclass
class OoT_Object(OoT_DataElement):
    pass # nothing else to define :shrug: 
```

fast64_internal/oot/data/oot_getters.py
```diff
@@ -0,0 +1,37 @@
...
+
+def getIDFromKey(key: str, dataList: list) -> (str | None):
+    """Returns the actor/object ID using the key"""
+    if not (key == "Custom"):
+        for elem in dataList:
+            if elem.key == key:
+                return elem.id
+    else:
+        return key
+    return None
```
Dragorn421, line 37: I would create a dictionary as well as a list instead of looking for elements like this, so the lookup would be like
```py
dataDictByKey.get(key)
```

fast64_internal/oot/oot_level_writer.py
```diff
@@ -84,6 +84,13 @@ def ootExportSceneToC(originalSceneObj, transformMatrix, f3dType, isHWv1, sceneN
         if exportSubdir == "":
             raise PluginError("Scene folder " + sceneName + " cannot be found in the ootSceneDirs list.")
 
+    roomObjects = [
+        obj for obj in bpy.data.objects if ((obj.parent == originalSceneObj) and (obj.ootEmptyType == "Room"))
+    ]
+    actorList = ootData.actorData.actorList
+    for i in range(len(scene.rooms)):
+        ootData.objectData.addMissingObjectsToList(roomObjects[i], scene.rooms[i], actorList, 0, None)
```
Dragorn421, line 92:
```suggestion
    for roomObject, room in zip(roomObjects, scene.rooms):
        ootData.objectData.addMissingObjectsToList(roomObject, room, actorList, 0, None)
```

fast64_internal/oot/oot_scene_room.py
```diff
@@ -172,13 +172,24 @@ def drawExitProperty(layout, exitProp, index, headerIndex, objName):
...
+    objectIDCustom: bpy.props.StringProperty(default="OBJECT_CUSTOM")
+
+    @staticmethod
+    def upgrade_object(obj):
+        if obj.data is None:
+            if (obj.ootEmptyType == "Scene") or (obj.ootEmptyType == "Room"):
+                print(f"Processing '{obj.name}'...")
+                for child in obj.children:
+                    OOTObjectProperty.upgrade_object(child)
+            elif obj.fast64.oot.version < obj.fast64.oot.cur_version:
```
Dragorn421, line 186: so, `if (obj.ootEmptyType == "Scene") or (obj.ootEmptyType == "Room")` then `upgradeObjectInit` doesn't run? what (blender) objects does it actually run on?

also
```suggestion
            elif obj.fast64.oot.version < 1:
```
(for example changing from 1 to 2 wouldn't need upgradeObjectInit)

fast64_internal/oot/oot_scene_room.py
```diff
@@ -172,13 +172,24 @@ def drawExitProperty(layout, exitProp, index, headerIndex, objName):
 
 class OOTObjectProperty(bpy.types.PropertyGroup):
     expandTab: bpy.props.BoolProperty(name="Expand Tab")
-    objectID: bpy.props.EnumProperty(items=ootEnumObjectID, default="OBJECT_HUMAN")
-    objectIDCustom: bpy.props.StringProperty(default="OBJECT_HUMAN")
+    objectID: bpy.props.EnumProperty(items=ootData.objectData.ootEnumObjectID, default="obj_human")
+    objectIDLegacy: bpy.props.EnumProperty(items=ootData.objectData.ootEnumObjectIDLegacy, default="OBJECT_HUMAN")
```
Dragorn421, line 176: this breaks old blends, doesn't it? `objectID` with the new enum won't retain old values with the old enum, they'll be invalid and discarded afaik (but I may be wrong?)

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,91 @@
...
+    def upgradeObjectInit(self, obj, objectList):
+        """Object upgrade logic"""
+        if obj.data is None and obj.ootEmptyType == "Room":
+            for i in range(len(obj.objectList)):
+                obj.objectList[i].objectIDLegacy = obj.objectList[i].objectID
+                for object in objectList:
+                    if obj.objectList[i].objectIDLegacy == object.id:
+                        obj.objectList[i].objectID = object.key
+
+            obj.fast64.oot.version = obj.fast64.oot.cur_version
```
Dragorn421, line 47: setting the version probably should be done somewhere else, as this function deals specifically with version 0 -> version 1 and for the object list

probably in `upgrade_changed_props` ?
(`SM64_ObjectProperties.upgrade_changed_props` and another one does so it looks like)

fast64_internal/oot/oot_level_classes.py
```diff
@@ -443,6 +444,23 @@ def hasAlternateHeaders(self):
...
         )
 
+    def getObjectLengthDefineName(self, headerIndex: int):
+        return f"LENGTH_{self.objectListName(headerIndex).upper()}"
+
+    def getActorLengthDefineName(self, headerIndex: int):
+        return f"LENGTH_{self.actorListName(headerIndex).upper()}"
+
+    def getObjectLengthDefine(self, headerIndex: int):
+        return f"#define {self.getObjectLengthDefineName(headerIndex)} {len(self.objectIDList)}\n"
```
Dragorn421, line 454: I think OOTRoom shouldn't be responsible for writing C, but also `getObjectLengthDefine` is only used once, so I'd put this code directly where it's used instead of `getObjectLengthDefine`

fast64_internal/oot/oot_level.py
```diff
@@ -149,10 +149,17 @@ def onUpdateOOTEmptyType(self, context):
...
+    cur_version = 1  # version after property migration
 
     scene: bpy.props.PointerProperty(type=OOTSceneProperties)
 
+    @staticmethod
+    def upgrade_changed_props():
+        for obj in bpy.data.objects:
+            if obj.data is None and (obj.ootEmptyType == "Room"):
+                OOTObjectProperty.upgrade_object(obj)
+                obj.fast64.oot.version = obj.fast64.oot.cur_version
```
Dragorn421, line 161:
```suggestion
            if obj.data is None and obj.ootEmptyType != "None":
                obj.fast64.oot.version = obj.fast64.oot.cur_version
```
meaning the version of all oot empties is upgraded here, otherwise non-room oot empties would still be "outdated" (their version would be 0 instead of cur_version==1)

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -382,33 +407,26 @@ def ootRoomCommandsToC(room, headerIndex):
...
-        altHeader.source += "\t0,\n"
-
-    if room.adultNightHeader is not None:
-        altHeader.source += "\t" + room.roomName() + "_header" + format(3, "02") + ",\n"
-        altData.append(ootRoomMainToC(scene, room.adultNightHeader, 3))
-    else:
-        altHeader.source += "\t0,\n"
+    for headerIndex, headerName in enumerate(["childNightHeader", "adultDayHeader", "adultNightHeader"], 1):
+        # equivalent to ``room.childNightHeader`` if ``headerName`` is "childNightHeader"
+        curHeader = getattr(room, headerName)
```
Dragorn421, line 419:
```suggestion
        curHeader = getattr(room, headerName)
```
readers should just read the doc if they don't know getattr :uhhuh_hackerman:

Dragorn421, line 419: also equivalent :uhhuh: :
```suggestion
    for headerIndex, curHeader in enumerate([room.childNightHeader, room.adultDayHeader, room.adultNightHeader], 1):
```

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -382,33 +407,26 @@ def ootRoomCommandsToC(room, headerIndex):
...
+        if curHeader is not None:
+            altHeader.source += indent + f"{room.roomName()}_header{headerIndex:02},\n"
+            altData.source += ootGetHeaderDefines(curHeader, headerIndex)
+            altData.append(ootRoomMainToC(scene, curHeader, headerIndex))
```
Dragorn421, line 423: why += `ootGetHeaderDefines()` here instead of in `ootRoomMainToC` ?

Couldn't `+= ootGetHeaderDefines(room, 0)` in `ootRoomMainToC` have `headerIndex` instead of 0 ?

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -839,3 +861,16 @@ def __init__(self):
         self.roomMainC = {}
         self.roomMeshInfoC = {}
         self.roomMeshC = {}
+
+
+def ootGetHeaderDefines(room: OOTRoom, headerIndex: int):
+    """Returns CData containing defines for actor and object lists lengths"""
```
Dragorn421, line 867:
```suggestion
    """Returns a string containing defines for actor and object lists lengths"""
```

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # list of tuples used by Blender's enum properties
+        self.ootEnumObjectKey: list[tuple] = []
+        self.ootEnumObjectIDLegacy: list[tuple] = []  # for old blends
+
+        # Path to the ``ObjectList.xml`` file
+        objectXML = path.dirname(path.abspath(__file__)) + "/xml/ObjectList.xml"
+
+        for obj in getXMLRoot(objectXML).iterfind("Object"):
+            objName = f"{obj.attrib['Name']} - {obj.attrib['ID'].replace('OBJECT_', '')}"
```
Dragorn421, line 29:
```suggestion
            objName = f"{obj.attrib['Name']} - {obj.attrib['ID'].removeprefix('OBJECT_')}"
```

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,89 @@
...
+            self.objectList.append(OoT_ObjectElement(obj.attrib["ID"], obj.attrib["Key"], objName))
+
+        self.objectsByID = {obj.id: obj for obj in self.objectList}
+        self.objectsByKey = {obj.key: obj for obj in self.objectList}
+        self.ootEnumObjectKey, self.ootEnumObjectIDLegacy = getEnumList(self.objectList, "Custom Object")
+
+    def upgradeObjectList(self, objList):
+        for obj in objList:
+            obj.objectKey = self.objectsByID.get(obj.objectID).key
```
Dragorn421, line 38:
```suggestion
            obj.objectKey = self.objectsByID[obj.objectID].key
```
it's better to use subscription (`[...]`) when you expect the key to be present, instead of get() which defaults to returning None

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def upgradeAltHeaders(self, roomObj):
+        altHeaders = roomObj.ootAlternateRoomHeaders
+        for header in ["childNightHeader", "adultDayHeader", "adultNightHeader"]:
+            curHeader = getattr(altHeaders, header)
```
Dragorn421, line 43: `for curHeader in [altHeaders.childNightHeader, ...`

fast64_internal/oot/oot_scene_room.py
```diff
@@ -172,13 +172,27 @@ def drawExitProperty(layout, exitProp, index, headerIndex, objName):
...
+    objectKey: bpy.props.EnumProperty(items=ootData.objectData.ootEnumObjectKey, default="obj_human")
+    objectID: bpy.props.EnumProperty(items=ootData.objectData.ootEnumObjectIDLegacy, default="OBJECT_HUMAN")
+    objectIDCustom: bpy.props.StringProperty(default="OBJECT_CUSTOM")
 
+    @staticmethod
+    def upgrade_object(obj):
+        if obj.fast64.oot.version < 1:
+            print(f"Processing '{obj.name}'...")
+            ootData.objectData.upgradeObjectList(obj.ootRoomHeader.objectList)
+            ootData.objectData.upgradeAltHeaders(obj)
```
Dragorn421, line 184: I feel like `upgradeObjectList(obj.ootRoomHeader.objectList)` could just go in `upgradeAltHeaders`, which could be renamed to `upgradeRoomHeaders`

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,89 @@
...
+            if curHeader is not None:
+                self.upgradeObjectList(curHeader.objectList)
+        for i in range(len(altHeaders.cutsceneHeaders)):
+            self.upgradeObjectList(altHeaders.cutsceneHeaders[i].objectList)
+
+    def addMissingObjectToUI(self, roomObj, headerIndex, objectKey, csHeaderIndex):
```
Dragorn421, line 49: I don't like that name, props are also stored in the blend to be saved, not just being displayed
no suggestion for now though

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,89 @@
...
+                collection[-1].objectKey = objectKey
+
+    def addMissingObjectsToList(self, roomObj, room, actorData, headerIndex, csHeaderIndex):
+        """Adds missing objects to the object list"""
+        if len(room.actorList) > 0:
+            for roomActor in room.actorList:
+                actor = actorData.actorsByID.get(roomActor.actorID)
+                if not (actor.key == "player") and len(actor.tiedObjects) > 0:
+                    for objKey in actor.tiedObjects:
+                        if not objKey in ["obj_gameplay", "obj_gameplay_dangeon_keep"]:
```
Dragorn421, line 75:
```suggestion
                        if objKey not in ["obj_gameplay", "obj_gameplay_dangeon_keep"]:
```
it's literally exactly the same thing

also what about field_keep ?

fast64_internal/oot/oot_level_writer.py
```diff
@@ -84,6 +84,12 @@ def ootExportSceneToC(originalSceneObj, transformMatrix, f3dType, isHWv1, sceneN
         if exportSubdir == "":
             raise PluginError("Scene folder " + sceneName + " cannot be found in the ootSceneDirs list.")
 
+    roomObjects = [
+        obj for obj in bpy.data.objects if ((obj.parent == originalSceneObj) and (obj.ootEmptyType == "Room"))
+    ]
+    for roomObj, room in zip(roomObjects, scene.rooms.values()):
+        ootData.objectData.addMissingObjectsToList(roomObj, room, ootData.actorData, 0, None)
+        ootData.objectData.addAltHeadersObjects(roomObj, room, ootData.actorData)
```
Dragorn421, line 92: I also feel like this `addMissingObjectsToList` could go into `addAltHeadersObjects`
and `addAltHeadersObjects` -> `addRoomHeadersObjects`

fast64_internal/oot/oot_level_writer.py
```diff
@@ -301,8 +307,13 @@ def readRoomData(room, roomHeader, alternateRoomHeaders):
     room.disableSkybox = roomHeader.disableSkybox
     room.disableSunMoon = roomHeader.disableSunMoon
     room.echo = roomHeader.echo
-    room.objectList.extend([getCustomProperty(item, "objectID") for item in roomHeader.objectList])
-    if len(room.objectList) > 15:
+    room.objectIDList.extend(
+        [
+            room.getCustomIDFromKey(ootData.objectData.objectsByKey, obj, "objectKey", "objectID")
```
Dragorn421, line 312: I feel like this is a very complicated function call for a two-lines function (which can be one line), it would probably be simpler to read if you move the function's content here

fast64_internal/oot/oot_scene_room.py
```diff
@@ -172,13 +172,27 @@ def drawExitProperty(layout, exitProp, index, headerIndex, objName):
 
 class OOTObjectProperty(bpy.types.PropertyGroup):
     expandTab: bpy.props.BoolProperty(name="Expand Tab")
-    objectID: bpy.props.EnumProperty(items=ootEnumObjectID, default="OBJECT_HUMAN")
-    objectIDCustom: bpy.props.StringProperty(default="OBJECT_HUMAN")
+    objectKey: bpy.props.EnumProperty(items=ootData.objectData.ootEnumObjectKey, default="obj_human")
+    objectID: bpy.props.EnumProperty(items=ootData.objectData.ootEnumObjectIDLegacy, default="OBJECT_HUMAN")
+    objectIDCustom: bpy.props.StringProperty(default="OBJECT_CUSTOM")
```
Dragorn421, line 177: can you add a comment on `objectID`, maybe move it down too, about it being legacy

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,40 @@
...
+                    actor.attrib["Name"],
+                    actor.attrib["Category"],
+                    tiedObjects,  # actors don't always use an object
+                )
+            )
+        self.actorsByKey = {actor.key: actor for actor in self.actorList}
+        self.actorsByID = {actor.id: actor for actor in self.actorList}
+        # list of tuples used by Blender's enum properties
+        # ``ootEnumActorIDLegacy`` is there for compatibility with older blends
+        self.ootEnumActorKey, self.ootEnumActorIDLegacy = getEnumList(self.actorList, "Custom Actor")
```
Dragorn421, line 40: for now the "legacy" one is still the one in use, so I'd just name it `ootEnumActorID` and rename to legacy in the other PR, I didn't realize this before

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,89 @@
...
+                collection.add()
+                collection.move(len(collection) - 1, (headerIndex + 1))
+                collection[-1].objectKey = objectKey
+
+    def addMissingObjectsToList(self, roomObj, room, actorData, headerIndex, csHeaderIndex):
+        """Adds missing objects to the object list"""
+        if len(room.actorList) > 0:
+            for roomActor in room.actorList:
+                actor = actorData.actorsByID.get(roomActor.actorID)
+                if not (actor.key == "player") and actor.tiedObjects is not None:
```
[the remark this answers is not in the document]
Yanis42: it's a me issue, ``!=`` is ugly imo lol, I'll change that 👍

fast64_internal/oot/oot_object.py
```diff
@@ -0,0 +1,40 @@
...
+    if roomObj is not None:
+        roomProp = ootGetSceneOrRoomHeader(roomObj, headerIndex, True)
+        if roomProp is not None:
+            collection = roomProp.objectList
+            collection.add()
+            collection.move(len(collection) - 1, (headerIndex + 1))
+            collection[-1].objectKey = objectKey
+
+
+def addMissingObjectsToList(roomObj: Object, room: OOTRoom, ootData: OoT_Data, headerIndex: int):
```
Dragorn421, line 19: I think I'd rename this function to `addMissingObjectsToRoomHeader`
and `addRoomHeadersObjects` to `addMissingObjectsToAllRoomHeaders`
?

fast64_internal/oot/oot_operators.py
```diff
@@ -51,6 +51,7 @@ def execute(self, context):
         emptyObj.name = "Door Actor"
         emptyObj.ootTransitionActorProperty.actor.actorID = "ACTOR_DOOR_SHUTTER"
         emptyObj.ootTransitionActorProperty.actor.actorParam = "0x0000"
+        emptyObj.fast64.oot.version = emptyObj.fast64.oot.cur_version
```
Dragorn421, line 54: I thought about this and what if the user creates an empty and sets its type manually?

Maybe it'd be a better idea to set the default value of the version prop to cur_version, that'd handle all new objects :thinking: 
But then it would mark "up-to-date" really old objects in really old blends that never had this property Sadge

idk...

fast64_internal/oot/data/oot_getters.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def getEnumList(dataList: list[OoT_BaseElement], customItemName: str):
+    """Returns lists containing data for Blender's enum properties"""
+    enumPropItems: list[tuple] = []
+    legacyEnumPropItems: list[tuple] = []  # for older blends
+    customItem = ("Custom", customItemName, "Custom")
+
+    for elem in dataList:
+        enumPropItems.append((elem.key, elem.name, elem.id))
```
Dragorn421, line 22: if we don't use the xml's index to create the items (as "number") then the xml ordering matters for the new object enum items and there's no purpose to the index?

fast64_internal/oot/oot_upgrade.py
```diff
@@ -0,0 +1,26 @@
+from bpy.types import Object, CollectionProperty
+from .data.oot_object_data import OoT_ObjectData
+
+
+def upgradeObjectList(objList: CollectionProperty, objData: OoT_ObjectData):
+    """Transition from the old object system to the new one"""
```
Dragorn421, line 6: "old" and "new" isn't specific enough, everything becomes "old" eventually
probably refer to the object-id-enum vs object-key-enum-from-xml instead

fast64_internal/oot/oot_level_writer.py
```diff
@@ -86,6 +88,11 @@ def ootExportSceneToC(
         if exportSubdir == "":
             raise PluginError("Scene folder " + sceneName + " cannot be found in the ootSceneDirs list.")
 
+    roomObjects = [
+        obj for obj in bpy.data.objects if ((obj.parent == originalSceneObj) and (obj.ootEmptyType == "Room"))
+    ]
+    for roomObj, room in zip(roomObjects, scene.rooms.values()):
```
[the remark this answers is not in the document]
Yanis42: I have no idea lol

fast64_internal/oot/oot_object.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def addMissingObjectToProp(roomObj: Object, headerIndex: int, objectKey: str):
+    """Add the missing object to the room empty object OoT object list"""
+    if roomObj is not None:
+        roomProp = ootGetSceneOrRoomHeader(roomObj, headerIndex, True)
+        if roomProp is not None:
+            collection = roomProp.objectList
+            collection.add()
+            collection.move(len(collection) - 1, (headerIndex + 1))
```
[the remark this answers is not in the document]
Yanis42: idk either, I just thought it'd be useful since it's being executed when you add an object (see ``OOTCollectionAdd``)

fast64_internal/oot/oot_operators.py
Yanis42: if someone adds an empty manually then it'd be set the next time they'll open the blend I think

fast64_internal/oot/data/oot_getters.py
```diff
@@ -0,0 +1,27 @@
...
+        legacyEnumPropItems.append((elem.id, elem.name, elem.id))
+
+    enumPropItems.insert(0, customItem)
+    legacyEnumPropItems.insert(0, customItem)
+    return enumPropItems, legacyEnumPropItems
```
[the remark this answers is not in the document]
Yanis42: I thought the same then figured it'd be harmless to keep it, I'll need this anyway later (when I'll re-work again the actor panel)

fast64_internal/oot/data/oot_getters.py
Yanis42: when I did this I tried with a messed up XML, like I switched a lot of lines (for instance I had like object number 19 on the first entry) and it worked properly 👍

Yanis42: maybe I should remove this function for now, it's a bit useless since the object list need something different

fast64_internal/oot/oot_level_writer.py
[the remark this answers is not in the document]
Dragorn421: I figure it's pure luck, this will eventually lead to weird bugs

See `addRoom` for how roomObj are mapped to `OOTRoom`s

This is a weird operation because exporting goes top down from blender objs and props to C, and this goes back up (edits props too), I don't see how to fix this cleanly

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,51 @@
...
+        lastIndex = self.objectsByKey["obj_timeblock"].index
+        self.ootEnumObjectIDLegacy = [None] * lastIndex
+        self.ootEnumObjectIDLegacy.insert(0, ("Custom", "Custom Object", "Custom"))
+        for obj in self.objectList:
+            if obj.index < lastIndex + 1:
+                self.ootEnumObjectIDLegacy[obj.index] = (obj.id, obj.name, obj.id)
+
+        # validate the legacy list, if there's any None element then something's wrong
+        if None in self.ootEnumObjectIDLegacy:
+            raise PluginError("ERROR: Legacy Object List doesn't match!")
```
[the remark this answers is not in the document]
Yanis42: this check makes sense since we want the previous and hardcoded object list, so yea I think it should stay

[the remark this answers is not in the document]
Dragorn421: we only want it in the case the user has old blends, but it'd prevent fast64 from enabling even if they were starting from scratch and just deleted an object in the xml that they also removed in their repo

again this can be addressed later if needed though, just something to be aware of

Yanis42:
found and fixed a minor issue in ``ootConvertScene()``, ``for obj in sceneObj.children`` doesn't look for objects parented to a scene but also parented to something else, fixed by using ``children_recursive`` (``for obj in sceneObj.children_recursive``)

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,49 @@
...
+                    int(actor.attrib["Index"]),
+                    actor.attrib["Category"],
+                    tiedObjects,
+                )
+            )
+        self.actorsByKey = {actor.key: actor for actor in self.actorList}
+        self.actorsByID = {actor.id: actor for actor in self.actorList}
+
+        # list of tuples used by Blender's enum properties
+        lastIndex = max(1, *(int(actor.attrib["Index"]) for actor in actorRoot.iterfind("Actor")))
```
Dragorn421, line 44:
```suggestion
        lastIndex = max(1, *(actor.index for actor in self.actorList))
```

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,49 @@
...
+        self.ootEnumActorID = [("None", "(Deleted from the XML)", "None")] * lastIndex
```
Dragorn421, line 45:
```suggestion
        self.ootEnumActorID = [("None", f"{i} (Deleted from the XML)", "None") for i in range(lastIndex)]
```

fast64_internal/oot/data/oot_actor_data.py
```diff
@@ -0,0 +1,49 @@
...
+        self.ootEnumActorID.insert(0, ("Custom", "Custom Actor", "Custom"))
+        for actor in self.actorList:
+            if actor.index < lastIndex + 1:
+                self.ootEnumActorID[actor.index] = (actor.id, actor.name, actor.id)
```
Dragorn421, line 49: the if isn't needed

fast64_internal/oot/data/oot_object_data.py
```diff
@@ -0,0 +1,55 @@
...
+            objName = f"{obj.attrib['Name']} - {obj.attrib['ID'].removeprefix('OBJECT_')}"
+            self.objectList.append(
+                OoT_ObjectElement(obj.attrib["ID"], obj.attrib["Key"], objName, int(obj.attrib["Index"]))
+            )
+
+        self.objectsByID = {obj.id: obj for obj in self.objectList}
+        self.objectsByKey = {obj.key: obj for obj in self.objectList}
+
+        # list of tuples used by Blender's enum properties
+        self.customEntry = ("None", "(Deleted from the XML)", "None")
```
Dragorn421, line 36: customEntry?

fast64_internal/oot/oot_level_writer.py
```diff
@@ -296,8 +305,14 @@ def readRoomData(room, roomHeader, alternateRoomHeaders):
...
     room.echo = roomHeader.echo
-    room.objectList.extend([getCustomProperty(item, "objectID") for item in roomHeader.objectList])
-    if len(room.objectList) > 15:
+    objectsByKey = ootData.objectData.objectsByKey
+    room.objectIDList.extend(
+        [
+            objectsByKey.get(obj.objectKey).id if obj.objectKey != "Custom" else obj.objectIDCustom
+            for obj in roomHeader.objectList
+        ]
+    )
```
Dragorn421, line 314:
```suggestion
    room.objectIDList.extend(
        
            objectsByKey[obj.objectKey].id if obj.objectKey != "Custom" else obj.objectIDCustom
            for obj in roomHeader.objectList
        
    )
```

Yanis42:
merged main, is there anything holding this PR from being merged?

kurethedead (changes requested):
I've tested by exporting a kakariko import and it seems to work fine. I'm not sure exactly where the object list updating is occurring, but I think something needs to communicate to the user how many objects are actually being exported, so they don't get confused when something breaks unexpectedly.

sauraen (approved):
Approved. Please merge ASAP.

Tested and it does what it says. I was concerned about support for non-decomp custom actors/objects (which only have a number), but it works and exports them.

One annoyance--when making actors (by adding an empty and changing its type to Actor), or when making other types of objects by this method (even though for scenes and rooms there's an operator available to make them), it says the data is out of date and gives the upgrade button. It works after clicking the button, but things shouldn't be created out of date. Changing the dropdown for the empty type should also do the initialization.

fast64_internal/oot/oot_level_writer.py
```diff
@@ -499,7 +518,7 @@ def ootConvertScene(originalSceneObj, transformMatrix, f3dType, isHWv1, sceneNam
     if bpy.context.scene.exportHiddenGeometry:
         hideObjsInList(hiddenObjs)
 
-    roomObjs = [child for child in sceneObj.children if child.data is None and child.ootEmptyType == "Room"]
```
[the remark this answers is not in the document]
sauraen: How about a check and error if the hierarchy is not correct (room in room, actor directly in scene, etc.)?

Dragorn421:
fwiw I don't remember having issues with this PR last time I was actively exchanging about it with Yanis, besides a bit of jank around data upgrading which is tied to the issue mentioned by Sauraen about empty type switching.

So once you guys are happy I +1 merging

fast64_internal/oot/oot_actor.py
```diff
@@ -97,9 +97,9 @@ class OOT_SearchActorIDEnumOperator(bpy.types.Operator):
     bl_property = "actorID"
     bl_options = {"REGISTER", "UNDO"}
 
-    actorID: bpy.props.EnumProperty(items=ootEnumActorID, default="ACTOR_PLAYER")
-    actorUser: bpy.props.StringProperty(default="Actor")
-    objName: bpy.props.StringProperty()
+    actorID : bpy.props.EnumProperty(items = ootData.actorData.ootEnumActorID, default = "ACTOR_PLAYER")
```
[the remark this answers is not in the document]
Yanis42: the order should be the same

fast64_internal/oot/oot_level_writer.py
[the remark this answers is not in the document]
Yanis42: I don't understand why the diff on github shows ``sceneObj.children``, it should be ``sceneObj.children_recursive``, which checks every objects, so this won't be an issue at all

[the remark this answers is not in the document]
Yanis42: ok I booted my brain, I changed ``children`` to ``children_recursive``, the difference between these two: children don't go through child objects, children_recursive do

so if a room has a weird hierarchy this will check the object, so I don't think it could cause issues

fast64_internal/oot/oot_level_writer.py
```diff
@@ -744,45 +763,47 @@ def ootProcessEmpties(scene, room, sceneObj, obj, transformMatrix):
...
-                    actorProp.actorParam,
-                    None
-                    if not actorProp.rotOverride
-                    else (actorProp.rotOverrideX, actorProp.rotOverrideY, actorProp.rotOverrideZ),
-                ),
-                actorProp,
-                "actorList",
-                obj.name,
-            )
+            if actorProp.actorID != "None":
```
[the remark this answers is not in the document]
Yanis42: ``ootEnumActorID`` it filled with ``("None", f"{i} (Deleted from the XML)", "None")`` for every actors, since Blender saves the index of the list when you have an enum prop, you can change the identifier without breaking anything, so this will prevent a data loss if the user deletes an actor in the XML file

To answer your question, we're just checking if the ID is not "None" for that case when the user deleted the current Actor from the list

Yanis42:
I can add an "Add Actor" button in the tools panel but if the user adds an actor manually this issue will stay and idk how to fix that properly

fast64_internal/oot/oot_level_writer.py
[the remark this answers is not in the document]
kurethedead: Ah gotcha, makes sense. You should add a comment here with that explanation.

fast64_internal/oot/oot_level_writer.py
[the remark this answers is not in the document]
kurethedead: Sorry my bad, I think I meant to highlight the line of code underneath this one. My issue is that children_recursive should not be used, because having a room nested in another room doesn't make sense and would cause strange behaviour.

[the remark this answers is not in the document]
Yanis42: I think it'd be better to raise an error if the user tries to export with a room inside a room (and eventually add a warning for that on the UI), because there's the case when you have a room inside an empty with the None type and this wouldn't export that room (unless going through bpy.data but it'd be the same thing in the end)

I understand your concern though, but I can't think of a solution without children_recursive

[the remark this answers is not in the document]
kurethedead: Personally I think rooms being a child of an None type shouldn't be processed either, but if people want to use it that way then I guess its fine.

kurethedead:
I've made a comment in discord about a way to fix sauraen's problem with object versioning. Otherwise, I still think there should be something that ensures that the user is aware of when objects are auto-added to the object list (error-checking, warning UI?) especially if those added objects would cause a crash.

Yanis42:
I think I understand how this will fix the issue but I'm not sure how to implement that 🤔

kurethedead (commented):
Right now the PR uses the object properties version to handle updates to the OOT object list enum, but its broken currently (#179). The versioning system I suggested in discord also doesn't work, as I realized that old properties that haven't had their default changed are also not accessible in the way I suggested. As a workaround, we can apply the same concept to the `objectID` property on the `OOTObjectProperty` itself. Remove the `objectID` definition, then in code the presence of `objProp["objectID"]` indicates an un-upgraded property. The property can be updated, then `del objProp["objectID"]` is called. The exporter can handle both upgraded and un-upgraded versions using this check, so that even if the user doesn't update the export will still be correct.

Yanis42:
I understand how your idea is working but I don't know how to implement that

Yanis42:
Updated the upgrade system with Kure's suggestions, also I thought it'd be fine if I removed the version system completely since the property trick works better, now we just need to figure out a standard to do this but it can be done later

Changed the UI to make it draw, if it's still using old data everything on the object list will be disabled but visible with the upgrade button at the top, also exporting should work properly if the blend isn't updated yet

kurethedead (approved):
Looks good to me. Will merge soon if unless anyone objects.

PR #97, opened 2022-05-26
Prevent Fast64 from enabling in Blender 3.1.0+

Dragorn421:
Can't enable Fast64 anymore in Blender 3.1.0+ (raises an exception with an hopefully friendly enough message)

Any suggestion on the message?

Note: I can't test in Blender < 3.1.0 I hope this works as intended (do nothing) in those versions :cold_sweat:

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -357,6 +357,23 @@ def after_load(_a, _b):
 # register operators and panels here
 # append menu layout drawing function to an existing window
 def register():
+
+	if bpy.app.version >= (3, 1, 0):
+		import types
+
+		msg = "\n".join(
+			(
+				"This version of Fast64 does not work properly in Blender 3.1.0 and later Blender versions.",
+				"Your Blender version: " + ".".join(str(i) for i in bpy.app.version),
+				"This is a known issue, the fix is not trivial and is in progress.",
+				"See the GitHub issue: https://github.com/Fast-64/fast64/issues/85",
+				"If it has been resolved, update Fast64.",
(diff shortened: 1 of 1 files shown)


thecozies (commented):
this is awesome, i will test it here in a few

__init__.py
```diff
@@ -357,6 +357,23 @@ def after_load(_a, _b):
 # register operators and panels here
 # append menu layout drawing function to an existing window
 def register():
+
+	if bpy.app.version >= (3, 1, 0):
+		import types
```
thecozies, line 362: why import types here?

__init__.py
```diff
@@ -357,6 +357,23 @@ def after_load(_a, _b):
...
+
+		msg = "\n".join(
+			(
+				"This version of Fast64 does not work properly in Blender 3.1.0 and later Blender versions.",
+				"Your Blender version: " + ".".join(str(i) for i in bpy.app.version),
```
thecozies, line 367: one small word 
```suggestion
				"Your Blender version is: " + ".".join(str(i) for i in bpy.app.version),
```

thecozies (changes requested):
one tiny word and one tiny question!

tested in 3.0, i was able to install successfully!

__init__.py
Dragorn421: because I'm a moron who doesn't re-read his code

thecozies (approved):
###### _it's amazing_

PR #98, opened 2022-05-29
update readme with more info about actors (oot)

krm01:
Adds a bit of clarity around getting actors to work properly in scenes.

### DIFF
--- a/fast64_internal/oot/README.md
+++ b/fast64_internal/oot/README.md
@@ -23,7 +23,10 @@ Properties for empties/meshes can be found in the Blender object properties wind
 
 To export a scene, the "Scene Object" must be set in the "OOT Scene Exporter" section in the tool properties sidebar. When you click "Add Scene" this is set automatically.
 
-### Actors Variables
+### Actors
+To add actors to a scene, create a new Empty and parent it to a Room. The in the Object Properties panel select "Actor" as the Object Type. Use the "Select Actor ID" button to choose an actor, and then set the Actor Parameter value as desired (see the list of Actor Variables below). Finally, all actors you are using must be added to the parent Room's object dependecies, otherwise they will not load into the scene at all. To do this select the Room that your actor is parented to, select the "Objects" tab in its Object Properties window, and click "Add Item". Then "Search Object ID" to find the actor object you need. For example, if adding a Deku Baba actor (EN_DEKUBABA) you need to add the "Dekubaba" object to the Room's object dependencies.
+
+#### Actor Variables
 Actor variables can be found at https://wiki.cloudmodding.com/oot/Actor_List_(Variables).
 
 ### Exits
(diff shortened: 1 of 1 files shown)


Yanis42:
I did something similar not so long ago but I never PR'd it since I wasn't really satisfied, feel free to copy pasta the infos from there to be more accurate: https://github.com/Yanis42/fast64/blob/oot-readme-update/fast64_internal/oot/README.md

If this file is changed then let's do a good update (part of the reason why I never PR'd this)

fast64_internal/oot/README.md
```diff
@@ -23,7 +23,10 @@ Properties for empties/meshes can be found in the Blender object properties wind
 
 To export a scene, the "Scene Object" must be set in the "OOT Scene Exporter" section in the tool properties sidebar. When you click "Add Scene" this is set automatically.
 
-### Actors Variables
+### Actors
+To add actors to a scene, create a new Empty and parent it to a Room. The in the Object Properties panel select "Actor" as the Object Type. Use the "Select Actor ID" button to choose an actor, and then set the Actor Parameter value as desired (see the list of Actor Variables below). Finally, all actors you are using must be added to the parent Room's object dependecies, otherwise they will not load into the scene at all. To do this select the Room that your actor is parented to, select the "Objects" tab in its Object Properties window, and click "Add Item". Then "Search Object ID" to find the actor object you need. For example, if adding a Deku Baba actor (EN_DEKUBABA) you need to add the "Dekubaba" object to the Room's object dependencies.
```
Dragorn421, line 27:
```suggestion
To add actors to a scene, create a new Empty and parent it to a Room. Then in the Object Properties panel select "Actor" as the Object Type. Use the "Select Actor ID" button to choose an actor, and then set the Actor Parameter value as desired (see the list of Actor Variables below). Finally, all actors you are using must be added to the parent Room's object dependencies, otherwise they will not load into the room at all. To do this select the Room that your actor is parented to, select the "Objects" tab in its Object Properties window, and click "Add Item". Then "Search Object ID" to find the actor object you need. For example, if adding a Deku Baba actor (EN_DEKUBABA) you need to add the "Dekubaba" object to the Room's object dependencies.
```

Dragorn421 (commented):
Thanks for taking the time to write some doc

Note the following two PRs
- #56
- #95

, improve a lot on adding actors, e.g. with a "Add Actor" button, a UI for picking actor params, automatically adding objects to the room's object list

In the mean time this can still be merged. Just a few things

fast64_internal/oot/README.md
krm01: Yikes, that's what I get for editing in a terminal without a spell check. thanks. I did not see the other PRs, those improvements are extremely helpful. great stuff. I'll have to test them out tomorrow.

Dragorn421 (approved):
As I said this will be outdated eventually as Fast64 improves (thankfully), but good to have for now

(Yanis, PR your readme zfgMad)

Yanis42:
I'll PR it when both PRs 56 and 95 will be merged so we do only one commit to change that file

PR #100, opened 2022-06-10
[OoT] Fixed wrong C type for mesh and other small changes

Yanis42:
## Things that change with this PR:
* fixed outdated mesh type: ``MeshEntry0`` -> ``Mesh01Entry`` (same thing for mesh type 2)
* removed useless ``(u32)`` casts as it's not necessary to most of those
* removed useless ``&`` since using an array name is the adress to the first element (some u32 casts prevented warnings at compilation)
* export draw config enum name, not really needed but it looks better to use those (I wasn't sure about the implementation)
* enhanced ``MeshHeader``0/2 code and output

Also sorry for having all of this on one commit, I made those changes without thinking about the history/review

<details closed>
<summary>More details about the ``&`` thing</summary>

```c
// this might not actually work but it's just an example
// to understand what I've said about using the array name instead of ``&``
void example() {
    char my_array[] = { 'A', 'B' };
    short example1, example2, adrBool;

    example1 = &my_array[0];
    example2 = my_array;

    adrBool = (example1 == example2);

    printf("adrBool is true! %X\n", adrBool);
}
```
* there's other usage like that but it's not really useful to explain there
</details>

PS: yes I know there's stuff I could have changed like removing every ``\t`` but it's already in another PR


fast64_internal/oot/oot_collision.py
```diff
@@ -383,11 +383,11 @@ def ootCameraPosToC(camPos):
...
-	return "{ " +\
-		str(camPos.camSType) + ', ' +\
-		('3' if camPos.hasPositionData else '0') + ', ' +\
-		(("&" + camData.camPositionsName()  + '[' + str(camPosIndex) + ']') 
-		if camPos.hasPositionData else "0") + ' },\n'
+	return " ".join((
+		f"{'{'} {camPos.camSType},",
+		f"{'3' if camPos.hasPositionData else '0'},",
+		f"{f'&{camData.camPositionsName()}[{camPosIndex}]' if camPosIndex > 0 else camData.camPositionsName()} {'}'}\n"
+	))
```
Dragorn421, line 390: please change the `f"{'{'}"` stuff to something more sane :sweat_smile: like
```py
	return " ".join((
		"{",
		f"{camPos.camSType},",
		"3" if camPos.hasPositionData else "0",
		",",
		f"&{camData.camPositionsName()}[{camPosIndex}]" if camPosIndex > 0 else camData.camPositionsName(),
		"}\n",
	))
```

also what does the `if camPosIndex > 0` ternary achieve? if it's 0 it just should be `&...[0]`, or can it be < 0

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -195,18 +195,23 @@ def ootRoomMeshToC(room, textureExportSettings):
...
+		f"{indent}{len(mesh.meshEntries)},",
+		f"{indent}{mesh.entriesName()},",
+		f"{indent}{mesh.entriesName()} + sizeof({mesh.entriesName()})",
+		f"{'}'};"
+	)) + "\n\n"
 
 	meshEntries = CData()
-	meshEntries.header = "extern MeshEntry" + str(mesh.meshType) + " " + mesh.entriesName() + "[" + str(len(mesh.meshEntries)) + "];\n"
-	meshEntries.source = "MeshEntry" + str(mesh.meshType) + " " + mesh.entriesName() + "[" + str(len(mesh.meshEntries)) + "] = {\n"
+	meshEntryType = f"MeshHeader{'01' if not int(mesh.meshType, base=10) else '2'}Entry "
```
Dragorn421, line 209:
```suggestion
	meshEntryType = f"MeshHeader{'01' if mesh.meshType == '0' else '2'}Entry "
```
(what about mesh type 1 then, or is it handled somewhere else?)

Dragorn421 (commented):
note the polygon* types are only there temporarily for ZAPD compatibility https://github.com/zeldaret/oot/blob/826924afeae9c336f0783b502b6d36fc71c06a66/include/z64scene.h#L153
```c
// TODO update ZAPD
typedef MeshHeader01Entry PolygonDlist;
typedef MeshHeader0 PolygonType0;
typedef MeshHeader2Entry PolygonDlist2;
typedef MeshHeader2 PolygonType2;
```

they should be the MeshHeader structs

but those MeshHeader structs will also be renamed soon :tm: so it's not urgent either

fast64_internal/oot/c_writer/oot_level_c.py
Yanis42: mesh type 1 isn't supported atm since, I assume, pre-rendered backgrounds aren't extracted yet, since there's a PR opened for that I think we should implement type 1

also could be cool to export skyboxes (idk how it works tho so idk if that'd be useful or not)

fast64_internal/oot/oot_collision.py
Yanis42: ``&...[0]`` looks weird imo, using the array name looks cleaner to me but yea it's the same thing so I guess it doesn't matter

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -195,18 +195,23 @@ def ootRoomMeshToC(room, textureExportSettings):
...
-		"{ {" + str(mesh.meshType) + "}, " + str(len(mesh.meshEntries)) + ", " +\
-		"(u32)&" + mesh.entriesName() + ", (u32)&(" + mesh.entriesName() + ") + " +\
-		"sizeof(" + mesh.entriesName() + ") };\n\n"
+	meshHeader.header = "extern MeshHeader" + mesh.meshType + " " + mesh.headerName() + ";\n"
+	meshHeader.source = '\n'.join((
+		"MeshHeader" + mesh.meshType + " " + mesh.headerName() + "= {",
+		indent + mesh.meshType + ",",
+		indent + str(len(mesh.meshEntries)) + ",",
+		indent + mesh.entriesName() + ",",
+		indent + mesh.entriesName() + " + sizeof(" + mesh.entriesName() + ")",
```
Dragorn421, line 204:
```suggestion
		indent + mesh.entriesName() + " + ARRAY_COUNT(" + mesh.entriesName() + ")",
```
with pointer arithmetics, `meshEntries + sizeof(meshEntries)` would be `(u32)meshEntries + sizeof(MeshHeaderEntry)*sizeof(meshEntries)` instead of `(u32)meshEntries + sizeof(meshEntries)`

array_count if it works would be the expected behavior

note this isn't critical as the entries end is not used by the game but may as well get it right

fast64_internal/oot/c_writer/oot_level_c.py
```diff
@@ -195,18 +195,23 @@ def ootRoomMeshToC(room, textureExportSettings):
...
+		"};"
+	)) + "\n\n"
 
 	meshEntries = CData()
-	meshEntries.header = "extern MeshEntry" + str(mesh.meshType) + " " + mesh.entriesName() + "[" + str(len(mesh.meshEntries)) + "];\n"
-	meshEntries.source = "MeshEntry" + str(mesh.meshType) + " " + mesh.entriesName() + "[" + str(len(mesh.meshEntries)) + "] = {\n"
+	meshEntryType = "MeshHeader" + ('01' if mesh.meshType == '0' else '2') + "Entry "
+	meshEntries.header = "extern " + meshEntryType + mesh.entriesName() + "[" + str(len(mesh.meshEntries)) + "];\n"
```
Dragorn421, line 210:
```suggestion
	meshEntryType = "MeshHeader" + ('01' if mesh.meshType == '0' else '2') + "Entry"
	meshEntries.header = "extern " + meshEntryType + " " + mesh.entriesName() + "[" + str(len(mesh.meshEntries)) + "];\n"
```

fast64_internal/oot/c_writer/oot_level_c.py
Yanis42: oh yea, I haven't thought about using array_count there

Dragorn421 (approved):

PR #101, opened 2022-06-10
update CameraSettingType enum to match decomp

krm01:
I tested with a handful of camera setting types and they worked. List matches as of https://github.com/zeldaret/oot/blob/2e6279bc8e528833704bc6b77635f0eeae6bbe44/include/z64camera.h#L30 in decomp.


fast64_internal/oot/oot_collision_classes.py
```diff
@@ -93,65 +93,65 @@
...
-    ("CAM_SET_TOWER1", "Tower1", "Tower1"),
-    ("CAM_SET_FIXED0", "Fixed0", "Fixed0"),
-    ("CAM_SET_FIXED1", "Fixed1", "Fixed1"),
-    ("CAM_SET_CIRCLE0", "Circle0", "Circle0"),
-    ("CAM_SET_CIRCLE2", "Circle2", "Circle2"),
-    ("CAM_SET_CIRCLE3", "Circle3", "Circle3"),
-    ("CAM_SET_PREREND0", "Prerend0", "Prerend0"),
-    ("CAM_SET_PREREND1", "Prerend1", "Prerend1"),
-    ("CAM_SET_PREREND3", "Prerend3", "Prerend3"),
+    ("CAM_SET_HORSE", "Horse", "Horse"),
```
Dragorn421, line 96: the first item in each tuple is the identifier for the enum item, it can't be changed without breaking older .blend files unfortunately. The rest is fine to change as you wish

For more information see https://docs.blender.org/api/current/bpy.props.html?highlight=items#bpy.props.EnumProperty

So to update these names we would need a identifier -> decomp name mapping to use after `getCustomProperty(camPosProp, 'camSType')`

@ me on the fast64 discord or DMs if needed

krm01: @Dragorn421 Oh didn't consider that. Updated in f4065cb with your suggestion.

Yanis42:
I was wondering, could it be worth to use descriptive names/descriptions? afaik it shouldn't matter since the first argument of a tuple is the identifier in blender's context

I'm not sure what this feature is about though

krm01:
@Yanis42 decomp renamed their camera enums, for example `CAM_SET_HORSE0` is now `CAM_SET_HORSE` (https://github.com/zeldaret/oot/blob/2e6279bc8e528833704bc6b77635f0eeae6bbe44/include/z64camera.h#L37) 

What this does now is convert the old enum name to the new one, so it actually compiles in decomp. as far as descriptions... yeah it could be handy I suppose, but choosing a camera type is already a pretty rough UI:
image

adding more would make it worse without changing how this behaves. In the meantime it's easy to reference the decomp `z64camera.h` file to read the descriptions there.

Yanis42:
> @Yanis42 decomp renamed their camera enums, for example `CAM_SET_HORSE0` is now `CAM_SET_HORSE` (https://github.com/zeldaret/oot/blob/2e6279bc8e528833704bc6b77635f0eeae6bbe44/include/z64camera.h#L37)
> 
> What this does now is convert the old enum name to the new one, so it actually compiles in decomp. as far as descriptions... yeah it could be handy I suppose, but choosing a camera type is already a pretty rough UI: image
> 
> adding more would make it worse without changing how this behaves. In the meantime it's easy to reference the decomp `z64camera.h` file to read the descriptions there.

I was just wondering if it could be good to get rid of names like "normal0", "cs_3" etc and rename those to have a name that tells you what is it doing, or at least make a real description (the 3rd argument of the tuple) that can be seen when you let the cursor for a sec over one element

Like, we don't need to follow the decomp for that as long as the identifier stays faithful to whatever's on decomp side

Dragorn421:
If there are better names to use go for it, the names from decomp are not thought to be final (just doing their best based on what knowledge is there and how much time is put into it), camera is a very large system and documentation is ongoing

Camera settings describe how camera modes map to camera functions and camera function data, it's not trivial to document (z_camera is 8.2k lines, z_camera_data 2.4k, and camera interacts with a wide variety of systems like collision, cutscenes, player... :sweat_smile: )

The mapping from old names to new names on export looks good, I'll try to test soon

Yanis42:
I feel like we should try to document everything we can and do another pass later when the camera system will be more understood

Also, I was thinking, would it be worth to use external format like JSON or XML or whatever? I became addict to those files sorry MiniK

I'll see if I can understand some of the names

Dragorn421:
I don't think the complexity of handling an external file/format is worth it in this case, if you are modding camera settings you can definitely figure out how to mod fast64 too :smile: (or there's always just "custom" to handle that case)

Dragorn421 (approved):
I fixed what I think was an issue, if the value was custom it would still be mapped through `decomp_compat_map_CameraSType` which sounds undesirable

Dragorn421:
@Yanis42 @krm01 can you review?

Yanis42 (approved):

PR #104, opened 2022-06-21
Post-format cleanup of f3d_gbi.py

Dragorn421:
plus manually cleaning up stuff, mostly moving things to str.join()


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -1789,20 +1794,14 @@ def vertexScrollTemplate(
...
-            + "\tint height = "
-            + str(fScrollData.dimensions[1])
-            + " * 0x20;"
+        data = "\n".join(
+            (
+                "void scroll_" + name + "() {",
+                "\tint i = 0;",
+                f"\tint count = {count};",
+                f"\tint width = {fScrollData.dimensions[0]} * 0x20;",
+                f"\tint height = {fScrollData.dimensions[1]} * 0x20;",
```
thecozies, line 1803: ahhhh this is mostly a note to self, but this generated code can cause compiler warnings for unused variables, plus these aren't ever changed so they could be `const`

Dragorn421: this is a Python cleanup PR get your bad C out of here :japanese_goblin:

thecozies: lmaooo

thecozies (approved):
# much needed

PR #106, opened 2022-06-26
Fix OoT skeleton import failing from `LIMB_DONE`

Dragorn421:
at some point (not recently) the `0xFF` value for "no limb next" was changed to use `LIMB_DONE`, somehow this hasn't been noticed earlier and as a result fast64 currently can't import oot skeletons


Yanis42 (approved):
💀

thecozies (approved):
###### that's neat

PR #108, opened 2022-06-30
applied black formatting to files changed in the future

thecozies:
These are all of the files that I touched in V5 that need to be formatted.


pyproject.toml
```diff
@@ -6,8 +6,7 @@
 line-length = 120
 
 target-version = [
-    'py37', # used by Blender 2.80
-    'py39', # used by Blender 3.0.1
+    'py310', # used by Blender 3.1
```
Dragorn421, line 9: probably should be changed later, for now the repo is still for "old" blender?

fast64_internal/f3d/f3d_material_nodes.py
```diff
@@ -5,1123 +5,1122 @@
 from .f3d_enums import *
```
Dragorn421, line 5: this file is deleted in the #109 diff, may as well not format it?

thecozies: i formatted it since that original format commit had done that and i was avoiding conflicts haha

pyproject.toml
thecozies: well the formatter needs to be 3.10 supported, plus this is going into the V5 branch, i figured we would merge the formatting changes into master until we were ready with V5

PR #114, opened 2022-07-02
Format oot_level_classes.py

Dragorn421:


thecozies (approved):
# unk unk unk unk unk unk unk unk unk

PR #118, opened 2022-07-03
Format/cleanup oot_level_c.py

Dragorn421:
Missed this one (*bleh*)


thecozies (approved):
lgtm, though this reminds me that we should really have a library for joining *args and other C stuff even though it's simple to do anyways

PR #121, opened 2022-07-08
fixed exporting with modifiers that rely on other objects

thecozies:
When working on a level, I was using boolean modifiers in order to create some intersecting geometry that I would be able to change on the fly. Turns out that is broken because the modifiers weren't applied to the original object position. This fixes that issue.

PRed to V5 to avoid more merge issues and V5 is imminent (right?)

### DIFF
--- a/fast64_internal/utility.py
+++ b/fast64_internal/utility.py
@@ -816,6 +816,23 @@ def get_obj_temp_mesh(obj):
         if o.get("temp_export") and o.get("instanced_mesh_name") == obj.get("instanced_mesh_name"):
             return o
 
+def apply_objects_transformations(allObjs: Iterable[bpy.types.Object]):
+    # first apply modifiers so that any objects that affect each other are taken into consideration
+    for selectedObj in allObjs:
+        bpy.ops.object.select_all(action="DESELECT")
+        selectedObj.select_set(True)
+        bpy.context.view_layer.objects.active = selectedObj
+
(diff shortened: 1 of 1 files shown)


thecozies:
sounds good, i do need to make one more update to this as well, just found this morning

thecozies:
updated: this needed to be applied to collision and dl exports as well

fast64_internal/utility.py
```diff
@@ -754,6 +755,23 @@ def get_obj_temp_mesh(obj):
         if o.get("temp_export") and o.get("instanced_mesh_name") == obj.get("instanced_mesh_name"):
             return o
 
+def apply_objects_transformations(allObjs: Iterable[bpy.types.Object]):
```
Dragorn421, line 758:
```suggestion
def apply_objects_modifiers_and_transformations(allObjs: Iterable[bpy.types.Object]):
```
ok this isn't the best name but at least it's 100% descriptive rather than 50% pepega

Dragorn421: @thecozies madge you don't like my name?
if no you can merge this

thecozies: oops! i like the name, just didnt see your comment, gonna post another update to this PR

thecozies:
@Dragorn421 updated with your suggestion! im fine with merging this whenever you are.

Dragorn421 (approved):

PR #123, opened 2022-07-28
Add option to boot directly to a scene in HackerOOT

kurethedead:
For scene exporting, adds an option to boot directly to the scene on startup, with spawn/header options. Also adds an operator to undo this (may be less robust), and a confirmation dialog for removing scenes in general.


krm01:
fwiw I would have loved this option when I first started using Fast64, but now that I'm actually using it for more than just experimenting, I am never _not_ using the custom export path and managing it all myself on the decomp side.

kurethedead:
Yeah, I think making the process as painless as possible for newbies is important. I did end up adding more to the pr than I thought I would, which made me re-evaluate my position on it. Making a patch / using a hacker specific fork is probably a better idea.

kurethedead:
I've removed all the patching code, and replaced it with a scene bootup config file export for HackerOOT. I've also added a HackerOOT toggle to show/hide specific features.

fast64_internal/oot/c_writer/oot_scene_bootup.py
```diff
@@ -0,0 +1,139 @@
+import os, re, bpy
+import string
+from bpy.utils import register_class, unregister_class
+from ...utility import PluginError, writeFile
+from ..oot_constants import ootEnumHeaderMenuComplete
+from typing import Callable, Iterable, Any, List
+
+
+def setBootupScene(exportPath: str, entranceIndex: str, options: Any):
```
Dragorn421, line 9: fyi for type annotation with types not yet defined, you can use strings:
```suggestion
def setBootupScene(exportPath: str, entranceIndex: str, options: "OOTBootupSceneOptions"):
```

fast64_internal/oot/c_writer/oot_scene_bootup.py
```diff
@@ -0,0 +1,139 @@
...
+    bootToScene: bpy.props.BoolProperty(default=False, name="Boot To Scene")
+    overrideHeader: bpy.props.BoolProperty(default=False, name="Override Header")
+    headerOption: bpy.props.EnumProperty(items=ootEnumHeaderMenuComplete, name="Header", default="Child Day")
+    spawnIndex: bpy.props.IntProperty(name="Spawn", min=0)
+    newGameOnly: bpy.props.BoolProperty(default=False, name="Override Scene On New Game Only")
+    newGameName: bpy.props.StringProperty(default="Link", name="New Game Name")
+    bootMode: bpy.props.EnumProperty(default="Play", name="Boot Mode", items=ootEnumBootMode)
+
+    # see src/code/z_play.c:Play_Init() - can't access more than 16 cutscenes?
+    cutsceneIndex: bpy.props.IntProperty(min=4, max=19, default=4, name="Cutscene Index")
```
Dragorn421, line 127: yeah, there's no way to load the 17th cutscene header if a scene has one, due to how `cutsceneIndex` values are (0xFFF0-0xFFFF)
I don't know exactly what's going on with this stuff though
there's also this weird bit in play_init
```c
if (gSaveContext.cutsceneIndex == 0xFFFD) {
    gSaveContext.cutsceneIndex = 0;
}
```
which'd suggest you can't use the 14th cutscene header (index 0xD=13)
:shrug: not a big deal for the purpose of this tool

(and fwiw it'd make more sense that this value starts at 0, like "cutscene 0")

fast64_internal/oot/oot_level_writer.py
```diff
@@ -673,7 +680,7 @@ def ootRemoveSceneC(exportInfo):
 
 class OOT_RemoveScene(bpy.types.Operator):
     bl_idname = "object.oot_remove_level"
-    bl_label = "Remove Scene"
+    bl_label = "Are you sure you want to remove this level?"
```
Dragorn421, line 683:
```suggestion
    bl_label = "Are you sure you want to remove this scene?"
```
("level" isn't really used in oot)

A better way to do this may be to implement operator.draw and use a layout.label(text=) to print this warning, bl_label isn't really intended for this
For example this message shows up in operator search
image

fast64_internal/oot/c_writer/oot_scene_bootup.py
kurethedead: Yeah, I thought about 0 indexing, but at this point the first index=4 appears so many times in the UI/codebase that I thought it would be better to keep it consistent.

fast64_internal/oot/__init__.py
```diff
@@ -88,6 +89,7 @@ def oot_register(registerPanels):
     bpy.types.Scene.ootBlenderScale = bpy.props.FloatProperty(name="Blender To OOT Scale", default=10)
     bpy.types.Scene.ootActorBlenderScale = bpy.props.FloatProperty(name="Blender To OOT Actor Scale", default=1000)
     bpy.types.Scene.ootDecompPath = bpy.props.StringProperty(name="Decomp Folder", subtype="FILE_PATH")
+    bpy.types.Scene.ootHackerFeaturesEnabled = bpy.props.BoolProperty(name="Enable HackerOOT Features")
```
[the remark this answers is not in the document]
kurethedead: I added hackerFeaturesEnabled to the OOT_Properties, but since ootBootupSceneOptions are settings specifically for the scene export operator, I thought it would be better to leave it as is.

fast64_internal/oot/oot_level_writer.py
kurethedead: That's true about the bl_label, the main reason I did that was because I thought there wasn't any other way to get a confirmation dialog with a message. But I found out I could use invoke_props_dialog() which uses the draw function instead of invoke_confirm(). I changed the text as well.

fast64_internal/oot/__init__.py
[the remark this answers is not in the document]
Dragorn421: The purpose of `scene.fast64.oot` is avoiding bloating bpy.types.Scene, I think it'd be fine if it contains export stuff, it's just a different location for data stored on the same data-block in the end (the scene)

But no big deal, it's okay

Dragorn421 (approved):

fast64_internal/oot/__init__.py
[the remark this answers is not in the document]
kurethedead: Oh, I see what you mean, its supposed to be for all OOT related properties. I thought it was intended for OOT globally relevant settings. In that case I'll add the scene bootup options in.

kurethedead:
I've moved the scene bootup options property into OOT_Properties.

fast64_internal/oot/c_writer/oot_scene_bootup.py
```diff
@@ -0,0 +1,137 @@
...
+
+ootEnumBootMode = [("Play", "Play", "Play"), ("File Select", "File Select", "File Select")]
+
+
+class OOTBootupSceneOptions(bpy.types.PropertyGroup):
+    bootToScene: bpy.props.BoolProperty(default=False, name="Boot To Scene")
+    overrideHeader: bpy.props.BoolProperty(default=False, name="Override Header")
+    headerOption: bpy.props.EnumProperty(items=ootEnumHeaderMenuComplete, name="Header", default="Child Day")
+    spawnIndex: bpy.props.IntProperty(name="Spawn", min=0)
+    newGameOnly: bpy.props.BoolProperty(default=False, name="Override Scene On New Game Only")
```
Dragorn421, line 122: maybe a description could help clear this up
`description="Only use this starting scene after loading a new save file"`

fast64_internal/oot/c_writer/oot_scene_bootup.py
```diff
@@ -0,0 +1,137 @@
...
+        if options.headerOption == "Adult Day" or options.headerOption == "Adult Night"
+        else "LINK_AGE_CHILD"
+    )
+
+    return timeOfDay, linkAge
+
+
+# converts ascii text to format for save file name.
+# see src/code/z_message_PAL.c:Message_Decode()
+def stringToSaveNameBytes(name: str) -> bytearray:
```
Zelllll, line 74: One little thing to note here is that soon I intend to change it to just use ascii in hackeroot, but this is good for now.

Zelllll (approved):
Looks good

kurethedead:
I've made everything up to date with most recent version of HackerOOT.

Yanis42:
Bug report for the latest changes:
- issue: using ``undo boot to scene`` doesn't replace the filename, instead it's prepending the bytes
- small detail: the spawn index shouldn't be seen with boot options "file select" and "map select" as it won't affect anything since it's not the Play state

Everything else seems to work properly 👍

Yanis42:
Also small precision, for the player name bug it's not only related to undoing the changes, exporting the scene too cause that

kurethedead:
I fixed the filename issue. I've hidden the spawn options for Map Select, but for File Select I think they should still be present. My intention was that the file select would be for choosing a specific health/equipment load out to test. Alternatively, combined with the `newGameOnly` option you could set the starting scene of the game with the file select intact.

Yanis42:
Everything seems ok now!

Also I love your idea with the file select

PR #124, opened 2022-07-28
Fix incorrect CI textures on import for OOT

kurethedead:
Fixed bug where CI textures would be corrupted when importing OOT display lists.

### DIFF
--- a/fast64_internal/f3d/f3d_parser.py
+++ b/fast64_internal/f3d/f3d_parser.py
@@ -475,8 +475,8 @@ class F3DContext:
 
         self.tlutAppliedTextures = []  # c name
         self.currentTextureName = None
-        self.imagesLoadedFromFile = [] # image
-        self.CIImageFilesStoredAsFullColor = True # determines whether to apply tlut to file or import as is
+        self.imagesDontApplyTlut = set() # image
+        self.ciImageFilesStoredAsFullColor = True # determines whether to apply tlut to file or import as is
 
         # This macro has all the tile setting properties, so we reuse it
         self.tileSettings = [
@@ -715,7 +715,7 @@ class F3DContext:
             and texProp.tex_set
             and texProp.tex_format[:2] == "CI"
             and (texProp.tex not in self.tlutAppliedTextures or texProp.use_tex_reference)
-            and (texProp.tex not in self.imagesLoadedFromFile or not self.CIImageFilesStoredAsFullColor) # oot currently stores CI textures in full color pngs
+            and (texProp.tex not in self.imagesDontApplyTlut or not self.ciImageFilesStoredAsFullColor) # oot currently stores CI textures in full color pngs
         ):
 
             # Only handles TLUT at 256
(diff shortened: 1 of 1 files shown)


Zelllll (approved):
Awesome job fixing this!

fast64_internal/f3d/f3d_parser.py
```diff
@@ -475,6 +475,8 @@ def __init__(self, f3d, basePath, materialContext):
 
         self.tlutAppliedTextures = []  # c name
         self.currentTextureName = None
+        self.imagesLoadedFromFile = [] # image
+        self.CIImageFilesStoredAsFullColor = True # determines whether to apply tlut to file or import as is
```
Dragorn421, line 479: what is `CIImageFilesStoredAsFullColor` for exactly? it's never written to, do you think this should be an option eventually or?

if it should be kept can you make "ci" lowercase so it looks more like a regular field

fast64_internal/f3d/f3d_parser.py
```diff
@@ -1361,7 +1364,9 @@ def loadTexture(self, data, name, region, tileSettings, isLUT):
 
         # TODO: Textures are sometimes loaded in with different dimensions than for rendering.
         # This means width is incorrect?
-        image = parseTextureData(data, textureName, self, tileSettings.fmt, siz, width, self.basePath, isLUT, self.f3d)
+        image, loadedFromImageFile = parseTextureData(data, textureName, self, tileSettings.fmt, siz, width, self.basePath, isLUT, self.f3d)
+        if loadedFromImageFile and image not in self.imagesLoadedFromFile:
+            self.imagesLoadedFromFile.append(image)
```
Dragorn421, line 1369: I suggest making the collection of images a set (in constructor `self.imagesLoadedFromFile = set()`)
and renaming that collection to what it's used for, like `imagesDontApplyTlut` maybe?
```suggestion
        if loadedFromImageFile:
            self.imagesDontApplyTlut.add(image)
```
or reversing things may be simpler:
```py
        if not loadedFromImageFile:
            self.imagesApplyTlut.add(image)
```
(and `and (texProp.tex in self.imagesApplyTlut`)

Dragorn421 (commented):
Nice!

<details>

<summary>
fast64/V5
</summary>

image

</details>

<details>

<summary>
kurethedead/fix_oot_import_ci
</summary>

image

</details>

Can you format your changes with black? (just f3d_parser.py) Or don't if you don't want to and some other PR will do it eventually

fast64_internal/f3d/f3d_parser.py
kurethedead: In the future, if support was added for another decomp project that stored CI textures in greyscale, then this could be turned into an option. I would have pushed this dependency up further but I didn't want to modify a bunch of files that might already be in use from the other PRs in the backlog.

Dragorn421: oh I see. It may be worth expanding the explanation on this then, maybe
```py

        # Determines if images in CI formats loaded from png files,
        # should have the TLUT set by the dlist applied on top of them (False),
        # or if the image file should just be loaded as is with no further change (True)
        # SM64 does (**idk what sm64 does, pls expand**)
        # OoT64 stores CI images as pngs in actual colors (with the TLUT accounted for)
        # For now this can be always True, which covers the (sm64?) and OoT64 use cases
        # In the future this could be an option if for example pngs for CI images were grayscale to represent the palette index.
        self.CIImageFilesStoredAsFullColor = True
```

Dragorn421 (approved):

fast64_internal/f3d/f3d_parser.py
Zelllll: To answer this, SM64 apparently has the luxury of never using CI textures at all.

PR #125, opened 2022-08-01
Fix OoT anim import (reading s16s)

Dragorn421:
Fixes reading an array of s16, currently fast64 expects all elements to be exactly `0xAABB` (hexadecimal, exactly 4 nibbles, no plus/minus sign)

Highlighted by caverill while figuring out something else

Draft because I feel like there's a `s16()`-like function somewhere already but idk where


kurethedead:
This will be fixed in #126 .

PR #129, opened 2022-08-10
OoT scene lighting

thecozies:


kurethedead:
I think the Add Scene Operator for OOT should also set the render settings scene object.

```suggestion
bpy.context.scene.fast64.renderSettings.ootSceneObject = sceneObj
```

Line 105 in `fast64_internal/oot/oot_operators.py`

thecozies:
CDi-Fails was having an issue with the colors being a little too dark, they might need to be gamma corrected. still checking in on that

kurethedead (changes requested):
Everything looks good to me. I have that one suggestion but I don't think GitHub lets me suggest changes to a file that hasn't been modified in a PR.

sauraen:
I've added the line suggested by kurethedead and updated to the current state of Fast-64/V5.

PR #131, opened 2022-09-01
[OoT] Fixed Room Export

Yanis42:
Room export broke with decomp PR #1283, I was able to compile and load my test scene with this fix

List of changes:
- ``meshHeader`` -> ``shapeHeader``
- ``meshType`` -> ``roomShape``
- Fixed types (like ``MeshHeader01``)
- Use enum names for the type
- Use ``ARRAY_COUNT`` for the number of entries
- ``ootEnumMeshType`` -> ``ootEnumRoomShapeType``
- Changed "0, 1, 2" in ``ootEnumRoomShapeType`` to decomp's enum names
- ``SCENE_CMD_MESH`` -> ``SCENE_CMD_ROOM_SHAPE``
- Using arrays to determine the correct type to export
- Used Black to format the changed files


fig02:
I would suggest `ootEnumShapeType` -> `ootEnumRoomShapeType`

"shape" in this case is just a general term to mean gfx/display data, so shape type on its own doesnt mean much

fast64_internal/oot/oot_scene_room.py
```diff
@@ -8,596 +8,626 @@
...
 
-	disableSkybox : bpy.props.BoolProperty(name = "Disable Skybox")
-	disableSunMoon : bpy.props.BoolProperty(name = "Disable Sun/Moon")
+    echo: bpy.props.StringProperty(name="Echo", default="0x00")
 
-	echo : bpy.props.StringProperty(name = "Echo", default = '0x00')
+    objectList: bpy.props.CollectionProperty(type=OOTObjectProperty)
 
-	objectList : bpy.props.CollectionProperty(type = OOTObjectProperty)
+    roomShape: bpy.props.EnumProperty(items=ootEnumRoomShapeType, default="ROOM_SHAPE_TYPE_NORMAL")
```
kurethedead, line 490: Just noting that changing the property name here does cause the previous setting to be lost, but I'm assuming everybody's using type 0 at this point so it should be fine.

kurethedead (approved):
The changes look good to me.

fast64_internal/oot/oot_scene_room.py
Yanis42: I forgot about that, though it takes 2 seconds anyway to change that, it's not like actors where it's really annoying, you won't have a lot of rooms in your scene anyway

fast64_internal/oot/oot_constants.py
```diff
@@ -1,1553 +1,1565 @@
...
+
+ootRoomShapeStructs = [
+    "RoomShapeNormal",
+    "RoomShapeCullable",
+]
+
+ootRoomShapeEntryStructs = [
+    "RoomShapeDListsEntry",
+    "RoomShapeCullableEntry",
 ]
```
Dragorn421, line 16: I think these as dicts would be easier to understand/maintain than the 0/1 indexing
e.g.
```py
ootRoomShapeStructs = {
    "ROOM_SHAPE_TYPE_NORMAL": "RoomShapeNormal",
    "ROOM_SHAPE_TYPE_CULLABLE": "RoomShapeCullable",
}
```
and
```py
ootRoomShapeStructs[mesh.roomShape]
```

fast64_internal/oot/oot_scene_room.py
Dragorn421: And everyone really must be using type 0 ("normal") anyway at the moment, since type 1 is unsupported and type 2 was just documented

Dragorn421 (approved):
I only tested shape 0 "normal", works just as it did before :+1:

PR #133, opened 2022-09-16
[OoT] Fixed camera position data issues

Yanis42:
cam pos data export was weird

### DIFF
--- a/fast64_internal/oot/oot_collision.py
+++ b/fast64_internal/oot/oot_collision.py
@@ -347,7 +347,7 @@ def ootWaterBoxToC(waterBox):
 def ootCameraDataToC(camData):
 	posC = CData()
 	camC = CData()
-	hasPosData = False
+	hasPosData = exportPosData = False
 	if len(camData.camPosDict) > 0:
 
 		camDataName = "CamData " + camData.camDataName() + "[" + str(len(camData.camPosDict)) + "]"
@@ -362,6 +362,7 @@ def ootCameraDataToC(camData):
 			if hasPosData:
 				posC.source += ootCameraPosToC(camData.camPosDict[i])
 				camPosIndex += 3
+				exportPosData = True
 		posC.source += '};\n\n'
 		camC.source += '};\n\n'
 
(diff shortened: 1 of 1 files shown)


fast64_internal/oot/oot_collision.py
```diff
@@ -347,28 +347,32 @@ def ootWaterBoxToC(waterBox):
 def ootCameraDataToC(camData):
 	posC = CData()
 	camC = CData()
+	hasPosData = exportPosData = False
```
Dragorn421, line 350:
```suggestion
	exportPosData = False
```
I think setting `hasPosData` here is basically useless?

fast64_internal/oot/oot_collision.py
```diff
@@ -382,20 +386,23 @@ def ootCameraPosToC(camPos):
...
 		str(camPos.unknown) + ' },\n'
 
-def ootCameraEntryToC(camPos, camData, camPosIndex):
+def ootCameraEntryToC(camPos, camData, camPosIndex, hasPosData):
 	return " ".join((
 		"{",
 		camPos.camSType + ',',
 		('3' if camPos.hasPositionData else '0') + ',',
-		('&' + camData.camPositionsName() + '[' + str(camPosIndex) + ']'),
+		(('&' + camData.camPositionsName() + '[' + str(camPosIndex) + ']') if hasPosData else "NULL"),
```
Dragorn421, line 394: `camPos.hasPositionData` and `hasPosData` are the same here afaict, probably don't need to add the `hasPosData` argument

Dragorn421 (changes requested):
I didn't test but if I read the code right, fast64 currently references "0-length arrays" for cameras which don't use "pos data" (BgCamData in decomp rn iirc), and this fixes that weird output by replacing those occurrences by `NULL`

Sounds about right?

Yanis42:
yea currently it's exporting an empty array, which obviously prevent compilation, this fix removes the array entirely if its content is empty

Dragorn421 (approved):

PR #135, opened 2022-09-23
Fixed color attribute add operator data_type argument for Blender 3.3

kurethedead:


fast64_internal/f3d/f3d_material.py
```diff
@@ -1935,17 +1937,20 @@ def addColorAttributesToModel(obj: bpy.types.Object):
...
 
+    useFloatColor: bool = bpy.app.version >= (3, 3, 0)
+    attrDataType: str = "FLOAT_COLOR" if useFloatColor else "COLOR"
+
     conv_col, has_col = shouldConvOrCreateColorAttribute(mesh, attr_name="Col")
     if conv_col:
         convertColorAttribute(mesh, attr_name="Col")
     elif not has_col:
-        bpy.ops.geometry.color_attribute_add(name="Col", domain="CORNER", data_type="COLOR")
+        bpy.ops.geometry.color_attribute_add(name="Col", domain="CORNER", data_type=attrDataType)
```
Dragorn421, line 1947: oh now I understand why this broke, it's the operator that changed I guess (that's weird but I guess?), not the api

afaict the api has always used `"FLOAT_COLOR"`
https://docs.blender.org/api/3.2/bpy.types.AttributeGroup.html#bpy.types.AttributeGroup.new
https://docs.blender.org/api/3.2/bpy.types.Attribute.html#bpy.types.Attribute.data_type

This should work in all versions (that have color atributes i.e. 3.2+), better test it, but I don't have Blender 3.2:
```py
mesh.color_attributes.new("Col", "FLOAT_COLOR", "CORNER")
```

thecozies: yup just the operator was updated without any warning!

3.2
https://docs.blender.org/api/3.2/bpy.ops.geometry.html#bpy.ops.geometry.color_attribute_add

3.3
https://docs.blender.org/api/3.3/bpy.ops.geometry.html#bpy.ops.geometry.color_attribute_add

pretty dumb. I'm sure the operator won't change but I'm fine with switching to this instead

thecozies (approved):
lgtm. I like dragorns suggestion if you want to give it a shot!

kurethedead:
I've updated the code to use the API, I've tested on 3.2 and 3.3 and the error doesn't appear.

thecozies (approved):
I approve even _MOOOORE_

Dragorn421 (approved):

PR #139, opened 2022-09-26
Format __init__.py and oot_constants.py

Yanis42:


Dragorn421 (approved):
(note this going into fast64/main is intended, this is already formatted in fast64/V5)

PR #148, opened 2022-09-30
[OoT] C-Writer cleanup & improvements

Yanis42:
I decided to improve the code of Fast64 (for now starting with OoT's C writer), I thought it might be good to avoid PR'ing everything so here's the first cleanup PR.

The file I did contained like ~840 lines, now reduced to ~70, everything is split in their own files for code readability, also to improve that I added hints/annotations (idk how you call those) on function's arguments (including some inside ``OOTRoom``, ``OOTScene`` and basically every place I needed those to understand what's going on)

I also improved the output a bit, for instance this is the new output for a single ``LightSettings`` entry:
```c
{
    // Dawn Lighting
    { 70, 45, 57 },       // Ambient Color
    { 73, 73, 73 },       // Diffuse0 Direction
    { 180, 154, 138 },    // Diffuse0 Color
    { -73, -73, -73 },    // Diffuse1 Direction
    { 20, 20, 60 },       // Diffuse1 Color
    { 140, 120, 100 },    // Fog Color
    ((1 << 10) | 993),    // Blend Rate & Fog Near
    12800,                // Fog Far
},
```
I also improve the imports in some places, I hope this won't break anything (it wouldn't surprise me lol)

----
For now I'm focused on ``fast64_internal/oot/c_writer``, next I'll do the OoT folder and then I'll see, since I don't know a lot about SM64 so I don't really want to mess with this, I need to spend some time to learn the basics before editing anything.

Small note, in some places I added some "TODO" comments, this is temporary and can go if you don't want them

(Also note that this PR change absolutely nothing for the end-user, except some minor bugfix like the dummy room checkbox displaying on every tab)


fast64_internal/oot/c_writer/oot_room_writer/oot_room_shape_to_c.py
```diff
@@ -0,0 +1,96 @@
...
+        roomShapeEntry += (
+            "{ "
+            + ", ".join([f"{pos}" for pos in meshEntry.cullGroup.position])
+            + "}, "
+            + f"{meshEntry.cullGroup.cullDepth}, "
+        )
+    elif roomShape == "ROOM_SHAPE_TYPE_IMAGE":
+        raise PluginError("Pre-Rendered rooms not supported.")
+
+    roomShapeEntry += f"{opaqueName}, {transparentName}" + " }\n"
```
kurethedead, line 23:
```suggestion
    roomShapeEntry += f"{opaqueName}, {transparentName}" + " }"
```

fast64_internal/oot/c_writer/oot_scene_writer/oot_path_to_c.py
```diff
@@ -0,0 +1,41 @@
...
+    pointName = f"Vec3s {path.pathName()}[]"
+    pointsData = " },\n".join(
+        [indent + "{ " + ", ".join([f"{round(point[i])}" for i in range(len(point) - 1)]) for point in path.points]
+    )
+
+    # .h
+    pointData.header = f"extern {pointName};\n"
+
+    # .c
+    pointData.source = f"Vec3s {pointName}" + " = {\n" + pointsData + " },\n};\n\n"
```
kurethedead, line 19:
```suggestion
    pointData.source = f"{pointName}" + " = {\n" + pointsData + " },\n};\n\n"
```

fast64_internal/oot/c_writer/oot_scene_writer/oot_entrance_exit_to_c.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # .h
+    exitListData.header = f"extern {exitListName};\n"
+
+    # .c
+    exitListData.source = (
+        (exitListName + " = {\n")
+        # @TODO: use the enum name instead of the raw index
+        + "".join([indent + f"{exitEntry.index}" for exitEntry in scene.exitList])
+        + "};\n\n"
```
kurethedead, line 42:
```suggestion
        + ", ".join([indent + f"{exitEntry.index}" for exitEntry in scene.exitList])
        + "\n};\n\n"
```

fast64_internal/oot/c_writer/oot_room_writer/oot_room_shape_to_c.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # .h
+    roomShapeEntryData.header = f"extern {roomShapeEntryName};\n"
+
+    # .c
+    roomShapeEntryData.source = (
+        roomShapeEntryName
+        + " = {\n"
+        + " },\n".join([indent + ootGetRoomShapeEntry(entry, mesh.roomShape) for entry in mesh.meshEntries])
+        + "};\n\n"
```
kurethedead, line 44:
```suggestion
        + ",\n".join([indent + ootGetRoomShapeEntry(entry, mesh.roomShape) for entry in mesh.meshEntries])
        + "\n};\n\n"
```

fast64_internal/oot/c_writer/oot_scene_writer/oot_scene_layer_to_c.py
```diff
@@ -0,0 +1,87 @@
...
+    if len(scene.transitionActorList) > 0:
+        layerData.append(ootTransitionActorListToC(scene, headerIndex))
+
+    # Write the room segment list
+    if headerIndex == 0:
+        layerData.append(ootRoomListHeaderToC(scene))
+
+    # Write the path list
+    if len(scene.pathList) > 0:
+        layerData.append(ootPathListToC(scene))
```
kurethedead, line 32:
```suggestion
        # Write the path list
        if len(scene.pathList) > 0:
            layerData.append(ootPathListToC(scene))
```
Technically paths can be header specific, but for now they should only be written once.

fast64_internal/oot/c_writer/oot_scene_writer/oot_trans_actor_to_c.py
```diff
@@ -0,0 +1,39 @@
...
+from ...oot_scene_room import OOTScene, OOTTransitionActor
+from ...oot_utility import indent
+
+
+def ootGetTransActorData(transActor: OOTTransitionActor):
+    """Returns the transition actors array's data"""
+    sides = [(transActor.frontRoom, transActor.frontCam), (transActor.backRoom, transActor.backCam)]
+    return (
+        "{ "
+        + (", ".join(["{ " + f"{room}, {cam}" + " }" for room, cam in sides]) + ", ")
```
kurethedead, line 11:
```suggestion
        + "{"
        + (", ".join(["{ " + f"{room}, {cam}" + " }" for room, cam in sides]) + ", ")
        + "}, "
```

kurethedead (changes requested):
Seems good, I skimmed over it a bit and tried exporting a level. There's a couple compilation errors due to formatting that can be easily fixed. Some minor changes not related to your code that are needed to fix compilation errors: 
1. In `oot_collision.py`, line 488, remove the `"&" + `
2. In `oot_level_classes.py`, line 534, fix the naming `addAtStartPosIndex` -> `addStartPosAtIndex`

Yanis42:
Finally updated this PR, first of all I apologies about the diff, it got bigger with the latest push, tell me if I should split things 👍

I updated the way c_writer files were organised, now it's called "exporter" and everything related to creating C code and export to decomp in general will be hosted there (this includes functions from ``oot_level_writer.py``)

I was able to compile my test scene (using different layers, paths and a cutscene), so I think I can consider build errors from Kure's review as solved (I'll take a look at the other comments though)

Anyway, now it should be ready to review, I should've make this PR a draft, that's my bad (but mistakes are how we learn 👍)

Note: you can see the "final" work [there](https://github.com/Yanis42/fast64/tree/oot_cleanup), there's minor changes I need to merge in here tho (the paths layer indices)

fast64_internal/oot/oot_level_writer.py
```diff
@@ -167,6 +167,7 @@ def readSceneData(scene, scene_properties, sceneHeader, alternateSceneHeaders):
     scene.skyboxID = getCustomProperty(sceneHeader, "skyboxID")
     scene.skyboxCloudiness = getCustomProperty(sceneHeader, "skyboxCloudiness")
     scene.skyboxLighting = getCustomProperty(sceneHeader, "skyboxLighting")
+    scene.lightMode = sceneHeader.skyboxLighting
```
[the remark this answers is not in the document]
Yanis42: because ``sceneHeader`` is from the scene empty, and I need that value later when exporting light settings

fast64_internal/oot/c_writer/oot_room_writer/oot_room_shape_to_c.py
Yanis42: since this is an array I think it should be ``" },\n"``, though I don't think it's ever exporting multiple values here (also that ``\n`` is to get the closing brackets of the array on the next line)

Yanis42: nvm I missed the other comment lol

fast64_internal/oot/c_writer/oot_scene_writer/oot_scene_layer_to_c.py
Yanis42: welp, now it's using the layer index for the array names so this shouldn't matter I guess? but since there's no layer choice in the UI I'll add this change 👍

fast64_internal/oot/c_writer/oot_scene_writer/oot_trans_actor_to_c.py
Yanis42: that part looks like this now: ``("{ { " + (", ".join([f"{room}, {cam}" for room, cam in sides]) + " }, "))`` (basically it's one line per argument), let me know if you still want this change 👍 (this file is called ``exporter/scene/to_c/transition_actor.py`` now)

fast64_internal/oot/oot_level_writer.py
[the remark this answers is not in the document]
kurethedead: But right now `scene.lightMode` is the exact same thing as `scene.skyboxLighting`, except it doesn't handle the custom enum value.

fast64_internal/oot/exporter/export.py
```diff
@@ -0,0 +1,219 @@
...
+    isHWv1: bool,  # is hardware v1
+    sceneName: str,
+    dlFormat: DLFormat,
+    savePNG: bool,
+    exportInfo: ExportInfo,
+    bootToSceneOptions: OOTBootupSceneOptions,
+):
+    # circular import fix
+    from .scene_folder import modifySceneFiles
+    from .spec import modifySegmentDefinition
```
kurethedead, line 117: These two imports can be moved to the top of the file without any issues.

fast64_internal/oot/exporter/export.py
```diff
@@ -0,0 +1,219 @@
...
+    from ..oot_level_writer import ootConvertScene
+
+    checkObjectReference(inSceneObj, "Scene object")
+    isCustomExport = exportInfo.isCustomExportPath
+    exportPath = exportInfo.exportPath
+
+    outScene: OOTScene = ootConvertScene(
+        inSceneObj, transformMatrix, f3dType, isHWv1, sceneName, dlFormat, not savePNG
+    )
```
kurethedead, line 126: I'm not a fan of this circular import fix. What we could do instead is move this block of code into `oot_level_writer.py` right before calling `exportScene`, then have `exportScene` take in an `OOTScene` argument. This way `export.py` doesn't have any dependency on `oot_level_writer.py`. This would also remove a lot of the parameters for `exportScene`.

kurethedead (changes requested):
1. Although its a compile warning and not an error, I still think you should apply the `oot_collision` fix that I proposed in my earlier comment. 
2. I've made some suggestions on the circular import issue in the exporter. 
3. I think that the `exporter` folder should also be placed in a `scene` folder, so that an `importer` folder can be added in that later. 
4. Aren't the extra `to_c` folders unnecessary? What other sibling folders would you add?

Yanis42:
right now it looks useless I agree, but later there will be the empty conversion code in here (I mean in the parent folder of ``to_c`` folders), here's how it looks in the branch where I have everything done:
<details closed>
<summary>Files</summary>

</details>

in the panel folder there's all the code for the UI stuff, split by location (viewport/properties) and the importer folder is on the same model as the exporter one

your idea would be doing it in reverse? like having scene, room, actor etc then inside those folders one to export, one to import, one to draw on the UI

Yanis42:
I think this needs more discussion so we can establish an organisation of the files that would convince people, I think what I did can be a nice base to decide how to split things and eventually enforcing a line number limit to avoid having 1000+ lines in a single file (which isn't great when you want to update something or just understand what's going on)

Also I can see something similar being applied to SM64, though idk how the code looks like in here

fast64_internal/oot/exporter/export.py
```diff
@@ -103,28 +105,16 @@ def ootPreprendSceneIncludes(outScene: OOTScene, fileData: CData):
...
     savePNG: bool,
     exportInfo: ExportInfo,
     bootToSceneOptions: OOTBootupSceneOptions,
 ):
-    # circular import fix
-    from .scene_folder import modifySceneFiles
-    from .spec import modifySegmentDefinition
-    from ..oot_level_writer import ootConvertScene
-
     checkObjectReference(inSceneObj, "Scene object")
```
kurethedead, line 114: `checkObjectReference` also should be moved out of this function, as `exportScene` shouldn't need to take in any blender objects as arguments.

kurethedead (changes requested):
The reason I'm thinking of categorizing by scene, room, skeleton, etc. is that it would easier to handle shared properties/classes common to importing/exporting for a single thing. I do think more feedback would be helpful.

Yanis42:
closing this for several reasons:
- conflicts are awful with this branch
- it was easier to start again from main
- I changed a bit how I will cleanup OoT, first I'm applying the new organisation, next I'm actually doing some code cleanup work and improvements

I'm opening a new PR with the new organisation for c_writer files, improvements added by this PR will be the subject of another PR whenever I'm done with re-organising the code, the reason for this is conflicts, it's easier to manage them if I work this way

PR #149, opened 2022-10-01
[OoT] Remove hardcoded draw config names list

Yanis42:
Currently it'd break if those names are changed, now it'll break if the enum's name (``SceneDrawConfig``) or the filename (z64scene.h) is changed

Also I think we should start to parse things from decomp when we can instead of hardcoding everything to avoid those kind of issues in the future


Yanis42:
(also small bonus, if the user adds a name in the decomp enum Fast64 will automatically use it)

Zelllll:
I think this is cool but imo it introduces multiple new issues. For example, I usually never specify a decomp path because my decomp repo isn’t structured the same as zeldaret anymore, so relying on a decomp path being specified may be a problem for some users. I will add though that I also manually export my scenes so this feature specifically technically doesn’t affect me, but implementing this on a broader level potentially could be an issue for some. Just something worth considering.

Yanis42:
oh yea I haven't thought about that, then I guess I'll add back the hardcoded list and use that only if it can't find the file in decomp, this should fix most issues (also it won't break Fast64 for everyone)

Alternatively I can add a textbox to set the path to the draw config enum file path, though I'm not really sure it'd be a good solution

Zelllll:
Yeah perhaps that is one viable option, I’m not saying that we can’t do this feature or anything, just that it might need a little more discussion.

sauraen:
How about:
- draw config is stored in the Blender just as an integer index
- if decomp path is set up, z64scene.h exists, has the right enum, etc., then it will show the name of the draw config
- otherwise just let you set the integer directly

Even in modding contexts where people don't massively change the decomp structure, people will often add or modify draw configs, so being able to just pick the index is good.

Yanis42:
oh yea I guess exporting the int is not that bad, I haven't thought about that lol

also I think most people will use decomp anyway so I think it's a good compromise 👍

Yanis42:
closing this for now, draw configs changed a bit, now it's an enum and the list is used for the scene import, I'll get back to this once I'm done with everything else as it's not really a priority

PR #151, opened 2022-10-02
Ship of Harkinian Compatibility

MelonSpeedruns:
These are in-progress changes to Fast64 that adds support to the Ship of Harkinian. (Ocarina of Time PC Port)

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -3347,6 +3347,8 @@ class SPMatrix:
             header += "segmented_to_virtual(" + str(self.matrix) + ")"
         else:
             header += str(self.matrix)
+            if bpy.context.scene.fast64.oot.shipOfHarkinianCompatible:
+                header += ' | 1'
         return header + ", " + str(self.param) + ")"
 
     def to_sm64_decomp_s(self):
@@ -3452,7 +3454,10 @@ class SPDisplayList:
 
     def to_c(self, static=True):
         if static:
-            return "gsSPDisplayList(" + self.displayList.name + ")"
+            if bpy.context.scene.fast64.oot.shipOfHarkinianCompatible and self.displayList.name.startswith('0x'):
+                return "gsSPDisplayList(" + self.displayList.name + ' | 1)'
+            else:
+                return "gsSPDisplayList(" + self.displayList.name + ")"
         elif self.displayList.DLFormat == DLFormat.Static:
             header = "gSPDisplayList(glistp++, "
             if bpy.context.scene.decomp_compatible:
@@ -4944,7 +4949,10 @@ class DPSetTextureImage:
         if not static and bpy.context.scene.decomp_compatible:
             header += "segmented_to_virtual(" + self.image.name + "))"
         else:
-            header += self.image.name + ")"
+            if bpy.context.scene.fast64.oot.shipOfHarkinianCompatible and self.image.name.startswith('0x'):
+                header += self.image.name + " | 1)"
+            else:
+                header += self.image.name + ")"
         return header
 
     def to_sm64_decomp_s(self):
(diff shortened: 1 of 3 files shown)


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3347,6 +3347,8 @@ def to_c(self, static=True):
             header += "segmented_to_virtual(" + str(self.matrix) + ")"
         else:
             header += str(self.matrix)
+            if bpy.context.scene.fast64.oot.shipOfHarkinianCompatible:
```
kurethedead, line 3350: I wouldn't add game specific code to f3d_gbi or f3d_writer if possible. For modifying material commands, you should look at `oot_model_classes.py -> OOTModel -> onMaterialCommandsBuilt`. `gfxList.commands` is a list of commands that you can iterate through, you can check for `isinstance(SPDisplayList)` and then modify that.

fast64_internal/f3d/f3d_writer.py
```diff
@@ -2336,6 +2336,19 @@ def saveOrGetTextureDefinition(fMaterial, fModel, image: bpy.types.Image, imageN
     print("Finished converting.")
     fModel.addTexture((image, (texFormat, "NONE")), fImage, fMaterial)
 
+    if bpy.context.scene.fast64.oot.shipOfHarkinianCompatible:
```
kurethedead, line 2339: For texture endian issues, instead of doing this, check  `oot_model_classes.py -> OOTGfxFormatter -> __init__`, you can conditionally change the `texArrayBitSize` argument to `8` which should save the texture array as a `u8` instead of a `u64`.

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3452,7 +3454,10 @@ def to_binary(self, f3d, segments):
 
     def to_c(self, static=True):
         if static:
-            return "gsSPDisplayList(" + self.displayList.name + ")"
+            if bpy.context.scene.fast64.oot.shipOfHarkinianCompatible and self.displayList.name.startswith('0x'):
+                return "gsSPDisplayList(" + self.displayList.name + ' | 1)'
```
kurethedead, line 3458: I really don't understand how this is supposed to change endianness, but if you're looking for segmented addresses, `gsSPMatrix` should also be taken into account.

kurethedead (commented):
I've made a couple suggestions/pointers for how to approach this.

fast64_internal/oot/__init__.py
```diff
@@ -32,13 +32,14 @@ def draw(self, context):
 
         prop_split(col, context.scene, "ootDecompPath", "Decomp Path")
         col.prop(context.scene.fast64.oot, "hackerFeaturesEnabled")
-
+        col.prop(context.scene.fast64.oot, "shipOfHarkinianCompatible")
```
Zelllll, line 35: I'm confused why you are calling it this considering that as far as I can tell this feature does not merely allow "compatibility" with the PC port, but rather it changes the output such that it is no longer compatible with N64. As an aside, I am a little confused why such a PR is really necessary when the PC port tools should already be able to handle converting N64 models to whatever format it is expecting. Can you explain why it is necessary that Fast64 be modified to support the PC port?

Zelllll, line 35: I'm confused why you are calling it this considering that as far as I can tell this feature does not merely allow "compatibility" with the PC port, but rather it changes the output such that it is no longer compatible with N64. Unless there is something I am missing, this variable should be changed to be more reflective of what it does (i.e. describing this as PC port *mode*, rather than "compatibility."

Zelllll (changes requested):
I have a few suggestions, but also have a few concerns about the overall idea of this PR that I hope you can make clearer to me.

Mainly I am wondering why such a PR is really necessary when the PC port tools should already be able to handle converting N64 models to whatever format it is expecting. Can you explain why it is necessary that Fast64 be modified to support the PC port so that I can better understand the viewpoint that this PR is coming from?

For example, would it not be possible for the PC port to simply create a macro that handles all of these changes automatically with the preprocessor, rather than requiring extra code to be added to the exporter itself?

PR #152, opened 2022-10-02
[OoT] Applied new organisation to Panels

Yanis42:
Another cleanup PR, this one is focused on the OoT panel (where you have the scene exporter, the tools etc), I had to improvise temp fixes because of circular import issues, hopefully they'll be fixed when the cleanup will be complete

Changes:
- applied [new organisation](https://pastebin.com/HAdmtDaq) on every panels (viewport + properties)
- removed useless splits
- cleaned up oot_level.py registers, now the classes are registering where it's declared
- rename oot_level.py to props_panel_main.py as it's the main code for the properties window panels (I'm not 100% satisfied with this name but I couldn't find anything better, if you have any suggestions feel free)

More cleanup soon™


sauraen (commented):
Please go over this again in light of #150 being now merged. But yes, this cleanup needed to happen, the circular import issues were causing problems for me as well.

Yanis42:
I'll update this PR (and the level_c cleanup one) when I'll be done with cleaning up the whole OoT folder, right now I just have oot_utility left (https://github.com/Yanis42/fast64/tree/oot_cleanup), then I'll see how I can improve things and then I'll update everything

Yanis42:
finally pushing some changes to this branch:
- converted this PR to a draft
- merged main, fixed conflicts
- applied new organisation (discussed on discord)
- removed useless splits
- started to split properties panels (I was thinking it was better to include every OoT panels)

more changes soon™

Yanis42:
new changes:
- applied the new organisation on more files
- cleaned up ``oot_level.py`` registers, now the class are registering where the classes are declared
- rename ``oot_level.py`` to ``props_panel_main.py`` as it's the main code for the properties window panels (I'm not 100% satisfied with this name but I couldn't find anything better, if you have any suggestions feel free)

almost done with this! I just need to figure out how to handle ``oot_model_classes.py`` and it should be ready for review

Yanis42:
This PR should be ready to review, I found no issues except the actor visibility one but it's there in main (reported on discord)

kurethedead:
Some of this I've already commented before, but my thoughts:

1. I don't think the `panel` subfolders are necessary.
2. The contents of `/other/panel/viewport` can just be in the top level `/oot` folder.
3. I'm not sure the `display_list` subfolder in `f3d/panel/viewport/display_list` is necessary, unless you were thinking of adding something else there? Also, it seems like `classes` here should just be part of the `properties` file.
4. `props_panel_main` -> `object_panel` maybe?
5. I'm a bit confused about the distinction between `viewport` and `properties` - I thought `viewport` was for UI, but it seems like in a lot of cases there are PropertyGroups and Operators in the `viewport` files/folders.

Also, for future reference we should use `__init__.py` more in order to prevent having long import paths for specific modules.

Yanis42:
1. how about removing that subfolder and instead rename ``viewport`` and ``properties`` folders to ``viewport_panels`` and ``properties_panels``?
2. same thing here
3. I'll merge the files 👍
4. 👍 probably better with this name
5. Panels are split by location on the UI (apparently the real term would be "editor type"), "viewport" is every panels that you can see on the viewport tab on Blender, same for properties, I think it's better to have them organised like this, hence my suggestion in 1., though I understand why it's confusing
image
image

Dragorn421:
fwiw the "viewport" editor type is `Area.type` in the api (areas "contain" editors)
https://docs.blender.org/api/current/bpy.types.Area.html#bpy.types.Area.type

And the way the 3d viewport is typically referenced is more like "view 3d"/`view3d`
https://docs.blender.org/api/current/bpy_types_enum_items/space_type_items.html#rna-enum-space-type-items
See also https://docs.blender.org/api/current/search.html?q=view3d

Yanis42:
I do think this distinction (for the location) is needed, though if there's better names of course feel free 😄 

So, View3D seems like the official name for what I'm calling "viewport", properties seems to be the proper name, but it can be confusing (but I think if it's under a "panel" folder or anything having "panel" in the name it clears the confusion)

kurethedead:
I think the naming is a bit confusing. I understand it now that you explained it, but when I see `properties` I assume that means `PropertyGroup`, which makes me think `viewport` means `Panel`. The `_panel` prefix is a bit odd too, since that makes it seem like its purely a UI folder when its not. Now that I'm thinking about it, a lot of the `viewport` code seems like it would be better moved to `exporter`. I think we should move the Operators and PropertyGroups in `viewport` to their respective `exporter` and `importer` folders, and then convert `viewport` into `view3d_panel`, which has the Panel. You can refactor the panel's draw() function and move its contents into the respective `importer`/`exporter` folders as separate draw functions, but I can handle that later if its too much work.

Yanis42:
note that for now there's only classes in there but later I plan on moving the drawing code too, in the end it's just gonna be a simple split between classes and draw functions

also I think it's a bit confusing to have UI stuff in an exporter folder

Yanis42:
I think it's confusing to put the actual panel code and its operators/propertygroups in another file, to me it's a part of the panel so it should stay with the panel's draw code, also it's only used for the panel (and doing this can avoid import issues in the future)

kurethedead:
In a similar way, it would be confusing to separate the operator from the code its executing. I also did think that the panel draw code should be split into importer/exporter folders, so they wouldn't be separated from their operators. The only reason the view3d_panel would exist at all anymore is that the import and export operators exist in the same panel, kind of like how the object properties panel contains UI for every empty type.

Yanis42:
fixed conflicts, waiting on #187 and #188 because those PRs will bring more conflicts 🎉 

also I think we need to discuss more about the names and the organisation for the panels, I just want this PR to be merged so I can move on other files 👍

Dragorn421:
I think kure mentioned it too, it's weird to me that so much falls under "panel", namely operators (code) and properties (data)

Yanis42:
where should I move them?

Dragorn421:
I don't have a clear idea

<details>

<summary>tree</summary>

```
oot
├── __init__.py
├── oot_actor.py
├── oot_anim.py
├── ...
├── oot_utility.py
├── props_panel_main.py
├── README.md
├── actor
│   └── panel
│       └── properties.py
├── animation
│   └── panel
│       ├── properties.py
│       └── viewport.py
├── collision
│   └── panel
│       ├── properties.py
│       └── viewport.py
├── cutscene
│   └── panel
│       ├── properties.py
│       └── viewport.py
├── data
│   ├── oot_actor_data.py
│   ├── oot_data.py
│   ├── oot_getters.py
│   ├── oot_object_data.py
│   └── xml
│       ├── ActorList.xml
│       └── ObjectList.xml
├── f3d
│   └── panel
│       ├── properties.py
│       └── viewport
│           └── display_list
│               ├── classes.py
│               ├── __init__.py
│               ├── operators.py
│               └── panel.py
├── other
│   └── panel
│       └── viewport
│           ├── file_settings.py
│           └── tools
│               ├── __init__.py
│               ├── operators.py
│               └── panel.py
├── room
│   └── panel
│       └── properties.py
├── scene
│   ├── exporter
│   │   └── to_c
│   │       ├── actor.py
│   │       ├── __init__.py
│   │       ├── ...
│   │       ├── scene_table_c.py
│   │       └── spec.py
│   └── panel
│       ├── properties
│       │   ├── classes.py
│       │   ├── __init__.py
│       │   ├── operators.py
│       │   └── panel.py
│       └── viewport
│           ├── classes.py
│           ├── __init__.py
│           ├── operators.py
│           └── panel.py
├── skeleton
│   └── panel
│       ├── properties.py
│       └── viewport
│           ├── classes.py
│           ├── __init__.py
│           ├── operators.py
│           └── panel.py
└── spline
    └── panel
        └── properties.py
```

</details>

overall i think the "panel" folders aren't needed and their "properties"/"viewport" contents can live in the parent of "panel"

I'll have to look in more details what distinction is made between properties.operators/panel and viewport.operators/panel

Yanis42:
how about:
- every ``PropertyGroup`` in a ``properties.py`` (or just property idk) file, one for each element (actor, collision, scene, cutscene etc) with comments specifying how each class is used for
- same for operators (``operators.py``?)
- every draw functions in the same place (and again split by theme (so actor collision etc), again with comments telling the purpose

so basically split property group and operator classes in their own file, same for draw functions, organised by scene element (so actor collision etc etc), also if a file becomes too large it'd be a subfolder to make it easier to maintain/read, and eventually the other files will have the same organisation as a lot of the code is basically the same (Operator -> PropertyGroup -> Draw Functions -> Logic Functions)

would that be fine?

sauraen:
What is this blocked waiting for?

Yanis42:
I had an idea the other day (my last comment) and I'm just waiting on feedback (@kurethedead)

kurethedead:
Overall it sounds good to me, but I think `operator.py` should also be split into `importer.py` and `exporter.py`. I'm assuming the code that the operator executes is going to be in those files as well?

Yanis42:
I'm a bit concerned about this Kure, I don't like having the function executed by the operator but not the sub-functions it need to execute, also file size concerns me

though I don't have any idea for this so I might do it anyway, we can always change that later

Yanis42:
I forgot about panel classes, I guess I'll just group them in a ``panels.py`` file to keep the spirit of splitting by class type

Dragorn421:
This now looks like something I can get behind, though I have yet to give it a proper look

Any other favorable impressions?

Yanis42:
note that I'm not 100% done, I'm working on moving props draw code to the PropertyGroup classes as discussed on Discord earlier with kure

except this detail I'm personally satisfied with this organisation and I hope this will be over so we can move on something else lol (aka code improvement for this, next it should be the start of the file-per-file cleanup PauseChamp)

Yanis42:
draw functions for properties should be moved, everything seems to work properly though idk how to test things like animations or skeletons, I'll move the draw sub-functions later, for now I think it's ok

I wanted to say something else but I forgot what it is, whatever lol

Yanis42:
I remembered, I was able to get rid of ``oot_scene_room.py`` and ``oot_actor.py`` as these two files contained draw code for properties and nothing else

Yanis42:
## Since this PR is way larger than I expected, here's a changelog

In general, panels are moved to ``panels.py`` files, operators to ``operators.py`` files and properties to ``properties.py``, draw functions for PropertyGroup elements are moved to each PropertyGroup classes with the name ``draw_props()`` and each file containing classes have register and unregister functions, which are imported by ``__init__.py`` instead of having them inside ``oot_level.py`` (now ``props_panel_main.py``)

Note: the organisation is the one described there: https://github.com/Fast-64/fast64/pull/152#issuecomment-1346204647

For the actual changes:
- ``__init__.py`` -> file settings panel moved to ``file_settings.py``
- ``oot_actor.py`` -> moved PropertyGroup and Operator classes to ``actor/properties.py`` and ``actor/operators.py``
- ``oot_anim.py`` -> moved PropertyGroup, Operator and Panel classes to ``anim/properties.py``, ``anim/operators.py`` and ``anim/panels.py``
- ``oot_collision.py`` -> moved PropertyGroup, Operator and Panel classes to ``collision/properties.py``, ``collision/operators.py`` and ``collision/panels.py``
- ``oot_cutscene.py`` -> moved PropertyGroup, Operator and Panel classes to ``cutscene/properties.py``, ``cutscene/operators.py`` and ``cutscene/panels.py``
- ``oot_f3d_writer.py`` -> moved PropertyGroup, Operator and Panel classes to ``f3d/properties.py``, ``f3d/operators.py`` and ``f3d/panels.py``
- ``oot_level.py`` -> ``props_panel_main.py``, contains most of the original file, the class register functions were split into individual functions inside each oot subfolder (actor, scene etc), and the corresponding classes moved there
- ``oot_level_parser.py`` -> moved the import operator to ``scene/operators.py`` with the (directly) related functions
- ``oot_level_writer.py`` -> moved Operator and Panel classes to ``cutscene/operators.py`` and ``cutscene/panels.py``
- ``oot_model_classes.py`` -> moved class ``OOTDynamicTransformProperty`` to ``skeleton/properties.py``
- ``oot_operators.py`` -> ``tools/operators.py`` and ``tools/panel.py``
- ``oot_scene_room.py`` -> moved PropertyGroup and Operator classes to ``room/properties.py`` and``room/operators.py``
- ``oot_skeleton.py`` -> moved PropertyGroup, Operator and Panel classes to ``skeleton/properties.py``, ``skeleton/operators.py``, ``skeleton/panels.py``
- ``oot_skeleton_import_data.py`` -> moved class ``OOT_SaveRestPose`` to ``skeleton/operators.py``
- ``oot_spline.py`` -> moved PropertyGroup and Panel classes to ``spline/properties.py`` and ``spline/panels.py``
- ``oot_utility.py`` -> moved some functions there because of circular import issues

kurethedead (approved):
The organization is good. I noticed some operator files have the "top level" function also included in them - I'm assuming this is because the actual code isn't handled yet, but in case its not, I want to say that I think these functions should be in the same file as the rest of the code instead of the operator file (assuming they're separate files). I did some really basic testing of every main operator and nothing seems to have broken.

PR #154, opened 2022-10-03
[OoT] Minor room behavior improvements

Yanis42:
I was thinking setting the room mode to "too hot" and using the heat timer might be confusing as it's hidden under Link Idle mode, this PR removes the 'too hot' entry from the idle enum and replaces it with a checkbox instead

Also I updated the room behavior enum names to be more accurate after doing some research with decomp

### DIFF
--- a/fast64_internal/oot/oot_scene_room.py
+++ b/fast64_internal/oot/oot_scene_room.py
@@ -476,7 +476,7 @@ class OOTRoomHeaderProperty(bpy.types.PropertyGroup):
     linkIdleMode: bpy.props.EnumProperty(name="Link Idle Mode", items=ootEnumLinkIdle, default="0x00")
     linkIdleModeCustom: bpy.props.StringProperty(name="Link Idle Mode Custom", default="0x00")
     roomIsHot: bpy.props.BoolProperty(
-        name="Use Room Heat Behavior",
+        name="Use Hot Room Behavior",
         description="Use heat timer/screen effect, overrides Link Idle Mode",
         default=False,
     )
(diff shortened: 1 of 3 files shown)


fast64_internal/oot/oot_scene_room.py
```diff
@@ -539,7 +543,11 @@ def drawRoomHeaderProperty(layout, roomProp, dropdownLabel, headerIndex, objName
         behaviourBox = layout.column()
         behaviourBox.box().label(text="Behaviour")
         drawEnumWithCustom(behaviourBox, roomProp, "roomBehaviour", "Room Behaviour", "")
-        drawEnumWithCustom(behaviourBox, roomProp, "linkIdleMode", "Link Idle Mode", "")
+
+        behaviourBox.prop(roomProp, "roomIsHot")
+        if not roomProp.roomIsHot:
+            drawEnumWithCustom(behaviourBox, roomProp, "linkIdleMode", "Link Idle Mode", "")
+
```
Dragorn421, line 550: can you `else: show text "Link Idle Mode: Hot room"` or something like that

fast64_internal/oot/oot_scene_room.py
```diff
@@ -476,6 +475,11 @@ class OOTRoomHeaderProperty(bpy.types.PropertyGroup):
     showInvisibleActors: bpy.props.BoolProperty(name="Show Invisible Actors")
     linkIdleMode: bpy.props.EnumProperty(name="Link Idle Mode", items=ootEnumLinkIdle, default="0x00")
     linkIdleModeCustom: bpy.props.StringProperty(name="Link Idle Mode Custom", default="0x00")
+    roomIsHot: bpy.props.BoolProperty(
+        name="Use Room Heat Behavior",
```
Dragorn421, line 479: room heat ? room hot ?

Dragorn421 (approved):

PR #157, opened 2022-10-13
[SM64] allow user to export a level from any of its children

thecozies:
This just checks up the tree from your currently selected object to see if it can be exported. Its pretty simple in functionality and the biggest convenience of this was actually being able to map a hotkey to the export level operator so that after i make changes i can just hit ctrl enter from my current object and it exports the whole level!


thecozies:
image

this is where i had mapped it

kurethedead (approved):
This seems good to me. You could also do something like OOT where the root object is stored as a reference in the export settings, so that repeated exports don't require selecting the root.

PR #158, opened 2022-10-13
[Performance] mute unused texture nodes

thecozies:
Its a small performance boost but it can add up.

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -325,39 +325,45 @@ def combiner_uses_alpha(material, checkList, is2Cycle):
 
     return display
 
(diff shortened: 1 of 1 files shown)


Dragorn421 (commented):
afaict looking at the nodes, the "3 Point Lerp" (texel0) node is never mute
I also don't see the texture image nodes themselves being muted

thecozies:
> afaict looking at the nodes, the "3 Point Lerp" (texel0) node is never mute
> I also don't see the texture image nodes themselves being muted
@Dragorn421 

I double checked, all of those were being muted except for the final node (`Tex#_I`), just fixed that and pushed. Here's the proof:

image

Dragorn421 (approved):
works as (I) expected

PR #159, opened 2022-10-14
[SM64] tex scrolling adjustments

thecozies:
i also black formatted f3d_material.py and it caught a few things outside of my change

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -1816,15 +1816,11 @@ class GfxFormatter:
         if scrollDataFields[0].animType == "None" and scrollDataFields[1].animType == "None":
             return ""
 
-        data = "\n".join(
-            (
-                "void scroll_" + name + "() {",
-                "\tint i = 0;",
-                f"\tint count = {count};",
-                f"\tint width = {fScrollData.dimensions[0]} * 0x20;",
-                f"\tint height = {fScrollData.dimensions[1]} * 0x20;",
-            )
-        )
+        data = [
+            "void scroll_" + name + "() {",
+            "\tint i = 0;",
+            f"\tint count = {count};",
+        ]
 
         variables = ""
         currentVars = ""
@@ -1836,6 +1832,7 @@ class GfxFormatter:
             field = "XYZ"[i]
             axis = ["width", "height"][i]
             if scrollDataFields[i].animType != "None":
+                data.append(f"\tint {axis} = {fScrollData.dimensions[i]} * 0x20;")
                 currentVars += "\tstatic int current" + field + " = 0;\n\tint delta" + field + ";\n"
                 checkOverflow += "\n".join(
                     (
@@ -1904,7 +1901,7 @@ class GfxFormatter:
 
         return "\n".join(
             (
-                data,
+                "\n".join(data),
                 variables,
                 currentVars + "\tVtx *vertices = " + segToVirtualFunc + "(" + name + ");",
                 "",
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -1722,39 +1722,47 @@ def update_tex_values_manual(material: bpy.types.Material, context, prop_path=No
     set_texture_settings_node(material)
 
 
-def getMaterialScrollDimensions(material):
-    useDict = all_combiner_uses(material)
+def shift_num(num: int, amt: int):
+    if amt < 0:
+        return num >> -amt
+    return num << amt
 
-    if (
-        useDict["Texture 0"]
-        and material.tex0.tex is not None
-        and useDict["Texture 1"]
-        and material.tex1.tex is not None
-        and material.tex0.tex.size[0] > 0
-        and material.tex0.tex.size[1] > 0
-        and material.tex1.tex.size[0] > 0
-        and material.tex1.tex.size[1] > 0
-    ):
-        if material.uv_basis == "TEXEL0":
-            return material.tex0.tex.size
-        else:
-            return material.tex1.tex.size
-    elif (
-        useDict["Texture 1"]
-        and material.tex1.tex is not None
-        and material.tex1.tex.size[0] > 0
-        and material.tex1.tex.size[1] > 0
-    ):
-        return material.tex1.tex.size
-    elif (
-        useDict["Texture 0"]
-        and material.tex0.tex is not None
-        and material.tex0.tex.size[0] > 0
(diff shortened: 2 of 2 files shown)


fast64_internal/f3d/f3d_material.py
```diff
@@ -1722,39 +1722,47 @@ def update_tex_values_manual(material: bpy.types.Material, context, prop_path=No
...
-    ):
-        return material.tex0.tex.size
+def shift_num(num: int, amt: int):
+    if amt < 0:
+        return num >> -amt
+    return num << amt
+
+
+def shift_dimensions(tex_prop: "TextureProperty", dimensions: tuple[int, int]):
+    shifted = (shift_num(dimensions[0], tex_prop.S.shift), shift_num(dimensions[1], tex_prop.T.shift))
```
Dragorn421, line 1732: About this

The texture is mapped in ST space from `0,0` to `dimensions[0],dimensions[1]`, right

When the scrolling sets a ST position `s,t`, that position becomes `s<<shiftS,t<<shiftT` (assume `<<-x ~ >>x` for simplicity), in the RDP, right

So in the end the scrolling code should ensure `s<<shiftS,t<<shiftT` ranges from `0,0` to `dimensions[0],dimensions[1]` (the ST space range the texture is mapped to), right

That would mean the max for `s,t` should be
```suggestion
    shifted = (shift_num(dimensions[0], -tex_prop.S.shift), shift_num(dimensions[1], -tex_prop.T.shift))
```
instead?

(I'm really not sure. I'm more asking for my own knowledge)

fast64_internal/f3d/f3d_material.py
```diff
@@ -1722,39 +1722,47 @@ def update_tex_values_manual(material: bpy.types.Material, context, prop_path=No
...
+        if f3dMat.tex1.use_tex_reference:
+            texDimensions1 = f3dMat.tex1.tex_reference_size
+        elif f3dMat.tex0.tex:
+            texDimensions1 = (f3dMat.tex1.tex.size[0], f3dMat.tex1.tex.size[1])
+
+    if texDimensions0 is not None:
+        texDimensions0 = shift_dimensions(f3dMat.tex0, texDimensions0)
     else:
-        return [32, 32]
+        texDimensions0 = (4, 4)
```
Dragorn421, line 1758: muh 2x2 or 8x1 textures sadge

fast64_internal/f3d/f3d_material.py
```diff
@@ -1722,39 +1722,47 @@ def update_tex_values_manual(material: bpy.types.Material, context, prop_path=No
...
+
+    if texDimensions1 is not None:
+        texDimensions1 = shift_dimensions(f3dMat.tex1, texDimensions1)
+    else:
+        texDimensions1 = (4, 4)
+
+    return (max(texDimensions0[0], texDimensions1[0]), max(texDimensions0[1], texDimensions1[1]))
```
Dragorn421, line 1765: so what about `material.uv_basis` ?

(I don't really know how UVs work when both tex0 and tex1 are used, based on fast64's `convertVertexData()` both textures seem to use the same ST area though? idk)

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -1904,7 +1901,7 @@ def vertexScrollTemplate(
 
         return "\n".join(
             (
-                data,
+                "\n".join(data),
```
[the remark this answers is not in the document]
thecozies: oh yeah i didnt even consider that here lmao

fast64_internal/f3d/f3d_material.py
thecozies: @Dragorn421 this is specifically for UV scrolling, you are correct with tile scrolling though, although you can set S/T low to numbers higher than the dimensions without any issue. also S/T low is pre-shift/scaling, so if you have a 32/32 texture and set S low to 32, that is always going to be the full width of the texture

fast64_internal/f3d/f3d_material.py
thecozies: its a fail safe if something is undefined 😔 plus the other texture's value will override this since its the max between the two

fast64_internal/f3d/f3d_material.py
thecozies: @Dragorn421 just tested with the UV basis on tex1 to be sure and it works! i tried with shifting tex1 up and down.

fast64_internal/f3d/f3d_material.py
Dragorn421: Uv scrolling and tile scrolling isn't the same thing?

fast64_internal/f3d/f3d_material.py
Dragorn421: Imo would be better as 0,0 and a min(..., ...) in the return?

But really textures can be 1px high afaik
And 2px wide if rgba32 (so the tex line size is 8-aligned, iirc that's the only requirement)

fast64_internal/f3d/f3d_material.py
thecozies: @Dragorn421 nope, tile can be per texture and thats how you get pretty water

fast64_internal/f3d/f3d_material.py
thecozies: > Imo would be better as 0,0 and a min(..., ...) in the return?

why min? this specifically needs the max width/height between the two textures. 

> But really textures can be 1px high afaik
> And 2px wide if rgba32
then `(2, 1)` seems like a sane default min yeah?

Dragorn421: About the min:
Me stupid, I meant max
Like `max(1, dim0, dim1)`

And why not just 1 for both s and t regardless of what's possible, it's not like it matters at that point anyway

thecozies: i guess im just being extra safe but idc either way. i need to get this fix in so i'll just do 1, 1

thecozies: (also i just realized that `max` allows *args 💀)

thecozies: okay @Dragorn421 pushed and updated

fast64_internal/f3d/f3d_material.py
Dragorn421: The PR is correct on this, I was wrong on shiftS being used as a left shift, it's a right shift

Dragorn421 (approved):

PR #160, opened 2022-10-15
Added LOD level option

thecozies:
Input only shows up if LOD is enabled.

image

Here's an example of it working:

https://user-images.githubusercontent.com/79979276/195990966-093db2e4-428b-457e-90bf-2a671de5f59e.mp4

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -410,7 +410,7 @@ def ui_upper_mode(settings, dataHolder, layout: bpy.types.UILayout, useDropdown)
                 box = inputGroup.box()
                 box.alert = True
                 box.label(text="WARNING: Fast64 does not support setting more than two textures.", icon="LIBRARY_DATA_BROKEN")
-                box.label(text="Additional texture tiles will need to be set up manually")
+                box.label(text="Additional texture tiles will need to be set up manually.")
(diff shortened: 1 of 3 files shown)


sauraen (changes requested):
I think the name and description for this parameter should be changed.

As far as I can tell, this parameter being set, which is currently named `lod_level`, gets put into the triangle edge coefficients command sent from the RSP to RDP in the field `Level`. However, the description from the official RDP command summary is "number of mip-maps minus one", and the description from the F3DEX2 documentation is "Maximum number of mipmap levels aside from the first". So, I think this parameter is setting the number of mip maps, where 0 = one texture, 1 = two textures, etc. So the parameter should be called something like `lod_count` (or perhaps even better, replace references to "Texture LOD" with something about "mip maps", as geometry LoD is also supported on N64 and used in OoT). It should have a range of 1-8 and write to the C as this value minus 1.

I believe the RDP will automatically choose which pair of mip maps to interpolate between, on a per-pixel basis, if there are more than two defined. They have to be set up in consecutive tile descriptors. Note that fast64 only supports defining up to two textures, whereas the N64 can have up to eight for this, so perhaps the range of this parameter should only be 1-2.

Also, it looks like there is no importer code for this parameter--ability to import a DL which uses this parameter should be added, like all other material parameters.

Other than those things, this change is fine.

sauraen (approved):

PR #164, opened 2022-11-03
[SM64] Import Cleanup

Yanis42:
Removed every instance of ``from file import *`` in SM64 files

**Important note**: since I have no knowledge about this game I can't test myself if things are broken or not, it'd be nice if a SM64 person could test and tell me if anything broke 🙏


Yanis42:
(about the ``removed sm64_enum_parser as it's unused`` commit I just forgot to merge main before doing these changes)

fast64_internal/sm64/sm64_f3d_writer.py
```diff
@@ -1,12 +1,69 @@
...
-from ..panels import SM64_Panel
-from ..f3d.f3d_writer import *
-from ..f3d.f3d_material import *
-from ..f3d.f3d_gbi import FMaterial
-from .sm64_texscroll import *
-from .sm64_utility import *
+import shutil, copy, bpy, re, os
+from io import BytesIO
+from math import ceil, log, radians
+from mathutils import Matrix, Vector
```
Dragorn421, line 4: I'm personally not a fan of several module imports on the same line (`import shutil, copy, ...`)
nor importing specific things from modules with short and recognizable names (`from math import ceil, ...`)

But it's a style thing that doesn't affect much in the end so I don't mind too much. It's okay like this to me

Dragorn421 (approved):
Looks okay to me but at the end of the day it just needs testing, we've accidentally broken things doing much smaller changes so I would be surprised if everything is fine

Let's say I merge this on December 1st ~~2023~~ 2022 if no testing happens and it'll be real field testing then :sunglasses:

jesusyoshi54:
```
File "C:\Users\scut\AppData\Roaming\Blender Foundation\Blender\3.2\scripts\addons\fast64\fast64_internal\sm64\sm64_geolayout_classes.py", line 142, in sortGeolayouts
    callOrder = copy.copy(callOrder)
AttributeError: 'function' object has no attribute 'copy'
```
relevant line is as follows:
```
                    callOrder = copy.copy(callOrder)
```
and the offending import error stems from changing import copy, to from copy import copy.

fast64_internal/sm64/sm64_geolayout_classes.py
```diff
@@ -1,9 +1,52 @@
-from .sm64_geolayout_constants import *
-from .sm64_geolayout_utility import *
-from .sm64_constants import *
+import bpy
+from struct import pack
+from copy import copy
```
jesusyoshi54, line 3: this change results in an error on line 142.
copy.copy needs to be changed to copy

jesusyoshi54 (approved):

Dragorn421 (approved):

PR #165, opened 2022-11-03
OOT Scene Importing

kurethedead:
This PR adds support for OOT scene importing, along with multiple minor bugfixes, optimizations, scene-related property edits, and new scene-related features.


sauraen (approved):
This is very good work overall and should be merged ASAP. Generally I'm in favor of merging things as soon as they're usable and not broken--better to let the community use something even if isn't perfect, and folks will be more motivated to fix things which are in the main public version than things stuck in branch hell.

What I tested:
- Importing a few decomp scenes
- Importing a non-decomp romhack scene which had been created/exported months ago with fast64

This is NOT a code review; often this project gets code reviews that are not actually usability reviews, so I think the latter is more valuable.

Here are a few suggestions for improvements:

Interface:
- Make the separation between the "Export Scene" and "Import Scene" controls more clear, like it is in the other exporter panes because they have a single button at the top of each.
- Clarify the difference between Remove Scene and OOT Remove Scene (both undocumented).
- It looks like the choose scene for export and for import are the same underlying properties; I'd suggest deduplicating them.

Naming: I'd suggest getting rid of unnecessary generated parts of Blender object names, which will be re-generated at export time. For example:
- spot04_sceneCommands -> spot04_scene
- spot04_room_0Commands -> spot04_room_0
- spot04_sceneCollisionHeader_008918Polygons -> Collision
- spot04_sceneCollisionHeader_008918CamDataList_014 -> CamPos.014

Importing non-decomp scenes: This actually worked better than I expected, though there were still a few problems with older files. With a scene exported from the current fast64 and then immediately re-imported, it worked correctly with no issues. However, with a scene exported from a fast64 version from a few months ago, there were two issues:
- Old format has `&` in front of many names, for example `SCENE_CMD_ALTERNATE_HEADER_LIST(&myScene_alternateHeaders)`, which the current processing doesn't support. This should be easy to strip off the `&` if present.
- Old format has `(u32)&myRoom0_meshDListEntry` and `(u32)&(myRoom0_meshDListEntry)` in `MeshHeader0`; the `(u32)` should also be stripped if possible.
- The `PluginError` for not matching in `parseCollisionHeader` is in the wrong place; it should be moved up a few lines.

kurethedead:
I've implemented all of these changes.

Zelllll (approved):
looks good to me and it seems that the feature is working well for people 👍🏻

ariahiro64:
still getting exported maps using room_0.h instead of scene.h even though scene.h is generated

kurethedead:
I fixed the camera position data (merge error). The room error isn't related to this PR but since its a tiny fix I've added that as well.

fast64_internal/f3d/f3d_writer.py
```diff
@@ -1,3 +1,4 @@
+from email.mime import image
```
Yanis42, line 1: I feel like this is added by auto-completion

Yanis42: (nothing else to add, I just wanted to check if the imports have anything weird)

kurethedead: I swear I deleted this. Why does this keep coming back lmao

Yanis42: sometimes I hate auto-completion lmao

PR #167, opened 2022-11-05
Fixed OOT collision export scale

kurethedead:


Dragorn421 (approved):
duuuh
I should have noticed that

PR #168, opened 2022-11-05
Use object name for OOT exporters

kurethedead:
Instead of using a specified name in export settings, the name of the object is used instead. This applies to skeleton/DL/collision/animation exporters.

### DIFF
--- a/fast64_internal/oot/oot_anim.py
+++ b/fast64_internal/oot/oot_anim.py
@@ -38,7 +38,6 @@ class OOTAnimExportSettingsProperty(bpy.types.PropertyGroup):
     customPath: bpy.props.StringProperty(name="Folder", subtype="FILE_PATH")
     folderName: bpy.props.StringProperty(name="Animation Folder", default="object_geldb")
     isLink: bpy.props.BoolProperty(name="Is Link", default=False)
-    skeletonName: bpy.props.StringProperty(name="Skeleton Name", default="gGerudoRedSkel")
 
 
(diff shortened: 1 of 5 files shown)


fast64_internal/oot/oot_anim.py
```diff
@@ -34,11 +34,12 @@
 
 
 class OOTAnimExportSettingsProperty(bpy.types.PropertyGroup):
+    isCustomFilename: bpy.props.BoolProperty(name="Use Custom Filename")
+    filename: bpy.props.StringProperty(name="Filename")
     isCustom: bpy.props.BoolProperty(name="Use Custom Path")
     customPath: bpy.props.StringProperty(name="Folder", subtype="FILE_PATH")
     folderName: bpy.props.StringProperty(name="Animation Folder", default="object_geldb")
```
Dragorn421, line 41: I've tried reviewing before but I just don't understand how filename / customPath / folderName play together, there may be something to improve here

To give a more concrete suggestion, could the "is custom" bool properties have a description with the behavior difference between not-custom and custom?

kurethedead: This PR changes things so that the filename of an export by default will be based off the object/skeleton/animation name in blender. `filename` is used to provide a custom filename option to override that, if the user wants. `customPath` is for exporting to any non-decomp directory.

sauraen (approved):
Looks okay to me I think. Not quite sure what the impetus for this was, but manually specifying a filename is supported as an option so that's fine. If you have tested this @kurethedead and it works as you expect, go ahead and merge.

PR #172, opened 2022-11-08
Removed object list limit for OOT

kurethedead:
A warning message appears instead in the object list panel if going over the limit.

### DIFF
--- a/fast64_internal/oot/oot_level_writer.py
+++ b/fast64_internal/oot/oot_level_writer.py
@@ -349,8 +349,11 @@ def readRoomData(room, roomHeader, alternateRoomHeaders):
     room.disableSunMoon = roomHeader.disableSunMoon
     room.echo = roomHeader.echo
     room.objectList.extend([getCustomProperty(item, "objectID") for item in roomHeader.objectList])
-    if len(room.objectList) > 15:
-        raise PluginError("Error: A scene can only have a maximum of 15 objects (OOT, not blender objects).")
+
+    if len(room.objectList) > 16:
+        print(
+            "Warning: A room can only have a maximum of 16 objects in its object list, unless more memory is allocated in code.",
+        )
 
     if alternateRoomHeaders is not None:
         if not alternateRoomHeaders.childNightHeader.usePreviousHeader:
--- a/fast64_internal/oot/oot_scene_room.py
+++ b/fast64_internal/oot/oot_scene_room.py
@@ -601,6 +601,9 @@ def drawRoomHeaderProperty(layout, roomProp, dropdownLabel, headerIndex, objName
     elif menuTab == "Objects":
         objBox = layout.column()
         objBox.box().label(text="Objects")
+        if len(roomProp.objectList) > 16:
+            objBox.label(text="You are over the 16 object limit.", icon="ERROR")
+            objBox.label(text="You must allocate more memory in code.")
         for i in range(len(roomProp.objectList)):
             drawObjectProperty(objBox, roomProp.objectList[i], headerIndex, i, objName)
         drawAddButton(objBox, len(roomProp.objectList), "Object", headerIndex, objName)


Yanis42:
might be worth to mention the [location of the alloc size](https://github.com/zeldaret/oot/blob/d8175501adac32dd10918f67f1a7ea278e2ed862/src/code/z_scene.c#L36)

also idk how this is working exactly but users might need to increase [this](https://github.com/zeldaret/oot/blob/d8175501adac32dd10918f67f1a7ea278e2ed862/include/z64object.h#L4) if they want more objects in the list (this number might be 16 for the room's object list + Link's object + gameplay_keep + global scene object (dangeon_keep or field_keep))

maybe these informations could go in the oot readme on fast64?

kurethedead:
Honestly I don't know much about the process myself - I think someone else who can explain it better should add it to the PR.

sauraen:
I think this should either just go in as is--it's a warning, if you have modified the decomp code and you know you're good then you can ignore the warning--or there should be some way to specify or auto-detect from the decomp the current list size. The error message is a little misleading as it's a maximum number of objects in the list, not a maximum memory size (well you can hit that too). But even as is it's better than just giving an error always.

kurethedead:
I've asked @krm01 to add a small description to the readme - after that we can merge.

fig02:
Is there any possibility for a checkbox you can set to disable the warning, or something along those lines?  Or just making the checkbox toggle the ability to go over the limit?

Its obviously not the biggest deal to just ignore the warning if you know your stuff is set up properly, but I think itd be nice to disable it, so you dont see it everytime youre exporting a new scene. The warnings should remain for important issues, instead of having the user train themselves to ignore it imo

BeverlyBean:
a button just for that warning sounds unnecessary

Zelllll:
I agree that it isn't really necessary. If the user is aware that they have already changed the limit in their project, then they should just be able to ignore the warning since they know it doesn't apply to them.

kurethedead:
This isn't a warning on export, just a notice at the top of the object list if there's more than 16 objects. It prints to the console on export, but the user isn't going to be seeing that.

fig02:
alright. I still dont think "the user can just ignore it" is better than a setting the user can explicitly set themselves to allow the cap to be bypassed. If one trains themselves to ignore that warning, it may lead to them ignoring others.
But, if its not on export, then my concern matters less

BeverlyBean:
is it really that hard to ignore a simple warning??

fig02:
i dont think you understand my point, but im not gonna bother explaining it to someone whos being snarky lmao

BeverlyBean:
i guess i dont, i just dont think the "training yourself" thing applies to everybody

Zelllll:
Personally, I just feel that this is a feature that would pollute the UI and ultimately add more confusion than necessary.

Zelllll (approved):
Good changes 👍🏻

fast64_internal/oot/oot_level_writer.py
```diff
@@ -349,8 +349,11 @@ def readRoomData(room, roomHeader, alternateRoomHeaders):
     room.disableSunMoon = roomHeader.disableSunMoon
     room.echo = roomHeader.echo
     room.objectList.extend([getCustomProperty(item, "objectID") for item in roomHeader.objectList])
-    if len(room.objectList) > 15:
-        raise PluginError("Error: A scene can only have a maximum of 15 objects (OOT, not blender objects).")
+
+    if len(room.objectList) > 16:
+        print(
+            "Warning: A room can only have a maximum of 16 objects in its object list, unless more memory is allocated in code.",
+        )
```
fig02, line 356: so the limit is actually 19, though 16 is a good assumption as long as you assume that the "normal" scenario of gameplay_keep/player/subkeep are there

If it wouldnt be annoying to implement, it may be good to account for those other 3 (which are set via other commands) isntead of implying the limit in the code is directly set to 16. At the very least, it may confuse someone who then goes into to expand this slot count when they are expecting 16 but see 19.
But yea, not the biggest deal

fig02 (commented):
unrelated to the conversation above

fast64_internal/oot/oot_level_writer.py
kurethedead: The issue is that blender doesn't do word wrap on labels so its difficult to add more info without truncating the text and making it unreadable. Ideally I would make a warning flash on export, with a checkbox to disable that warning, but the way blender works makes it really difficult to flash warnings without stopping the operator execution.

fig02: yea my comment is less so about adding more info the warning, but possibly making the check more accurate. an implied limit of 16 will get the job done, but if we wanted it to be more precise, it can check if link/subkeep are present from the other commands in the room that supply them

Dragorn421:
Merging, knowing the object list max length logic could use more work

PR #173, opened 2022-11-12
Kcs main

jesusyoshi54:
optimizes syncs, syncs now only happen at the DL start and on revert
adds FMesh bleed and gfx inline exporting
adds in method to reload fast64 while running
added the ability to load 8 textures
added the ability to reference a TMEM value instead of loading a texture
cleaned up f3d_gbi.py string write methods
turned f3d_gbi gbi macros into subclasses of a GbiMacro class and made them all dataclasses


jesusyoshi54:
Separating this PR into several smaller PRs with better code.

PR #174, opened 2022-11-13
GBI Macro Refactoring + F3D Sync Optimization

jesusyoshi54:
cleaned up string formatting in f3d_gbi
removed useless syncs in material generation

these changes are also in the KCS_main PR, but this one was made separately in order to break apart changes

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -1,6 +1,7 @@
 # Macros are all copied over from gbi.h
 from typing import Sequence
-import bpy, os, enum
+from dataclasses import dataclass, fields
+import bpy, os, enum, copy
 from ..utility import *
 
 
@@ -1981,75 +1982,10 @@ class Vtx:
         )
 
     def to_c(self):
-        if bpy.context.scene.decomp_compatible:
-            return (
-                "{{"
-                + ", ".join(
-                    (
-                        (
-                            "{"
-                            + ", ".join(
-                                (
-                                    str(self.position[0]),
-                                    str(self.position[1]),
-                                    str(self.position[2]),
-                                )
-                            )
-                            + "}"
-                        ),
-                        "0",
-                        "{" + str(self.uv[0]) + ", " + str(self.uv[1]) + "}",
-                        (
-                            "{"
-                            + ", ".join(
-                                (
-                                    "0x" + format(self.colorOrNormal[0], "X"),
-                                    "0x" + format(self.colorOrNormal[1], "X"),
-                                    "0x" + format(self.colorOrNormal[2], "X"),
-                                    "0x" + format(self.colorOrNormal[3], "X"),
-                                )
-                            )
-                            + "}"
-                        ),
-                    )
-                )
-                + "}}"
-            )
-        else:
-            return (
-                "{"
-                + ", ".join(
-                    (
-                        str(self.position[0]),
-                        str(self.position[1]),
-                        str(self.position[2]),
-                        "0",
-                        str(self.uv[0]),
-                        str(self.uv[1]),
-                        "0x" + format(self.colorOrNormal[0], "X"),
-                        "0x" + format(self.colorOrNormal[1], "X"),
-                        "0x" + format(self.colorOrNormal[2], "X"),
-                        "0x" + format(self.colorOrNormal[3], "X"),
-                    )
-                )
-                + "}"
-            )
+        def spc(x):
+            return "{" + ", ".join([str(a) for a in x]) + "}"
 
-    def to_sm64_decomp_s(self):
-        return "vertex " + ", ".join(
-            (
-                str(self.position[0]),
-                str(self.position[1]),
-                str(self.position[2]),
-                "0",
-                "0x" + format(self.uv[0], "X"),
-                "0x" + format(self.uv[1], "X"),
-                "0x" + format(self.colorOrNormal[0], "X"),
-                "0x" + format(self.colorOrNormal[1], "X"),
-                "0x" + format(self.colorOrNormal[2], "X"),
-                "0x" + format(self.colorOrNormal[3], "X"),
-            )
-        )
+        return "{{ " + ", ".join([spc(self.position), "0", spc(self.uv), spc(self.colorOrNormal)]) + " }}"
 
 
 class VtxList:
@@ -2079,17 +2015,11 @@ class VtxList:
 
     def to_c(self):
         data = CData()
-        data.header = "extern Vtx " + self.name + "[" + str(len(self.vertices)) + "];\n"
-        data.source = "Vtx " + self.name + "[" + str(len(self.vertices)) + "] = {\n"
+        data.header = f"extern Vtx {self.name}[{len(self.vertices)}];\n"
+        data.source = f"Vtx {self.name}[{len(self.vertices)}] = {{\n"
         for vert in self.vertices:
-            data.source += "\t" + vert.to_c() + ",\n"
+            data.source += f"\t{vert.to_c()},\n"
         data.source += "};\n\n"
-        return data
-
-    def to_sm64_decomp_s(self):
-        data = self.name + ":\n"
-        for vert in self.vertices:
-            data += vert.to_sm64_decomp_s() + "\n"
         return data
 
 
@@ -2104,29 +2034,28 @@ class GfxList:
     def set_addr(self, startAddress, f3d):
         startAddress = get64bitAlignedAddr(startAddress)
         self.startAddress = startAddress
-        print("GfxList " + self.name + ": " + str(startAddress) + ", " + str(self.size(f3d)))
+        print(f"GfxList {self.name}: {str(startAddress)}, {str(self.size(f3d))}")
         return startAddress, startAddress + self.size(f3d)
 
     def save_binary(self, romfile, f3d, segments):
-        print("GfxList " + self.name + ": " + str(self.startAddress) + ", " + str(self.size(f3d)))
+        print(f"GfxList {self.name}: {str(self.startAddress)}, {str(self.size(f3d))}")
         romfile.seek(self.startAddress)
         romfile.write(self.to_binary(f3d, segments))
 
     def size(self, f3d):
-        size = 0
-        for command in self.commands:
-            size += command.size(f3d)
-        return size
+        return sum([command.size(f3d) for command in self.commands])
 
     # Size, including display lists called with SPDisplayList
     def size_total(self, f3d):
-        size = 0
-        for command in self.commands:
-            if isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static:
-                size += command.displayList.size_total(f3d)
-            else:
-                size += command.size(f3d)
-        return size
+        def use_siz_tot(command):
+            return isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static
+
+        return sum(
+            [
+                command.displayList.size_total(f3d) if use_siz_tot(command) else command.size(f3d)
+                for command in self.commands
+            ]
+        )
 
     def get_ptr_addresses(self, f3d):
         ptrs = []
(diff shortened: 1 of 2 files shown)


kurethedead (changes requested):
You should name your PR to something descriptive of what its purpose is. Also, this PR doesn't load in blender - there's some leftover code that's causing some issues. Otherwise, I tested this with a kakariko map export for OOT, and it seems to compile correctly.

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -1984,75 +1984,8 @@ def to_binary(self):
...
-                "0x" + format(self.uv[0], "X"),
-                "0x" + format(self.uv[1], "X"),
-                "0x" + format(self.colorOrNormal[0], "X"),
-                "0x" + format(self.colorOrNormal[1], "X"),
-                "0x" + format(self.colorOrNormal[2], "X"),
-                "0x" + format(self.colorOrNormal[3], "X"),
-            )
-        )
+        spc = lambda x: ", ".join([str(a) for a in x])
+        return (f"{{{{ {{{spc(self.position)}}}, 0, {{{spc(self.uv)}}}, {{{spc(self.colorOrNormal)}}} }}}}")
```
Dragorn421, line 1988: it's not a dealbreaker to me but I think this is a lot less readable than the previous 30-lines version
I really hate `{` in f-strings
but if others are okay i'm okay

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2107,29 +2034,21 @@ def __init__(self, name, tag, DLFormat):
...
     def size_total(self, f3d):
-        size = 0
-        for command in self.commands:
-            if isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static:
-                size += command.displayList.size_total(f3d)
-            else:
-                size += command.size(f3d)
-        return size
+        siz_tot = lambda x: isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static
+        return sum([command.size(f3d) if siz_tot(command) else command.displayList.size_total(f3d) for command in self.commands])
```
Dragorn421, line 2051: the ternary is backwards
also would be nice to name the lambda a better name, maybe after "recurse" idk

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3000,45 +2923,7 @@ def to_binary(self):
...
-                    "0x" + format(self.normal[0], "X"),
-                    "0x" + format(self.normal[1], "X"),
-                    "0x" + format(self.normal[2], "X"),
-                    "0x00",
-                    "0x00, 0x00, 0x00, 0x00",
-                )
-            )
-            + "\n"
-        )
+        return ', '.join([f"0x{a:X}" for a in (*self.color, *self.normal)])
```
Dragorn421, line 2926: please run the black formatter on this file
idk if it's the only thing but use of `'` tells me it needs to be

kurethedead: I actually forgot to mention this myself - there's a couple formatting differences in this file when running black formatter on it.

jesusyoshi54:
I've addressed the comments so far in this last commit

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2737,6 +2649,8 @@ def __init__(self, name, DLFormat):
         self.triangleGroups: list[FTriGroup] = []
         # VtxList
         self.cullVertexList = None
+        # dict of mat: mesh_desc used for inline Gfx only
+        self.MatGroups = dict()
```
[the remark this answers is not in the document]
jesusyoshi54: removed

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2107,29 +2035,26 @@ def __init__(self, name, tag, DLFormat):
     def set_addr(self, startAddress, f3d):
         startAddress = get64bitAlignedAddr(startAddress)
         self.startAddress = startAddress
-        print("GfxList " + self.name + ": " + str(startAddress) + ", " + str(self.size(f3d)))
+        print(f"GfxList {self.name}: {str(startAddress)}, {str(self.size(f3d))}")
         return startAddress, startAddress + self.size(f3d)
 
     def save_binary(self, romfile, f3d, segments):
-        print("GfxList " + self.name + ": " + str(self.startAddress) + ", " + str(self.size(f3d)))
+        print(f"GfxList {self.name}: {str(startAddress)}, {str(self.size(f3d))}")
```
kurethedead, line 2042: `startAddress` -> `self.startAddress`

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2107,29 +2035,26 @@ def __init__(self, name, tag, DLFormat):
...
     # Size, including display lists called with SPDisplayList
     def size_total(self, f3d):
-        size = 0
-        for command in self.commands:
-            if isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static:
-                size += command.displayList.size_total(f3d)
-            else:
-                size += command.size(f3d)
-        return size
+        use_siz_tot = lambda x: isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static
```
kurethedead, line 2051: `lambda x` -> `lambda command`

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -5344,110 +4378,47 @@ def gsDPLoadTileGeneric(c, tile, uls, ult, lrs, lrt):
...
-            + ", "
-            + str(self.lrs)
-            + ", "
-            + str(self.lrt)
-        )
-
-    def size(self, f3d):
-        return GFX_SIZE
+    def is_LOADTILE(self):
+        return self.t == G_TX_LOADTILE
```
kurethedead, line 4393: I think `G_TX_LOADTILE` isn't globally defined like this?

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -5471,76 +4442,17 @@ def to_binary(self, f3d, segments):
...
-            + ", "
-            + str(self.masks)
-            + ", "
-            + str(self.shifts)
-        )
-
-    def size(self, f3d):
-        return GFX_SIZE
+    def is_LOADTILE(self):
+        return self.tile == G_TX_LOADTILE
```
kurethedead, line 4446: Same issue as before.

kurethedead (changes requested):
Sorry, just looked through vscode compile warnings and found a couple more errors.

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -1984,75 +1985,8 @@ def to_binary(self):
...
-                "0",
-                "0x" + format(self.uv[0], "X"),
-                "0x" + format(self.uv[1], "X"),
-                "0x" + format(self.colorOrNormal[0], "X"),
-                "0x" + format(self.colorOrNormal[1], "X"),
-                "0x" + format(self.colorOrNormal[2], "X"),
-                "0x" + format(self.colorOrNormal[3], "X"),
-            )
-        )
+        spc = lambda x: ", ".join([str(a) for a in x])
```
Dragorn421, line 1988: this applies to other lambda usage
PEP8 https://peps.python.org/pep-0008/ :
> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier:

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2082,19 +2016,13 @@ def to_binary(self):
 
     def to_c(self):
         data = CData()
-        data.header = "extern Vtx " + self.name + "[" + str(len(self.vertices)) + "];\n"
-        data.source = "Vtx " + self.name + "[" + str(len(self.vertices)) + "] = {\n"
+        data.header = f"extern Vtx {self.name}[{str(len(self.vertices))}];\n"
```
Dragorn421, line 2019: with f-strings `str` can be omitted

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3093,43 +2949,14 @@ def __init__(self, name, x1, y1, x2, y2):
...
+        return self.x1, self.y1, self.x2, self.y2
+
     def to_binary(self):
-        return (
-            self.x1.to_bytes(4, "big")
-            + self.y1.to_bytes(4, "big")
-            + self.x2.to_bytes(4, "big")
-            + self.y2.to_bytes(4, "big")
-        )
+        return (a.to_bytes(4, "big") for a in self.fields)
```
Dragorn421, line 2956: I think you forgot `@property` on `fields` ? or what did you mean for this to do

fast64_internal/f3d/f3d_gbi.py
jesusyoshi54: In that case, I should change the defined lambdas to be like this then
```suggestion
    def size_total(self, f3d):
        def use_siz_tot(x):
            return isinstance(command, SPDisplayList) and command.displayList.DLFormat != DLFormat.Static
```
```suggestion
    def to_c(self):
        def spc(x):
            return ", ".join([str(a) for a in x])
```

fast64_internal/f3d/f3d_gbi.py
jesusyoshi54: Yeah this was an oversight. I have changed it to use @property

fast64_internal/f3d/f3d_gbi.py
jesusyoshi54: changed to f3d.G_TX_LOADTILE

fast64_internal/f3d/f3d_gbi.py
jesusyoshi54: changed to f3d.G_TX_LOADTILE

fast64_internal/f3d/f3d_gbi.py
jesusyoshi54: I can reduce the number of curly braces by including the braces around parameters inside the unrolling function spc
```
    def to_c(self):
        def spc(x):
            return f"{{', '.join([str(a) for a in x])}}"
        return f"{{{{ {spc(self.position)}, 0, {spc(self.uv)}, {spc(self.colorOrNormal)} }}}}"
```

jesusyoshi54: I could also do it this way, which may be the best as it reduces the number of curly braces. Let me know which way is preferred.
```
    def to_c(self):
        def spc(x):
            return "{" + ', '.join([str(a) for a in x] + "}"
        return f"{{{{ {spc(self.position)}, 0, {spc(self.uv)}, {spc(self.colorOrNormal)} }}}}"
        ```

**Comment by jesusyoshi54** [comment]
I have updated the branch with most of the requested changes, but some formatting is still not committed yet as it is up to preference/pep standards which I am not fully familiar with. I have left comments on these changes with my proposed changes.

If there are any other changes needed based on preference/formatting let me know. I will run black formatter on the next commit once these string formatting changes are done as well.

File: fast64_internal/f3d/f3d_gbi.py (hunk 9 above)
Comment (Dragorn421) [reply fast64_internal/f3d/f3d_gbi.py]: yeah thats the idea

File: fast64_internal/f3d/f3d_gbi.py (hunk 1 above)
Comment (kurethedead) [reply fast64_internal/f3d/f3d_gbi.py]: The latter one is good, maybe something like `"{{ " + ", ".join([spc(self.position), "0", spc(self.uv), spc(self.colorOrNormal)]) + " }}"`, which is pretty clear about brackets.

Comment (jesusyoshi54) [reply fast64_internal/f3d/f3d_gbi.py]: done

**Review by kurethedead — CHANGES_REQUESTED** [review]
I found another issue when looking at some previous conversations.

File: fast64_internal/f3d/f3d_gbi.py [hunk 12]
```diff
@@ -3093,43 +2953,15 @@ def __init__(self, name, x1, y1, x2, y2):
...
```
kurethedead, line 2961: I just realized, the replacement code isn't correct here - it returns a different thing than before. I think this was supposed to be a list comprehension, but even then its returning a list of bytearrays instead of one long bytearray.

kurethedead (changes requested):
Whoops, forgot to start the review, my bad.

fast64_internal/f3d/f3d_gbi.py
jesusyoshi54: replaced with
```
    def to_binary(self):
        return b"".join(a.to_bytes(4, "big") for a in self.fields)
```

kurethedead (approved):
Seems good now, unless anyone has anything else to say I'll merge this in a bit

Dragorn421 (approved):
I didn't take the time to hawkeye this, but didn't notice more things to address

also didn't test the syncs removal nor checked if they're consistent with the discord discussions

kurethedead:
I tested briefly for OOT - I was under the assumption from the discord discussion that this was already tested for correctness for console/emulator? Or did I misread that ?

jesusyoshi54:
I have tested the syncs and they are all good in G_PM_1PRIMITIVE for sure. I have not seen any issue in G_PM_NPRIMITIVE mode except for when using CI textures.
Adding in a DPLoadSync after the TLUT cmd remedies this.
After this fix, in my test levels I have tested every type of texture, several small textures with LoDs and many different combinations of multi textures as well as untextured faces, all without issue.
This should probably be tested on OOT on console to verify compat and maybe someone else should test some more SM64 levels with G_PM_NPRIMITIVE just to double check all cases.

sauraen (approved):
An OoT custom scene exported with this branch works correctly on console with no artifacts that I can see.

image

Besides the main scene geometry here, which of course is also working fine with the removed syncs, the three additional planes which look obviously out of place in the scene are three additional cases of texture setups:
- Two RGBA5551 textures for multitexture
- One RGBA32 texture in single-cycle mode
- One CI4 texture in single-cycle mode

Creating these test cases has revealed some other issues with fast64 texture setups, which I will address in a future PR, but these issues are not caused by changes in this PR. There are also additional changes to syncs I would like to make, but they can build off this PR.

If this has also been tested in SM64 on console, I would suggest merging.

PR #176, opened 2022-11-14
[OoT] Applied new organisation to ``c_writer/``

Yanis42:
Changes:
- nothing added or removed except one small change for the ``OOTScene`` and ``OOTRoom`` classes
- created a new class that contains the common commands (end marker and alt header one, see ``oot_level_classes.py`` diff)
- applied the [new code organisation](https://pastebin.com/HAdmtDaq) as discussed on Discord
- renamed ``c_writer`` to ``exporter`` as this will host the main code for the export (I just didn't know where to place those files)
- removed the oot prefix to the ``oot/exporter/`` files

Actual improvements will come later™


Yanis42:
(level_c should have a better name but I can't find one :peepoShrug:)

kurethedead:
I think the contents of the `/exporter` folder should be merged with `/scene`, since a scene and level are basically synonymous. Also, right now the files in `/actor`, `/cutscene`, and `/collision` are all functions related to scene c exporting. I think these should be put into `scene/exporter` as `scene_actor`, `scene_cutscene`, and `scene_collision`, the same way `room_exporter` has `room_shape` (the `scene_` prefix can be removed if you think its unnecessary). Then `/cutscene` and `/collision` would hold all the code that handles converting blender data into an intermediate representation, as well as any c exporting code that's not done in the context of a scene. If we do this then I don't think `/actor` is necessary, since this code only exists for scene exporting.

Yanis42:
about ``/exporter`` I just didn't know where to place this, I thought the ``scene/`` folder would be good but then I was thinking this folder handles the actual C code for scene files, I'll move that there tho 👍 

and yea I agree, it's better this way

Yanis42:
Changes:
- merged ``exporter/`` with ``scene/exporter/``
- merged scene content export code in ``scene/exporter/``
- merged room export code in ``scene/exporter/`` (but didn't rename the files) so we have all of the "to C" code in the same place (and split the "object properties to export classes" code in the different folders (like ``room/exporter/``, ``actor/exporter/``, etc)
- named ``scene_actor`` just ``actor`` since rooms also use this code (I just thought it'd be weird to keep the scene prefix for that file)

kurethedead (approved):
If you want you can put the `scene/exporter` files into some kind of `c` subfolder (depending on how the other exporter code is organized), but otherwise I'm good with this organization. I've tested on a map export/import and it doesn't seem to have broken anything.

Yanis42:
how about: ``scene/exporter/`` -> ``scene/to_c/`` ?

Yanis42:
alright, moved the files in a ``to_c`` subfolder, should be good to go now 👍 (tested and had no issues with exporting and importing scenes)

kurethedead:
Looks good. Unless anyone has any other comments I'll merge this in a bit.

Dragorn421 (commented):
looks good to me as well, from a very quick glance

PR #182, opened 2022-11-24
Import all Mario animations button.

Lilaa3:
Added a button to import all of mario´s animations easily and with naming.

Commented out an uncessary print at readValueIndex that was slowing down animation import times by about 4 times on my machine.


fast64_internal/sm64/sm64_constants.py
```diff
@@ -1768,3 +1768,216 @@ def __init__(self, geoAddr, level, switchDict):
     ("macro_yellow_coin", "Yellow Coin", "Yellow Coin"),
     ("macro_yellow_coin_2", "Yellow Coin 2", "Yellow Coin 2"),
 ]
+
+marioAnimations = [
+#   ( "animation name",                                             "adress" ),
+    ( "0 - Climb up ledge (right leg going up)",                    5162640  ),
```
Dragorn421, line 1774: this file is formatted, so to keep things simple we should just keep it that way
I understand formatting breaks the column layout you have going on, so what about you swap the columns:
```py
marioAnimations = [
    # (address, "animation name"),
    (5162640, "0 - Climb up ledge (right leg going up)"),
    (5165520, "1 - Thrown from painting"),
    (5165544, "2 - Blown by wind"),
```
(this is after formatting)

fast64_internal/sm64/sm64_anim.py
```diff
@@ -950,7 +951,43 @@ def execute(self, context):
...
+            return {"CANCELLED"}
+        try:
+            if len(context.selected_objects) == 0:
+                raise PluginError("Armature not selected.")
+            armatureObj = context.active_object
+            if type(armatureObj.data) is not bpy.types.Armature:
+                raise PluginError("Armature not selected.")
+
+            for animName, adress in marioAnimations:
+                importAnimationToBlender(romfileSrc, adress, armatureObj, segmentData, context.scene.isDMAImport, animName)
```
Dragorn421, line 989: what is `segmentData` here?

fast64_internal/sm64/sm64_constants.py
Dragorn421: Also could you indicate which rom are these addresses for? unless all roms of sm64 are alike
(I'm a OoT enjoyer and there are many ROM versions from where I come 8) )

fast64_internal/sm64/sm64_anim.py
jesusyoshi54: segmentData is an object describing the segment locations in sm64. It has type dict[segNum] = (segStart, segEnd) and all of them are ints

jesusyoshi54: it seems like segment data is inherited from the global context since it is not defined in that function, which should not affect any operation as far as mario animation importing goes.
other anim importing functions run this to get the segmentData
```
            levelParsed = parseLevelAtPointer(romfileSrc, level_pointers[context.scene.levelAnimImport])
            segmentData = levelParsed.segmentData
```

Dragorn421: thanks
it's not some kind of fast64 global, is it?
I don't see it set in this function

jesusyoshi54: I'm not entirely certain as I didn't trace the code, but I do know that mario animations are in the ROM and not in any specific segment, so the actual data shouldn't be needed. I would change it to an empty dict if possible.

Dragorn421 (approved):
can (has) any sm64 enjoyer test(ed) this? :sunglasses:

Dragorn421:
Ideally testing is carried out by a different person from the person who made the feature, but since this is simple I'm okay with going ahead like this

I'll merge in a few days to give some time for feedback just in case

jesusyoshi54:
I have tested the latest commit and can confirm that it works

PR #183, opened 2022-11-25
[SM64] Allow editing of group level segments on level export

jesusyoshi54:
This commit allows you to change the name of the segment and the group data loaded for segments 5 and 6 and their associated geo segments. You can choose from a common group, use a custom name or inherit the original script data (default behavior).


fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -445,6 +445,29 @@ def replaceSegmentLoad(levelscript, segmentName, command, changedSegment):
     changedLoad[1][2] = segmentName + "SegmentRomEnd"
 
 
+def replaceScriptLoads(levelscript, obj):
+    newFuncs = []
+    for jumpLink in levelscript.levelFunctions:
+        target = jumpLink[1][0]  # format is [macro, list[args], commnent]
```
Dragorn421, line 451:
```suggestion
        target = jumpLink[1][0]  # format is [macro, list[args], comment]
```
I have no idea what the sm64 sorcery is so I'm glad I can point out trivial things FeelsAmazingMan

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -445,6 +445,29 @@ def replaceSegmentLoad(levelscript, segmentName, command, changedSegment):
...
+        if "script_func_global_" not in target:
+            newFuncs.append(jumpLink)
+            continue
+        scriptNum = int(re.findall("\d+", target)[-1])
```
Dragorn421, line 455:
```suggestion
        scriptNum = int(re.findall(r"\d+", target)[-1])
```
so as to not rely on `'\d'` -> `'\\d'`

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1936,6 +1995,9 @@ def sm64_obj_register():
     bpy.types.Object.sm64_obj_use_act5 = bpy.props.BoolProperty(name="Act 5", default=True)
     bpy.types.Object.sm64_obj_use_act6 = bpy.props.BoolProperty(name="Act 6", default=True)
 
+    # write to level script out
+    bpy.types.Object.sm64_segment_loads = bpy.props.PointerProperty(type=SM64_SegmentProperties)
```
Dragorn421, line 1999: please add this propgroup to `SM64_ObjectProperties`

then access like `obj.fast64.sm64.segment_loads` instead of `obj.sm64_segment_loads`

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1845,6 +1850,59 @@ def upgrade_changed_props():
...
+            prop_split(col, self, "group5_seg_str", "Segment 5 Seg")
+            prop_split(col, self, "group5_str", "Segment 5 Group")
+        col = layout.column()
+        prop_split(col, self, "group6_enum", "Segment 6 Select")
+        if self.group6_enum == "Custom":
+            prop_split(col, self, "group6_seg_str", "Segment 6 Seg")
+            prop_split(col, self, "group6_str", "Segment 6 Group")
+
+    def convert_group(self, grp):
+        num = int(re.findall("\d+", grp)[-1]) + 1
```
Dragorn421, line 1874:
```suggestion
        num = int(grp.removeprefix("group")) + 1
```
seems like a clearer intent to me
can "convert_group" have a better name? I can't really suggest anything due to not knowing what this is about but something like `group_to_script_func_name` idk :shrug:

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1845,6 +1850,59 @@ def upgrade_changed_props():
             obj.fast64.sm64.version = SM64_ObjectProperties.cur_version
 
 
+class SM64_SegmentProperties(bpy.types.PropertyGroup):
+    group5_seg_str: bpy.props.StringProperty(name="Segment 5 Seg")
+    group5_str: bpy.props.StringProperty(name="Segment 5 Group")
+    group6_seg_str: bpy.props.StringProperty(name="Segment 6 Seg")
+    group6_str: bpy.props.StringProperty(name="Segment 6 Group")
+    group5_enum: bpy.props.EnumProperty(name="Segment 5 Group", default="Do Not Write", items=groupsSeg5)
+    group6_enum: bpy.props.EnumProperty(name="Segment 6 Group", default="Do Not Write", items=groupsSeg6)
```
Dragorn421, line 1859: "group5_" sounds like "Group 5" but the name is more like "Group of Segment 5", which is it?

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1845,6 +1850,59 @@ def upgrade_changed_props():
             obj.fast64.sm64.version = SM64_ObjectProperties.cur_version
 
 
+class SM64_SegmentProperties(bpy.types.PropertyGroup):
+    group5_seg_str: bpy.props.StringProperty(name="Segment 5 Seg")
+    group5_str: bpy.props.StringProperty(name="Segment 5 Group")
+    group6_seg_str: bpy.props.StringProperty(name="Segment 6 Seg")
+    group6_str: bpy.props.StringProperty(name="Segment 6 Group")
```
Dragorn421, line 1857: imo "str" would be better as something akin to "custom"

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: It is the group for segment 5. I can rename the things with "group" in them as "seg"

Dragorn421: `seg5_group`
`seg5_custom_seg`
`seg5_custom_group`
may be good?

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: renamed the function to jump_link_from_enum.

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: I chose seg5_load_custom, seg5_group_custom, and seg5_enum

jesusyoshi54:
I've addressed the comments of Dragorn's review in the latest commit.

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1834,6 +1894,8 @@ class SM64_ObjectProperties(bpy.types.PropertyGroup):
     level: bpy.props.PointerProperty(type=SM64_LevelProperties)
     area: bpy.props.PointerProperty(type=SM64_AreaProperties)
     game_object: bpy.props.PointerProperty(type=SM64_GameObjectProperties)
+    # write to level script out
+    sm64_segment_loads: bpy.props.PointerProperty(type=SM64_SegmentProperties)
```
Dragorn421, line 1898:
```suggestion
    segment_loads: bpy.props.PointerProperty(type=SM64_SegmentProperties)
```

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1826,6 +1831,61 @@ def get_behavior_params(self):
         return self.bparams
 
 
+class SM64_SegmentProperties(bpy.types.PropertyGroup):
+    seg5_load_custom: bpy.props.StringProperty(name="Segment 5 Seg")
+    seg5_group_custom: bpy.props.StringProperty(name="Segment 5 Group")
+    seg6_load_custom: bpy.props.StringProperty(name="Segment 6 Seg")
+    seg6_group_custom: bpy.props.StringProperty(name="Segment 6 Group")
+    seg5_enum: bpy.props.EnumProperty(name="Segment 5 Group", default="Do Not Write", items=groupsSeg5)
+    seg6_enum: bpy.props.EnumProperty(name="Segment 6 Group", default="Do Not Write", items=groupsSeg6)
```
Dragorn421, line 1840:
```suggestion
    seg5_group: bpy.props.EnumProperty(name="Segment 5 Group", default="Do Not Write", items=groupsSeg5)
    seg6_group: bpy.props.EnumProperty(name="Segment 6 Group", default="Do Not Write", items=groupsSeg6)
```

Dragorn421: sorry to be picky with these names btw but since we can't easily change them, may as well be willing to improve them now

jesusyoshi54: these enums are used for both the segment load and the group that is loaded using the script, so I don't really want to call them "group"

jesusyoshi54: Colloquially speaking hackers use that name, and it is how people will select, but I use it to add the load because that is implicitly defined by the group.
I want to make it clear that when using string input they are defined separately, but when using the enum the load and group are defined.

thecozies (approved):
tested with two random groups with no issue. also correctly retained the groups by default which is perfect


https://user-images.githubusercontent.com/79979276/209354015-ee4c2fda-cb37-4212-bfe3-32ba9e5cf14b.mp4

PR #184, opened 2022-11-26
Added feature to preserve include(s) at the top of spec for HackerOOT

sauraen:
`#include "include/config.h"` and other includes at the top of the spec will be preserved when fast64 modifies the spec. Or rather, all includes which were present anywhere in the original file will be re-added at the top of the new file when it is written. This is required for HackerOOT.


ariahiro64:
This PR fixes a HackerOOT PR that switches debugging features on and off https://github.com/HackerN64/HackerOOT/pull/15

ariahiro64 (approved):

Dragorn421 (approved):

PR #187, opened 2022-11-28
[OoT] Exporter code cleanup

Yanis42:
Changes: basically renamed and improved every functions/class and a lot of local variables, made an import cleanup pass on every (OoT) files and formatted with black, I also improved the C code a bit to make it easier to read/understand


Yanis42:
fixed conflicts, hopefully nothing broke for the spec changes 🙏 (idk how to test that properly, added an include in the spec file and it stayed so I believe it works as intended)

fast64_internal/oot/scene/exporter/to_c/actor.py
```diff
@@ -9,45 +9,38 @@
...
-                        str(int(round(actor.rotation[2]))),
-                    )
-                ),
-                str(actor.actorParam),
-            )
-        )
-        + " },\n"
+        (f"// {actor.actorName}\n" + indent if actor.actorName != "" else "")
+        + "{\n"
+        + ",\n".join((indent * 2) + f"/* {desc:10} */ {info}" for desc, info in zip(infoDescs, actorInfos))
```
kurethedead, line 23: I'm not sure if comments like this break scene importing, have you tried it out?

fast64_internal/oot/scene/exporter/to_c/room_commands.py
```diff
@@ -2,134 +2,77 @@
...
-def ootRoomCommandsToC(room, headerIndex):
-    commands = []
-    if room.hasAlternateHeaders():
-        commands.append(room.cmdAltHeaders(room.roomName(), room.alternateHeadersName(), headerIndex, len(commands)))
-    commands.append(cmdEchoSettings(room, headerIndex, len(commands)))
-    commands.append(cmdRoomBehaviour(room, headerIndex, len(commands)))
-    commands.append(cmdSkyboxDisables(room, headerIndex, len(commands)))
-    commands.append(cmdTimeSettings(room, headerIndex, len(commands)))
-    if room.setWind:
-        commands.append(cmdWindSettings(room, headerIndex, len(commands)))
```
kurethedead, line 118: I think this check is not handled correctly in the new code?

fast64_internal/oot/oot_level_writer.py
```diff
@@ -270,6 +259,7 @@ def readSceneData(scene, scene_properties, sceneHeader, alternateSceneHeaders):
     scene.skyboxID = getCustomProperty(sceneHeader, "skyboxID")
     scene.skyboxCloudiness = getCustomProperty(sceneHeader, "skyboxCloudiness")
     scene.skyboxLighting = getCustomProperty(sceneHeader, "skyboxLighting")
+    scene.lightMode = sceneHeader.skyboxLighting
```
kurethedead, line 262: I think I mentioned this before, but I don't know why this is being re-added; `lightMode` is the same property as `skyboxLighting` except it doesn't handle the custom option correctly.

kurethedead (changes requested):
I would test scene importing with this PR again. Namely, adding comments and reordering commands may break certain things.

fast64_internal/oot/scene/exporter/to_c/actor.py
Yanis42: I tried it with vanilla scenes and it worked, I forgot you can import fast64 scenes now lol (also does it work properly with non-single file mode?)

fast64_internal/oot/oot_level_writer.py
Yanis42: yea sorry, I was just confused by the new enum for a moment and forgot to remove this lol

fast64_internal/oot/scene/exporter/to_c/actor.py
Yanis42: (it doesn't seem to work with non-single file mode)

Yanis42: just in case, I'm working on a fix for this

Yanis42: fast64 scenes import should now be fixed, idk how to check if everything works for sure but I fixed every errors

something better can be used to detect if light settings are using the new format but I wasn't inspired lol, though the multiple files issue isn't fixed as it wasn't brought by this PR (and I have no idea how to fix that)

anyway, tell me if anything else broke 👍

fast64_internal/oot/scene/exporter/to_c/scene_header.py
```diff
@@ -1,208 +1,208 @@
...
+    fogData = [
+        (light.getBlendFogNear(), "Blend Rate & Fog Near"),
+        (f"{light.fogFar}", "Fog Far"),
+    ]
+
+    lightDescs = ["Dawn", "Day", "Dusk", "Night"]
+    if lightMode == "false":
+        lightDesc = f"// {lightDescs[index]} Lighting\n"
+    else:
+        lightDesc = f"// {'Indoor' if lightMode == 'true' else 'Custom'} n°{index + 1} Lighting\n"
```
kurethedead, line 40: This code throws an index error when exporting a kakariko import. I don't think I actually fully figured out how the lighting system worked, so the lighting mode enum representation might not be completely accurate. Kakariko has a `false` lighting mode but has 12 lights in it.

Yanis42: that's odd, the lighting is set to custom, the condition shouldn't return true

kurethedead: That member handles the custom property, so its going be equal to `"false"`.

Yanis42: should be fixed now

fast64_internal/oot/oot_level_writer.py
```diff
@@ -258,7 +258,8 @@ def readSceneData(scene, scene_properties, sceneHeader, alternateSceneHeaders):
     scene.naviCup = getCustomProperty(sceneHeader, "naviCup")
     scene.skyboxID = getCustomProperty(sceneHeader, "skyboxID")
     scene.skyboxCloudiness = getCustomProperty(sceneHeader, "skyboxCloudiness")
-    scene.skyboxLighting = getCustomProperty(sceneHeader, "skyboxLighting")
+    scene.skyboxLighting = sceneHeader.skyboxLighting
+    scene.skyboxLightingCustom = sceneHeader.skyboxLightingCustom
```
kurethedead, line 262: Sorry, I'm not a fan of this - the `OOTScene` is supposed to be a representation of the scene in C, so `skyboxLighting` shouldn't include details of the property implementation in Blender. Since this is only for comments I would prefer just simplifying the comments to not rely on this, but if you want to keep it an alternative would be to add something like an `isSkyboxLightingCustom` boolean to handle the comments, while retaining `getCustomProperty` for `skyboxLighting.`

fast64_internal/oot/scene/exporter/to_c/scene_header.py
```diff
@@ -34,6 +34,7 @@ def getLightSettingsEntry(light: OOTLight, lightMode: str, index: int):
     ]
 
     lightDescs = ["Dawn", "Day", "Dusk", "Night"]
+    print(lightMode)
```
kurethedead, line 37: This can be removed as well.

fast64_internal/oot/oot_level_writer.py
Yanis42: chose the bool method as it was easier to do

kurethedead (approved):

fast64_internal/oot/scene/exporter/to_c/room_header.py
```diff
@@ -1,78 +1,87 @@
...
+    # .c
+    altHeaderPtrList = (
+        f"{altHeaderPtrListName}[]"
+        + " = {\n"
+        + "\n".join(
+            indent + f"{curHeader.roomName()}_header{i:02},"
+            if curHeader is not None
+            else indent + "NULL,"
+            if i < 4
+            else ""
```
Dragorn421, line 65: I'm wondering if this (omitting None cutsceneHeaders) can ever happen, and if it does, does something break?

fast64_internal/oot/scene/exporter/to_c/scene_pathways.py
```diff
@@ -1,38 +1,48 @@
...
+
+    # .h
+    pathData.header = f"extern {pathName}[];\n"
+
+    # .c
+    pathData.source = (
+        f"{pathName}[]"
+        + " = {\n"
+        + "\n".join(
+            indent + "{ " + ", ".join(f"{round(curPoint):5}" for i, curPoint in enumerate(point) if i < 3) + " },"
```
Dragorn421, line 18: why `if i < 3`
if a point is 4d or more I'd consider that a bug anyway

fast64_internal/oot/oot_level_classes.py
```diff
@@ -187,10 +185,10 @@ def __init__(self, name, model):
         self.musicSeq = None
         self.nightSeq = None
 
-        self.childNightHeader = None
-        self.adultDayHeader = None
-        self.adultNightHeader = None
-        self.cutsceneHeaders = []
+        self.childNightHeader: "OOTScene" = None
+        self.adultDayHeader: "OOTScene" = None
+        self.adultNightHeader: "OOTScene" = None
```
Dragorn421, line 190: `Optional[OOTScene]`

fast64_internal/oot/scene/exporter/to_c/scene_header.py
```diff
@@ -1,208 +1,214 @@
...
-        sceneMainC.append(ootLightSettingsToC(scene, scene.skyboxLighting == "true", headerIndex))
+            if outScene.hasAlternateHeaders() and i == 0:
+                altHeaderListName = f"SceneCmd* {outScene.alternateHeadersName()}[]"
+                sceneC.header += f"extern {altHeaderListName};\n"
+                sceneC.source += altHeaderListName + " = {\n" + altHeaderPtrs + "\n};\n\n"
 
-    # Write the path data, if used
-    sceneMainC.append(pathData)
+                # Write the room segment list
+                sceneC.append(getRoomList(outScene))
```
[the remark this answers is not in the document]
Yanis42: should be fixed now

fast64_internal/oot/scene/exporter/to_c/scene_header.py
```diff
@@ -1,208 +1,214 @@
...
+
+    fogData = [
+        (light.getBlendFogNear(), "Blend Rate & Fog Near"),
+        (f"{light.fogFar}", "Fog Far"),
+    ]
+
+    lightDescs = ["Dawn", "Day", "Dusk", "Night"]
+
+    if not isLightingCustom and lightMode == "false":
+        lightDesc = f"// {lightDescs[index]} Lighting\n"
```
Dragorn421, line 39: what if `index >= 4` ?

Note the first four are for lightConfig 0, the next four are also dawn/day/dusk/night (but for lightConfig 1), ...

See `sTimeBasedLightConfigs` in the oot decomp for details, this pattern works for lightConfig 0, 1, 2, 3,
but lightConfig 4 uses settings 20 to 23 pepega

There's also a bunch of more or less hardcoded stuff. The switching between lightConfigs is mostly typically done by the en_weather_tag actor, see the `WeatherMode` enum

Dragorn421 (approved):

Dragorn421:
@kurethedead if you're okay with the latest changes (mostly fixes) we can merge whenever imo

PR #188, opened 2022-11-28
[OoT] Improvements and fixes for scene exporter panel

Yanis42:
Changes:
- Added custom path option for decomp mode, the remove operator grays out if the path doesn't exists
- Added a box for HackerOoT export options to make it less confusing

### DIFF
--- a/fast64_internal/oot/oot_level_writer.py
+++ b/fast64_internal/oot/oot_level_writer.py
@@ -1031,7 +1031,7 @@ class OOT_ExportScenePanel(OOT_Panel):
         col.prop(settings, "customExport")
 
         if context.scene.fast64.oot.hackerFeaturesEnabled:
-            hackerOoTBox = col.box()
+            hackerOoTBox = col.box().column()
             hackerOoTBox.label(text="HackerOoT Options")
 
             bootOptions = context.scene.fast64.oot.bootupSceneOptions
@@ -1095,7 +1095,7 @@ class OOT_RemoveScenePanel(OOT_Panel):
 
         if removeSettings.option == "Custom" and not os.path.exists(exportPath):
             removeRow.enabled = False
-            col.label(text="This path doesn't exists.")
+            col.label(text="This path doesn't exist.")
         else:
             removeRow.enabled = True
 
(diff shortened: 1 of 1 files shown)


fast64_internal/oot/oot_level_writer.py
```diff
@@ -1003,63 +1001,103 @@ def draw(self, context):
...
                 prop_split(col, settings, "subFolder", "Subfolder")
                 prop_split(col, settings, "name", "Name")
 
         prop_split(col, context.scene, "ootSceneExportObj", "Scene Object")
 
+        col.prop(settings, "singleFile")
+        col.prop(settings, "customExport")
+
         if context.scene.fast64.oot.hackerFeaturesEnabled:
+            hackerOoTBox = col.box()
```
kurethedead, line 1034:
```suggestion
            hackerOoTBox = col.box().column()
```
Don't use the results of box() directly, this causes big spaces between lines.

fast64_internal/oot/oot_level_writer.py
```diff
@@ -1003,63 +1001,103 @@ def draw(self, context):
...
+            exportPath = (
+                context.scene.ootDecompPath + f"assets/scenes/{removeSettings.subFolder}/{removeSettings.name}/"
+            )
+
+        removeRow = col.row()
+        removeRow.operator(OOT_RemoveScene.bl_idname, text="Remove Scene")
+
+        if removeSettings.option == "Custom" and not os.path.exists(exportPath):
+            removeRow.enabled = False
+            col.label(text="This path doesn't exists.")
```
kurethedead, line 1098:
```suggestion
            col.label(text="This path doesn't exist.")
```

kurethedead (changes requested):
I'm opposed to splitting up the scene operators into separate panels - I did this before with SM64 and it ended up being so overwhelming that the panels were merged together.

Yanis42:
The alternative is using ``col.box()`` to make things easier to read, I can revert these changes to that alternative if that's ok 👍

fast64_internal/oot/oot_level_writer.py
Yanis42: good to know

kurethedead (approved):
Seems good to me, will merge unless there any other objections.

PR #189, opened 2022-11-28
[bugfix] add hashing to gbi base classes

jesusyoshi54:
Title says it all. This fixes a bug with the last change to gbi base classes which made them unhashable, and therefore unable to be added to sets/dicts: a function that is used in material overrides for armature exports.

The danger of this is that mutating the objects after they are stored inside a hashtable object is they will become unsearchable in these objects. An example is given here (from stack exchange)
```python
cat = Category('foo', 'bar') # a dataclass with unsafe_hash = True
categories = {cat}
cat.id = 'baz'

print(cat in categories)  # False
```
Alternate means of fixing this bug would be to create a base class implementation of `__hash__` that inherited from the `object` type. This has the benefit of maintaining state within hashtables if mutated, but not generating equal hashes if the classes are equal.

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -3146,7 +3146,7 @@ def gsSPNoOp(f3d):
 
 
 # base class for gbi macros
-@dataclass
+@dataclass(unsafe_hash = True)
 class GbiMacro:
     _segptrs = False
     _ptr_amp = False
@@ -3179,7 +3179,7 @@ class GbiMacro:
         return GFX_SIZE
 
 
-@dataclass
+@dataclass(unsafe_hash = True)
 class SPMatrix(GbiMacro):
     matrix: int
     param: int
@@ -3196,7 +3196,7 @@ class SPMatrix(GbiMacro):
 # Divide mesh drawing by materials into separate gfx
 
 
-@dataclass
+@dataclass(unsafe_hash = True)
 class SPVertex(GbiMacro):
     # v = seg pointer, n = count, v0  = ?
     vertList: VtxList
@@ -3233,7 +3233,7 @@ class SPVertex(GbiMacro):
         return header + ", " + str(self.count) + ", " + str(self.index) + ")"
 
 
-@dataclass
+@dataclass(unsafe_hash = True)
 class SPViewport(GbiMacro):
     # v = seg pointer, n = count, v0  = ?
     viewport: Vp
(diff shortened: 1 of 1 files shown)


Dragorn421:
Why not frozen=True instead of unsafe_hash?

jesusyoshi54:
certain exports do mutate the base classes, leading to an exception

Dragorn421:
I'm not opposed to merging this as a temporary fix but I think long-term it would be better to commit to these classes being either frozen or mutable

I think it'd be simpler if they were frozen. Can you point me to code that assumes mutability?
I'd like to see if
`gbiThing = dlist[1]; gbiThing.value = 0` could be turned into
`dlist[1] = GbiThing(value=0, otherValue=gbiThing.otherValue)`
(edit the dlist with new instances)

jesusyoshi54:
The class `class DPSetRenderMode(GbiMacro)` itself is one example.
As well as some of the methods of constructing it and other code with flaglists, such as `saveOtherModeHDefinitionAll` and `saveOtherModeLDefinitionAll`
There is also the recent flipbook code added which edits TLUT cmds in `modifyDLForCIFlipbook`, though this may not be the case anymore with the recent change to flipbook code.

Dragorn421:
tbh it looks pretty straightforward to change
```py
        cmd = SPSetOtherMode("G_SETOTHERMODE_L", 0, 3, [])
        cmd.flagList.append(settings.g_mdsft_alpha_compare)
        cmd.flagList.append(settings.g_mdsft_zsrcsel)
```
and the likes to
```py
        flagList = []
        flagList.append(settings.g_mdsft_alpha_compare)
        flagList.append(settings.g_mdsft_zsrcsel)
        cmd = SPSetOtherMode("G_SETOTHERMODE_L", 0, 3, tuple(flagList)) # (change list to tuple too for immutability eventually maybe)
```

is there stuff that is less straightforward to adapt to having the classes immutable?

and yeah I think the OoT flipbook dlist editing stuff is all gone in #190 ? @kurethedead

jesusyoshi54:
I'm not certain if this is all the code that mutates these classes, I would have to parse through game specific f3d writers to make sure and I don't really know those code bases at all.

Dragorn421:
> certain exports do mutate the base classes, leading to an exception

then what did you refer to here?

jesusyoshi54:
The examples I posted about earlier were the ones causing exceptions.
An export that uses other modes was causing a failed export in my first test.

I would like to make these frozen but I can't guarantee to know all the cases where classes are edited, so I believe it is safer to use unsafe_hash until all of those cases are resolved.

sauraen:
What is this blocked waiting on?

Dragorn421:
I think the fix for the underlying issue should be implemented differently, by making the gbi classes frozen instead of mutable+hashable (unsafe_hash)

However making the gbi classes frozen apparently causes issues with sm64, so some stuff (I don't know what) would need rewriting

The example of `SPSetOtherMode` above is very straightforward to rewrite, so I'm waiting on either
- a list of the other things that break from making the dataclasses frozen
- a sm64 dev to pick this up
- we decide this is fine (i.e. you guys do, because I don't think it is) and we go with this fix anyway

thecozies (approved):

PR #195, opened 2022-12-04
Added Gfx tag system + Split vertex/tile scrolling

kurethedead:
I added a tag field to `GbiMacro`, which is what tile scrolling now uses to locate the `SPSetTileSize` command. I've also split tile scrolling and vertex scrolling in their own functions, as tile scrolling is per `FMaterial` and vertex scrolling is per `FTriangleGroup`. `fModel.to_c_scroll()` is now used to get all scrolling code. Also note that I moved `vertexScrollTemplate()` out of `GfxFormatter`, which makes the diff look really ugly but its not as bad as it looks.


sauraen (commented):
I can confirm that exporting a custom scene in OoT with this branch works correctly in game.

The scene has OoT-style tile scrolling, but this does not use the mechanism which is the subject of this PR, which to my knowledge is SM64 only. I don't have any SM64 modding experience so I have not tested this.

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3168,6 +3214,16 @@ class GbiMacro:
     _segptrs = False
     _ptr_amp = False
 
+    tag = GfxTag.NoTag
+    """Type: GfxTag. The tag's current use is to determine how to write gfx scrolling code for this given command."""
+
+    fMaterial = None
```
Dragorn421, line 3220: I can see this being an issue with having hashing on GbiMacro instances?

kurethedead: Would implementing id() based hashing fix that issue?

Dragorn421: Bureauctopus/Jesusyoshi said that without annotations, these wouldn't be "part" of the dataclass so not used for the hash (and equality comparison I assume)

so it apparently wouldn't be an issue, but worth making sure and documenting

kurethedead: I actually tried adding annotations, but I got an error about constructor arguments related to dataclasses. I'll add some comments on it not being part of the hash.

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2158,35 +2183,71 @@ def getCurrentAreaKey(self, material):
...
 class FModel:
-    def __init__(self, f3dType, isHWv1, name, DLFormat, matWriteMethod):
+    def __init__(self, f3dType: F3D, isHWv1: bool, name: str, DLFormat: "DLFormat", matWriteMethod: GfxMatWriteMethod):
         self.name = name  # used for texture prefixing
         # dict of light name : Lights
-        self.lights = {}
+        self.lights: TypedDict[str, Lights] = {}
         # dict of (texture, (texture format, palette format)) : FImage
-        self.textures = {}
+        self.textures: TypedDict[Union[FImageKey, FPaletteKey] : FImage] = {}
```
Dragorn421, line 2229:
```suggestion
        self.textures: TypedDict[Union[FImageKey, FPaletteKey], FImage] = {}
```
also afaict this isn't what `TypedDict` is for https://peps.python.org/pep-0589/ (just use `dict` ?)

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -23,6 +23,12 @@ class GfxListTag(enum.Enum):
     Draw = 3
 
 
+class GfxTag(enum.Enum):
+    NoTag = 0
+    TileScroll0 = 1
+    TileScroll1 = 2
```
Dragorn421, line 29:
```suggestion
class GfxTag(enum.Enum):
    NoTag = enum.auto()
    TileScroll0 = enum.auto()
    TileScroll1 = enum.auto()
```

Thought on making this `enum.Flag` instead of an Enum? then a single macro can be tagged different things

e.g. `macro.tags |= GfxTag.TAG_FOO` `macro.tags & GfxTag.TAG_BAR`

fast64_internal/f3d/f3d_gbi.py
kurethedead: Whoops, didn't realize. I'll edit that.

fast64_internal/f3d/f3d_writer.py
```diff
@@ -2150,7 +2113,11 @@ def saveTextureLoading(
                 ]
             )  # added in
 
+    # Tag tile size command for tile scrolling
     tileSizeCommand = DPSetTileSize(f3d.G_TX_RENDERTILE + texIndex, sl, tl, sh, th)
+    tileSizeCommand.tags = GfxTag.TileScroll0 if texIndex == 0 else GfxTag.TileScroll1
```
Dragorn421, line 2118:
```suggestion
    tileSizeCommand.tags |= GfxTag.TileScroll0 if texIndex == 0 else GfxTag.TileScroll1
```
(imagining there may be other tags at play, `=` would overwrite them)

kurethedead: The default is GfxTag.NoTag, so its the desired behaviour, unless there's a way to specify no flags.

Dragorn421: Oh right, NoTag should be removed, you can use `GfxTag(0)` for "no flags"

kurethedead: Gotcha, updated.

Dragorn421 (commented):
`sm64_geolayout_writer.py` has:
```
scrollData, hasScrolling = fModel.to_c_vertex_scroll(scrollName, gfxFormatter)
```
that needs updating, doesn't it?

kurethedead:
Turns out that needed a bit of a refactor, but its fixed now.

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2493,23 +2548,49 @@ def to_c(self, textureExportSettings: TextureExportSettings, gfxFormatter: GfxFo
...
-        for scrollFunc in scrollDefinesSplit:
-            if scrollFunc == "":
-                continue
-            scrollData.source += "\t" + scrollFunc
-        scrollData.source += "}\n"
+        data.topLevelScrollFunc = f"scroll_{funcName}"
+        data.source += f"void {data.topLevelScrollFunc}() {{\n"
+        for scrollFunc in data.functionCalls:
+            data.source += f"\t{scrollFunc}();\n"
+        data.source += f"}};\n"
```
thecozies, line 2565: this doesnt really need to be an f string since there aint no formmattin i reckon

thecozies (approved):
Small formatting change suggestion, but looks good and works! No compilation warnings as well.

https://user-images.githubusercontent.com/79979276/211585096-d00e7bd5-de8d-4ec6-8b2d-344b4af6ee6d.mp4

fast64_internal/f3d/f3d_gbi.py
kurethedead: I think it makes the curly brace balance more clear, especially with vscode syntax coloring

thecozies: yeah good point

PR #199, opened 2022-12-07
[OoT] Fixed issues with recent decomp changes

Yanis42:
Fast64 broke so here's the fix, the issue was the scene IDs and the draw config names, fixed both and it seems to work properly


fast64_internal/oot/oot_constants.py
```diff
@@ -708,56 +708,169 @@
...
-    ("SDC_TURIBORI", "Fishing Pond (Turibori)", "Turibori"),
-    ("SDC_GANON_SONOGO", "Ganon's Tower (Collapsing) (Ganon Sonogo)", "Ganon Sonogo"),
-    ("SDC_GANONTIKA_SONOGO", "Inside Ganon's Castle (Collapsing) (Ganontika Sonogo)", "Ganontika Sonogo"),
-]
+    ("SDC_FISHING_POND", "Fishing Pond (Turibori)", "Turibori"),
+    ("SDC_GANONS_TOWER_COLLAPSE_INTERIOR", "Ganon's Tower (Collapsing) (Ganon Sonogo)", "Ganon Sonogo"),
+    ("SDC_INSIDE_GANONS_CASTLE_COLLAPSE", "Inside Ganon's Castle (Collapsing) (Ganontika Sonogo)", "Ganontika Sonogo"),
+]
+
+ootSceneNameToDebugName = {
```
Dragorn421, line 765: instead of this, have 
```py
ootSceneNameToID = {val: key for key, val in ootSceneIDToName.items()}
```
and use `ootSceneNameToID` instead of `ootSceneNameToDebugName` as appropriate

Yanis42: I hope this was what you had in mind

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -103,12 +103,13 @@ def getSceneParams(scene, exportInfo, sceneNames):
...
     sceneIndex = getSceneIndex(sceneNames, bpy.context.scene.ootSceneExportSettings.option)
     sceneName = sceneTitle = sceneID = sceneUnk10 = sceneUnk12 = None
+    name = scene.name if scene is not None else exportInfo.name
 
     # if the index is None then this is a custom scene
     if sceneIndex is None and scene is not None:
         sceneName = scene.name.lower() + "_scene"
         sceneTitle = "none"
-        sceneID = "SCENE_" + (scene.name.upper() if scene is not None else exportInfo.name.upper())
+        sceneID = ootSceneNameToID.get(name, f"SCENE_{name.upper()}")
```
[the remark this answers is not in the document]
Yanis42: we need the new IDs and the old names, new IDs to compile and old names for the files (since the folders still use debug names), this part got weird with the newest changes (and I wasn't really inspired tbh)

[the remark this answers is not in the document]
Dragorn421: yeeaaah I guess this makes assets really weird to handle

For example what happens if I add a custom scene named "ydan", it's partially going to conflict with deku tree?
Well whatever I hope there won't be too many bugs

Dragorn421 (approved):
As-is this leaves room for oddities to happen, like for example having two `SCENE_CUTSCENE_MAP` in the enum (but no problem asset-folders-wise) if for example exporting to a "cutscene_map" scene (`hiral_demo` is the original debug name for the vanilla `SCENE_CUTSCENE_MAP`)



But I don't think it's a big deal, it's easy to catch and fix
If this provides a fix/check for this then cool, if not I'll merge anyway tomorrow or so

Yanis42:
a bit out of the scope of this PR but I fixed the type for cutscene arrays using s32 instead of CutsceneData (people were annoyed by that)

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -103,12 +103,17 @@ def getSceneParams(scene, exportInfo, sceneNames):
...
 
     # if the index is None then this is a custom scene
     if sceneIndex is None and scene is not None:
         sceneName = scene.name.lower() + "_scene"
         sceneTitle = "none"
-        sceneID = "SCENE_" + (scene.name.upper() if scene is not None else exportInfo.name.upper())
+        sceneID = ootSceneNameToID.get(name)
+
+        if sceneID is None and bpy.context.scene.ootSceneExportSettings.option == "Custom":
+            sceneID = f"SCENE_{name.upper()}"
```
Dragorn421, line 115: it's weird to use export settings here? like, what if this function is called for something else like deleting a scene or idk

Yanis42: I wasn't sure how to check that properly tbh

Dragorn421: also what happens if `sceneID` stays None

Dragorn421 (commented):
Custom export to overworld/spot00

image

Then to overworld/hyrule_field

image

-> duplicate scene "spot00", at least both use the "spot00_scene" segment

```diff
diff --git a/include/tables/scene_table.h b/include/tables/scene_table.h
index 0337e1252..75d0877d4 100644
--- a/include/tables/scene_table.h
+++ b/include/tables/scene_table.h
@@ -90,33 +90,34 @@
 /* 0x4E */ DEFINE_SCENE(mahouya_scene, g_pn_24, SCENE_POTION_SHOP_GRANNY, SDC_POTION_SHOP_GRANNY, 0, 0)
 /* 0x4F */ DEFINE_SCENE(ganon_demo_scene, none, SCENE_GANON_BOSS, SDC_OUTSIDE_GANONS_CASTLE, 0, 0)
 /* 0x50 */ DEFINE_SCENE(kinsuta_scene, g_pn_22, SCENE_HOUSE_OF_SKULLTULA, SDC_DEFAULT, 0, 0)
-/* 0x51 */ DEFINE_SCENE(spot00_scene, g_pn_27, SCENE_HYRULE_FIELD, SDC_HYRULE_FIELD, 0, 0)
+/* 0x51 */ DEFINE_SCENE(spot00_scene, none, SCENE_HYRULE_FIELD, SDC_DEFAULT, 0, 0)
+/* 0x52 */ DEFINE_SCENE(spot00_scene, none, SCENE_SPOT00, SDC_DEFAULT, 0, 0)
```

**though** the spec **does** have a new `hyrule_field_scene` segment

Dragorn421:
Again I think 6b5a0ba92c5a6e58e739790aa07fda558b47bdbb and the "oddities" I mentioned in https://github.com/Fast-64/fast64/pull/199#pullrequestreview-1209167938 would be fine, as they're easy to catch, easy to fix and should be rare in practice

I also think it's not worth making the code a lot more complicated just to try handle it better

Yanis42:
I don't even know why someone would want to use the custom option to use a vanilla name lol, I think we should just prevent users from doing that to "fix" these issues but yea whatever lol

I don't think it's worth a fix considering how weird this usage is, I'm fixing the conflict and also a small issue with the importer that came up since the ZAPD update and I think it will be good for now

Yanis42:
everything should be working properly now, the importer has some issues since the introduction of ``uintptr_t`` (and other things brought by HackerOoT)

Dragorn421 (approved):

Yanis42:
before merging this I need to check if the camera enum names are still accurate, it might have broke since the camera pr from decomp

Yanis42:
should be good now 👍

PR #203, opened 2022-12-16
[OoT] Fixed scene table for HackerOoT

Yanis42:
HackerOoT had few issues with the scene table code, this provides a fix that shouldn't affect normal decomp


fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -128,28 +142,35 @@ def sceneTableToC(data, header, sceneNames, scene):
...
     # add the actual lines with the same formatting
+    # add the ifdef if this is HackerOoT
     for i in range(len(data)):
         # adds the "// Debug-only scenes"
         # if both lastScene indexes are the same values this means there's no debug scene
         if ((i - 1) == lastNonDebugSceneIdx) and (lastSceneIdx != lastNonDebugSceneIdx):
+            if isHackerOoT:
+                fileData += "\n#ifdef INCLUDE_TEST_SCENES\n"
+
             fileData += "// Debug-only scenes\n"
```
Dragorn421, line 153: Could you externalize this to a "scene_table_debug_scenes_write_header" (or smth) function?

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -128,28 +142,35 @@ def sceneTableToC(data, header, sceneNames, scene):
...
-            fileData += "// Added scenes\n"
+            fileData += "\n// Added scenes\n"
 
         fileData += f"/* 0x{i:02X} */ DEFINE_SCENE("
         fileData += ", ".join(str(d) for d in data[i])
 
         fileData += ")\n"
 
+        if isHackerOoT and i == lastSceneIdx:
+            fileData += "#endif\n"
```
Dragorn421, line 165: same with `scene_table_debug_scenes_write_footer`

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -128,28 +142,35 @@ def sceneTableToC(data, header, sceneNames, scene):
...
+
     # return the string containing the file data to write
     return fileData
 
 
 def getDrawConfig(sceneName: str):
     """Read draw config from scene table"""
-    fileData, header, sceneNames = getSceneTable(bpy.path.abspath(bpy.context.scene.ootDecompPath))
+    fileData, header, sceneNames, isHackerOoT = getSceneTable(bpy.path.abspath(bpy.context.scene.ootDecompPath))
```
Dragorn421, line 173: okay no we can't have this as a random bool going through everything, we need something better

Dragorn421 (commented):
comments pertaining to keeping the hackeroot-specific fast64 hacks contained

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Yanis42: I'm not sure to understand 🤔

Dragorn421: By "externalize" I mean make a new function, call it here, and move stuff from here to that function

kurethedead (commented):
`isHackerOoT` corresponds to `OOT_Properties.hackerFeaturesEnabled`, so the latter value should be passed down to these functions instead of trying to detect it in code. I agree with Dragorn's comments about creating header/footer functions for HackerOOT specific stuff.

Yanis42:
I changed how it works, now it reads the file and ignore the HackerOoT things (so it basically generate a scene table from normal decomp) then right before writing the file it checks if there's HackerOoT stuff in the original file and it adds the necessary changes and then it writes the file

Also I found a small issue with removing scenes, I'll fix it soon™

Yanis42:
(also about using ``OOT_Properties.hackerFeaturesEnabled``, I think it's better if it's checking on its own, the users might not know they need to enable the button)

Yanis42:
I think the remove scene issue should be fixed now

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -225,7 +257,18 @@ def modifySceneTable(scene, exportInfo: ExportInfo):
...
 
+    # get the new file data
+    newFileData = sceneTableToC(fileData, header, sceneNames, scene)
+
+    # apply HackerOoT changes if needed
+    isHackerOoT = False
+
+    with open(os.path.join(exportPath, "include/tables/scene_table.h"), "r") as file:
+        if '#include "config.h"' in file.read():
+            isHackerOoT = True
```
Dragorn421, line 268: that's not enough for checking HackerOoT, another project could very well have a `#include "config.h"` for whatever reason

Why not just use the "use hackeroot features" property?

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -4,6 +4,13 @@
 from ....oot_utility import getCustomProperty, ExportInfo
 
 
+def getSceneSettingsOption(scene):
```
Dragorn421, line 7: I don't understand this function

mostly "option" is too vague,
and "scene" is confusing between an oot scene and a blender scene

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -158,6 +171,25 @@ def getDrawConfig(sceneName: str):
     raise PluginError(f"Scene name {sceneName} not found in scene table.")
 
 
+def addHackerOoTData(fileData: str):
+    """Reads the file and adds HackerOoT's modifications to the scene table file"""
+    newFileData = '#include "config.h"\n\n'
+
+    for line in fileData.split("\n"):
+        if "// Debug-only scenes" in line:
+            newFileData += "\n#ifdef INCLUDE_TEST_SCENES\n"
```
Dragorn421, line 180: this is very inefficient (performance wise)
when you do `strA + strB`, it copies strA and strB to new memory for the resulting string
it's fine for a few uses, but doing `+=` in big loops like here means the concatenation ends up being of quadratic complexity
instead:
use a `lines = []` list, append each line to it individually, and return `"".join(lines)` or `"\n".join(lines)` which achieves linear complexity

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Yanis42: https://github.com/Fast-64/fast64/pull/203#issuecomment-1382182404

though I'll use that I guess :peepoShrug: if it's just me then I'm fine with using it

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Yanis42: it's to get the proper ``.option`` from either ``ootExportSettings`` or ``ootRemoveSettings`` (idr the exact names), when you remove a scene it's using the export settings instead of the remove settings because it was hardcoded since the other one didn't exist

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Yanis42: so, instead of ``file.read()`` use readlines and then go through the list to do the same thing (except you replace the thing from the list of course)

and return with a str.join?

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Dragorn421: well if the checkbox is there... you should expect users to have it checked

overall fast64 should not contain things that change behavior to accomodate forks of decomp, that enable on their own without user input
otherwise it'd get messy really fast

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Dragorn421: well you could just change `newFileData = []` and `newFileData += s` -> `newFileData.append(s)` `return "".join(newFileData)`

I suggested `lines` because it's easier to grasp but it doesn't need to be one line per string in that list

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Dragorn421: yeah but what is `.option`

Yanis42: it's the scene choice from the panel

Dragorn421: so `getSceneSettingsOption` is like `getSceneSettingsSceneName` ?

Yanis42: I just called it ``getSceneSettingsOption`` because of the name of the things to get the scene value but yeah I guess scene name makes more sense

Yanis42: gonna name it ``getSceneNameSettings`` for "get scene name from settings" it should be better this way

Yanis42:
(I decided to use that PR to fix the wind command lacking a newline, it's just two characters I think it's not worth a seperate PR)

fast64_internal/oot/scene/exporter/to_c/room_commands.py
```diff
@@ -30,7 +30,8 @@ def getTimeSettingsCmd(outRoom: OOTRoom):
 
 def getWindSettingsCmd(outRoom: OOTRoom):
     return (
-        indent + f"SCENE_CMD_WIND_SETTINGS({', '.join(f'{dir}' for dir in outRoom.windVector)}, {outRoom.windStrength})"
+        indent
+        + f"SCENE_CMD_WIND_SETTINGS({', '.join(f'{dir}' for dir in outRoom.windVector)}, {outRoom.windStrength}),\n"
```
Dragorn421, line 34: :thinking: why only this one and not others
are you sure the issue you saw was specifically about the wind cmd

Dragorn421: oh it's inconsistent if the `getXxxCmd` functions return with or without `,\n`

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -15,7 +22,13 @@ def getSceneTable(exportPath):
         with open(os.path.join(exportPath, "include/tables/scene_table.h")) as fileData:
             # keep the relevant data and do some formatting
             for i, line in enumerate(fileData):
-                if not line.startswith("// "):
+                if (
+                    line != "\n"
+                    and '#include "config.h"\n' not in line
+                    and "#ifdef INCLUDE_TEST_SCENES" not in line
+                    and "#endif" not in line
```
Dragorn421, line 29: I didn't realize but this is HackerOoT-specific code in the middle of generic-decomp code

Yanis42: yeah well, I don't really have the choice

fast64_internal/oot/scene/exporter/to_c/room_commands.py
Yanis42: this one and the actor and object one need ``,\n`` since it's not part of a list that's being used in a join, I can always change that and add the thing after the call but idk it feels weird to me

Ticamus (approved):

thecozies (approved):

PR #204, opened 2022-12-19
[OoT] Applied new organisation to cutscene and skeleton

Yanis42:
Applied the new organisation to ``oot_cutscene.py`` and ``oot_skeleton.py``, doing this first because the cutscene implementation will need an update when the cutscene system is documented in decomp (and doing skeleton with it for no reason, I just wanted to do files where I don't have to think hours first), also skeleton has a ``utility.py`` file, that's just to avoid having circular imports

For the changes, it's just split ``oot_cutscene.py`` and ``oot_skeleton.py`` into ``cutscene/`` and ``skeleton/`` folders, I can write a proper changelog if you need it 👍


Dragorn421 (approved):

Dragorn421:
Sorry for retracting my approval, I messed up and have only looked at a fraction of the changes :sweat:

Yanis42:
any update on this?

kurethedead (changes requested):
I think operator code in folders (ex. `cutscene/exporter/cutscene.py`) should be named something like `functions.py`, which can be standardized across all operators (same for skeleton, etc.) Otherwise the organization looks good.

Yanis42:
tbh I wasn't 100% satisfied with the current name, ``functions`` looks a lot better 👍

kurethedead (approved):

PR #206, opened 2022-12-23
Shadows in armateurs

Lilaa3:
you know, like we already do for render areas. who the fuck forgor

reasons this should be merged:
it would be so awesome
doesnt explode (i think)
doesnt need a start bone if you only want shadows
kaze asked for it so atleast one person will use it
changes

reason this shouldnt be merged:
none

### DIFF
--- a/fast64_internal/sm64/sm64_geolayout_parser.py
+++ b/fast64_internal/sm64/sm64_geolayout_parser.py
@@ -1345,7 +1345,7 @@ def parseShadow(
         else:  # Invalid shadow
             shadowType = 0
     shadowSolidity = int.from_bytes(command[4:6], "big")
-    shadowScale = int.from_bytes(command[6:8], "big")
+    shadowScale = int.from_bytes(command[6:8], "big") / bpy.context.scene.blenderToSM64Scale
 
     if not ignoreNode:
         boneName = format(nodeIndex, "03") + "-shadow"
@@ -1354,7 +1354,7 @@ def parseShadow(
             bone = armatureObj.data.bones[boneName]
             bone.shadow_type = str(shadowType)
             bone.shadow_solidity = shadowSolidity / 0xFF
-            bone.shadow_scale = shadowScale / bpy.context.scene.blenderToSM64Scale
+            bone.shadow_scale = shadowScale
     else:
         boneName = None
 
--- a/fast64_internal/sm64/sm64_level_writer.py
+++ b/fast64_internal/sm64/sm64_level_writer.py
@@ -699,7 +699,7 @@ class SM64OptionalFileStatus:
 def exportLevelC(
     obj, transformMatrix, f3dType, isHWv1, levelName, exportDir, savePNG, customExport, levelCameraVolumeName, DLFormat
 ):
-
+    
     fileStatus = SM64OptionalFileStatus()
 
     if customExport:
@@ -761,8 +761,8 @@ def exportLevelC(
 
         # Needs to be done BEFORE collision parsing
         setRooms(child)
-
-        geolayoutGraph, fModel = (
+        
+        geolayoutGraph, fModel =  convertObjectToGeolayout(
             obj,
             transformMatrix,
             f3dType,
@@ -774,8 +774,9 @@ def exportLevelC(
             DLFormat,
             not savePNG,
         )
-        geolayoutGraphC = geolayoutGraph.to_c()
 
+        geolayoutGraphC = geolayoutGraph.to_c()
+        
         # Write geolayout
         geoFile = open(os.path.join(areaDir, "geo.inc.c"), "w", newline="\n")
         geoFile.write(geolayoutGraphC.source)
@@ -1129,7 +1130,7 @@ class SM64_ExportLevel(ObjectDataExporter):
     bl_idname = "object.sm64_export_level"
     bl_label = "Export Level"
     bl_options = {"REGISTER", "UNDO", "PRESET"}
-
+    
     def execute(self, context):
         if context.mode != "OBJECT":
             raise PluginError("Operator can only be used in object mode.")
@@ -1160,6 +1161,7 @@ class SM64_ExportLevel(ObjectDataExporter):
         except Exception as e:
             raisePluginError(self, e)
             return {"CANCELLED"}  # must return a set
+        
         try:
             self.store_object_data()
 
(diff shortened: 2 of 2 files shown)


KazeEmanuar:
i approve

fast64_internal/sm64/sm64_geolayout_bone.py
```diff
@@ -197,6 +197,13 @@ def draw(self, context):
             col.box().label(text="This is in blender units.")
             prop_split(col, obj, "culling_radius", "Culling Radius")
 
+        col.prop(obj, "add_shadow")
+        if obj.add_shadow:
+            col.box().label(text="This is not in blender units.")
```
Dragorn421, line 202: it would be more helpful to describe what the unit actually is then than what it isn't :sunglasses:

Dragorn421 (commented):
I found something to criticize actually yay

jesusyoshi54:
When using this feature to export a shadow in the armature's "Object" panel, and also exporting a shadow with a bone and using a render area, there is a duplicate shadow cmd. Here is an example:
```c
const GeoLayout test_box_geo[] = {
	GEO_CULLING_RADIUS(1000),
	GEO_OPEN_NODE(),
		GEO_SHADOW(1, 255, 100),
		GEO_SHADOW(0, 255, 100),
		GEO_OPEN_NODE(),
			GEO_DISPLAY_LIST(LAYER_OPAQUE, test_box_DL_mesh_layer_1),
		GEO_CLOSE_NODE(),
		GEO_DISPLAY_LIST(LAYER_OPAQUE, test_box_material_revert_render_settings),
	GEO_CLOSE_NODE(),
	GEO_END(),
};
```
This is not proper behavior, there is no reason why the culling radius and the shadow should not be siblings, and then the first bone is after the first open node.

There is also some behavior that should likely be addressed about redundancy with the first bones choice and the armature object's rendering options.

Render area, start, and now shadow are all redundant options when chosen as the root bones option, and there is no reason why such a garbage geo layout should be generated.

Also while you are adding in root geo layout options, billboard would be a good one to have, and a very common choice for a start, rather than forcing it to be chosen as the first bone option.

fast64_internal/sm64/sm64_geolayout_bone.py
```diff
@@ -228,27 +237,38 @@ def draw(self, context):
 
         prop_split(col, obj, "geo_cmd_static", "Geolayout Command")
         drawLayerWarningBox(col, obj, "draw_layer_static")
-        col.prop(obj, "use_render_area")
+
+        renderAndShadowBox = col.box()
```
jesusyoshi54, line 241: I disagree with the use of separate boxes for all of these properties. Properties should be chunked according to their use, and micro splitting each individual feature into its own box looks ugly.
Also selectively adding in the note only after clicking is pretty annoying imo, as it constantly moves the checkbox. I would place all three of these features in a single box, and make the comment permanent.

fast64_internal/sm64/sm64_geolayout_bone.py
```diff
@@ -228,27 +237,38 @@ def draw(self, context):
...
         if obj.use_render_area:
-            renderAreaBox = col.box()
-            renderAreaBox.label(text="This is in blender units.")
-            renderAreaBox.label(text="This only applies if this is the root object of an object geolayout.")
-            renderAreaBox.label(text="For armature geolayouts, see the armature's object properties instead.")
-            prop_split(col, obj, "culling_radius", "Culling Radius")
-        col.prop(obj, "use_render_range")
+            prop_split(renderAreaBox, obj, "culling_radius", "Culling Radius")
+
+        renderRengeBox = renderAndShadowBox.box()
```
jesusyoshi54, line 254: `renderRangeBox` is probably what you meant, but ideally all variables should use lowercase_with_underscores, though specifically UI layout options in blender should be referred to as temporary single word variables. See: https://docs.blender.org/api/current/info_best_practice.html for UI layout naming practices.

KazeEmanuar:
"This is not proper behavior, there is no reason why the culling radius and the shadow should not be siblings, and then the first bone is after the first open node."
Having siblings as a root is invalid behavior, there can be no siblings before the first GEO_OPEN_NODE. and GEO_CULLING_RADIUS only ever works if it is the root.


"Also while you are adding in root geo layout options, billboard would be a good one to have, and a very common choice for a start, rather than forcing it to be chosen as the first bone option."

Billboard bones only apply to the display list they point to, so this is a pointless option as there's no way to link it to any model like this. If anything, billboard should be an option for the "Export Object" instead of the "Export Armature" button - that way it could optionally export GEO_DISPLAY_LIST or GEO_BILLBOARD.

jesusyoshi54:
> "This is not proper behavior, there is no reason why the culling radius and the shadow should not be siblings, and then the first bone is after the first open node." Having siblings as a root is invalid behavior, there can be no siblings before the first GEO_OPEN_NODE. and GEO_CULLING_RADIUS only ever works if it is the root.

Ok my mistake, I didn't know this. Then the ideal behavior would be for cases where you are using a culling radius, and other root type options, would be to remove potential conflicting sibling geos. Potentially the easiest way to do this would be to not export the option if the root bone had the same option, though maybe simply adding in a warning in the object panel UI is sufficient.

> Billboard bones only apply to the display list they point to, so this is a pointless option as there's no way to link it to any model like this.

Billboards will affect all bones below it by changing their matrix state, so that even when there is no linked DL, everything in the model will be billboarded. This can be used to billboard multiple child bones, or the entire model if chosen as the root.
I have actually gone and checked this just now to make sure, and it does work as one would expect.
```c
const GeoLayout test_box_geo[] = {
	GEO_BILLBOARD_WITH_PARAMS(LAYER_OPAQUE, 0, 0, 0),
	GEO_OPEN_NODE(),
			GEO_ANIMATED_PART(LAYER_OPAQUE, 0, 6, 0, test_box_DL_mesh_layer_1),
			GEO_ANIMATED_PART(LAYER_OPAQUE, 0, 6, 0, test_box_DL_2_mesh_layer_1),
			GEO_ANIMATED_PART(LAYER_OPAQUE, 0, 6, 0, test_box_DL_3_mesh_layer_1),
		GEO_DISPLAY_LIST(LAYER_OPAQUE, test_box_material_revert_render_settings),
	GEO_CLOSE_NODE(),
	GEO_END(),
};
```
> If anything, billboard should be an option for the "Export Object" instead of the "Export Armature" button - that way it could optionally export GEO_DISPLAY_LIST or GEO_BILLBOARD.

Billboard already in available for object export in the geo layout command enum located in the mesh object's object panel

PR #207, opened 2022-12-23
[bugfix] SM64 mat override fixes

jesusyoshi54:
This PR fixes an issue with material overrides where excessive reverts were added to display lists with the recent changes to gbiMacro hashing behavior.
This change opts to instead not use sets, and manage the commands in command lists by adding and removing commands via id.


fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -2497,16 +2501,18 @@ def saveOverrideDraw(obj, fModel, material, specificMat, overrideType, fMesh, dr
                         command.displayList = fOverrideMat.material
                     # Check if this material is the same as the prevous loaded material in the DL
                     if prevMaterial == curMaterial:
-                        # If so, tag this material load for removal
-                        removeCommands.add(command)
+                        # If so, remove this material load
+                        remove_by_id(meshMatOverride.commands, command)
+                        commandIdx -= 1
```
Dragorn421, line 2506: `command` is from `command = meshMatOverride.commands[commandIdx]`
so just do
```suggestion
                        # If so, remove this material load
                        meshMatOverride.commands.pop(commandIdx)
                        commandIdx -= 1
```
?
more or less same with others

jesusyoshi54: You need to use the ID. Remove, pop, find etc. all use equality checks, aka `command == meshMatOverride.commands[commandIdx]`. With the way dataclasses work currently, this resolves to true on commands with the same argument, but in different positions in the list. I need to remove the specific command object in that slot from the list, not simply a command that is equal to the current command.
That is why I use `remove_by_id` so that my check for removal uses the `is` keyword, which checks if the objects are the same.

jesusyoshi54: This difference is actually what caused this bug to begin with, otherwise this code would have worked in its original form (though it did produce some extra benign reverts in some instances).

Dragorn421: Yeah I understand that, but why does that make my suggestion not work?

jesusyoshi54: Ok I may have just not read this properly lmfao.
Yeah this would probably work, I'll test it out to make sure.

jesusyoshi54: Ok this definitely doesn't work. I'm not entirely sure yet exactly why it is so broken, but the results in game speak for themselves.
Here is an example of some of the results.
image

jesusyoshi54: Ok I actually did manage to get this change to work.
Idk what was wrong with my testing earlier but it seems fine now.
I will implement this change.

kurethedead (approved):
I played played SM64 from new game to 8 stars to confirm that it seems to work correctly.

PR #208, opened 2023-01-10
[OoT] Cutscene Implementation Update 

Yanis42:
Opening this as a draft so people can make suggestions

Current changes:
- Added more ``layout.box()`` to improve the UI
- Updated the command names
- Added enums where it's needed (cutscene destination, ocarina action, BGM, fade out player and misc)
- Renamed "Cutscene Terminator" to "Cutscene Destination"
- Moved the commands before the list tabs to make the UI easier to use
- Renamed "Cmd 9" to "Rumble Controller" and changed the icon to a wave-like one
- Removed "Cmd Unk" as it's not implemented
- Used proper names for rumble command
- Used more indentation to improve the C output
- Added "custom" option for the enums
- Use "0" as the default value for unused/not implemented arguments
- Use proper names for commands sub-properties
- Used enum names for the music enum
- Probably other things I forgot

Notes:
- Current renames are mostly on the UI (but not everything has been renamed yet), it will be part of the code later
- There's not updater (yet) for older blends
- zcamedit merge won't be a part of this PR

<details>
<summary>Output Example</summary>

```c
CutsceneData TEST[] = {
    CS_BEGIN_CUTSCENE(10, 100),
        CS_DESTINATION(CS_DEST_CUSTOM, 99, 100),
        CS_TEXT_LIST(3),
            CS_TEXT(0x0001, 31, 51, CS_TEXT_CUSTOM, 0x0000, 0x0000),
            CS_TEXT_NONE(188, 206),
            CS_TEXT_OCARINA_ACTION(OCARINA_ACTION_CUSTOM, 311, 447, 0x0045),
        CS_TRANSITION(CS_TRANS_CUSTOM, 429, 490),
        CS_LIGHT_SETTING_LIST(1),
            CS_LIGHT_SETTING(154, 2574, 0, 0, 0, 0, 0, 0, 0, 0, 0),
        CS_TIME_LIST(1),
            CS_TIME(0, 112, 0, 10, 25),
        CS_START_SEQ_LIST(1),
            CS_START_SEQ(NA_BGM_CUSTOM1, 2100, 0, 0, 0, 0, 0, 0, 0, 0, 0),
        CS_STOP_SEQ_LIST(1),
            CS_STOP_SEQ(NA_BGM_CUSTOM, 3247, 0, 0, 0, 0, 0, 0, 0, 0, 0),
        CS_FADE_OUT_SEQ_LIST(1),
            CS_FADE_OUT_SEQ(CS_FADE_OUT_BGM_MAIN, 3277, 3066, 0, 0, 0, 0, 0, 0, 0, 0),
        CS_MISC_LIST(1),
            CS_MISC(CS_MISC_CUSTOM, 2437, 4689, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0),
        CS_RUMBLE_CONTROLLER_LIST(1),
            CS_RUMBLE_CONTROLLER(0, 7084, 0, 0x0054, 0x0024, 0x0032, 0, 0),
    CS_END(),
};
```
</details>

As always I'm open to suggestions!


Yanis42:
New changes:
- renamed the old "subprops" (to explain it better, it's organised like this: 1. list of cs commands then 2. lists of command's data and 3. the actual props to set the data, I renamed those props which are called "subprops" in the code)
- introduced an upgrade system for old blends, I haven't double checked yet but, as they say, it works on my machine™
- other minor changes, like comments and small fixes

Yanis42:
Marking this as ready to review as I don't have anything else to change (I think)

fast64_internal/oot/scene/exporter/to_c/scene_cutscene.py
```diff
@@ -1,144 +1,95 @@
...
+                csData.source += f"0, {e.startFrame}, 0, {e.hour}, {e.minute}"
+
+            elif list.listType == "RumbleList":
+                # same as above
+                csData.source += (
+                    f"0, {e.startFrame}, 0, {e.rumbleSourceStrength}, {e.rumbleDuration}, {e.rumbleDecreaseRate}, 0, 0"
                 )
+
+            elif list.listType in ["StartSeqList", "StopSeqList", "FadeOutSeqList"]:
+                endFrame = e.endFrame if list.listType == "FadeOutSeqList" else "0"
```
mzxrules, line 79: why do you force endFrame to be 0?

Yanis42: in decomp it says the end frame isn't used to the value doesn't matter at all (unless I misread the line, which is possible)

Yanis42: (https://github.com/zeldaret/oot/blob/9f09505d34619883748a7dab05071883281c14fd/include/z64cutscene_commands.h#L197 & https://github.com/zeldaret/oot/blob/9f09505d34619883748a7dab05071883281c14fd/include/z64cutscene_commands.h#L212)

Yanis42: but yea I guess I can use the default value of the class, those need a small cleanup anyway but idk if I should do it in this PR

mzxrules: It should be noted that decomp's implementation of the cutscene structs doesn't match the original implementation. For example, all evidence suggests that `CsCmdMisc`, `CsCmdLightSetting`, `CsCmdStartSeq`, `CsCmdStopSeq` `CsCmdFadeOutSeq` are all the same type as `CsCmdActorCue`. Decomp just implements them separately so that they can hide editor only properties not read by the game, and give a few properties more specialized names.

`endFrame` is one such property that all commands have, but isn't always needed by the game. But I don't think it makes any sense to deliberately force the value to 0 just because you can get away with it.

fig02: lol
obv mzx and I dont agree on this, shown by the discussions in the z_demo pr for this topic. but imo its fine to use 0 when it doesnt matter

fig02: i understand your point of view a little from the perspective of vanilla data.

but for custom cutscenes, how is it corrupt data? you want the commands that dont use an endframe to have one anyway and be confusing for the user? the ui might as well hide it so people dont need to think about it. cause again, it does not matter

Yanis42: So, with your (mzx) logic we should also use 1 for unused args? It doesn't make sense to me, I mean I understand the point of view, like ok I get it originally it wasn't split etc, but for Fast64 in a modding context where I and other devs try our best to get something easy to use for everyone I don't think consistency should be considered for this particular thing as it's just confusing, alternatively I can accept using a macro but idk it doesn't seem really useful, also I don't really understand how 0 is corrupted data, especially for a frame value where you can actually want 0 to be used (well, ok 0 for a end frame is weird but you get what I mean hopefully)

Yanis42: Tldr I just want Fast64 to be easy to use and understand and having a value for this would be confusing for the users

mzxrules: Two simple reasons:

* Storing the `endFrame` for all commands is a nice thing to have, and the structs already support this.
* Given that the structs all call the property `endFrame` in decomp, it makes zero sense to deliberately generate incorrect data for this property, since the user could have their own reasons for reading the data.

fast64_internal/oot/README.md
```diff
@@ -166,8 +166,17 @@ To be able to actually watch your cutscene you need to have a way to trigger it,
...
+
 Note that you can have the actual address of your cutscene if you use ``sym_info.py`` from decomp. Example with ``gHyruleFieldIntroCs``:
 - Command: ``./sym_info.py gHyruleFieldIntroCs``
 - Result: ``Symbol gHyruleFieldIntroCs (RAM: 0x02013AA0, ROM: 0x27E9AA0, build/assets/scenes/overworld/spot00/spot00_scene.o)``
 
 If you have a softlock in-game then you probably did something wrong when creating the cutscene. Make sure you set up the bones properly. The softlock means the game is playing a cutscene but it's probably reading wrong data. Make sure the cutscene is exported, if it's not export it again.
+
+If the game crash check the transitions if you use the transition command (check both the ones from the entrance table and your cutscene script), also it will crash if you try to use the map select without having a 5th entrance (or more depending on the number of cutscenes you have) in the group for your scene.
```
Dragorn421, line 182:
```suggestion
If the game crashes check the transitions if you use the transition command (check both the ones from the entrance table and your cutscene script), also it will crash if you try to use the map select without having a 5th entrance (or more depending on the number of cutscenes you have) in the group for your scene.
```

fast64_internal/oot/oot_constants.py
```diff
@@ -190,113 +190,122 @@
...
-    ("0x64", "Ganondorf Battle", "Ganondorf Battle"),
-    ("0x65", "Ganon Battle", "Ganon Battle"),
-    ("0x66", "Seal of Six Sages", "Seal of Six Sages"),
-    ("0x67", "End Credits I", "End Credits I"),
-    ("0x68", "End Credits II", "End Credits II"),
-    ("0x69", "End Credits III", "End Credits III"),
-    ("0x6A", "End Credits IV", "End Credits IV"),
-    ("0x6B", "King Dodongo & Volvagia Boss Battle", "King Dodongo & Volvagia Boss Battle"),
-    ("0x6C", "Mini-Game", "Mini-Game"),
+    ("NA_BGM_GENERAL_SFX", "General Sound Effects", "General Sound Effects"),
```
Dragorn421, line 193: could you quote a decomp link here too like you did elsewhere

fast64_internal/oot/oot_upgrade.py
```diff
@@ -35,3 +47,167 @@ def upgradeRoomHeaders(roomObj: Object, objData: OoT_ObjectData):
...
+
+                if newName in ["rumbleSourceStrength", "rumbleDuration", "rumbleDecreaseRate"]:
+                    value = int(getEvalParams(data[oldName]), base=16)
+
+                data[newName] = value
+
+            del data[oldName]
+
+
+def convertOldDataToEnumData(data, oldDataToEnumData: list[tuple[str, str, list[tuple[str, str, str]]]]):
```
Dragorn421, line 72: `list[tuple[str, str, list[tuple[str, str, str]]]]`
monkaW
what about one or two dataclasses to replace some tuples?

fast64_internal/oot/oot_upgrade.py
```diff
@@ -35,3 +47,167 @@ def upgradeRoomHeaders(roomObj: Object, objData: OoT_ObjectData):
...
+                    # else raise an error to default to custom
+                    raise IndexError
+            except:
+                setattr(data, newName, "Custom")
+                setattr(data, f"{newName}Custom", str(oldData))
+
+                # @TODO: find a way to check properly which seq command it is
+                if newName == "csSeqID":
+                    setattr(data, "csSeqPlayer", "Custom")
+                    setattr(data, "csSeqPlayerCustom", str(oldData))
```
Dragorn421, line 132: can't say I understand the todo

like, the seqId value here may be `seqPlayer` from `CS_FADE_OUT_SEQ` ?

fast64_internal/oot/scene/exporter/to_c/scene_cutscene.py
```diff
@@ -1,144 +1,95 @@
...
-            "\tCS_TERMINATOR("
-            + str(csParent.csTermIdx)
-            + ", "
-            + str(csParent.csTermStart)
-            + ", "
-            + str(csParent.csTermEnd)
-            + "),\n"
+def ootCutsceneDataToC(csParent: OOTCutscene | OOTScene, csName: str):
+    csData = CData()
+    arrayName = f"CutsceneData {csName}[]"
```
Dragorn421, line 9: imo the name of the array is just `csName`, I would call this `declarationBase` since it's used for the extern declaration and the definition (which is also a declaration)

(declaration: "this thing exists"
definition: "this thing is this")

fast64_internal/oot/scene/exporter/to_c/scene_cutscene.py
```diff
@@ -1,144 +1,95 @@
...
-                    + e.unk11
-                    + ", "
-                    + e.unk12
+                    csData.source += f"{e.startFrame}, {e.endFrame}"
+
+                elif e.textboxType == "OcarinaAction":
+                    csData.source += f"{e.ocarinaAction}, {e.startFrame}, {e.endFrame}, {e.ocarinaMessageId}"
+
+            elif list.listType == "LightSettingsList":
+                # the endFrame variable is not used in the implementation of the commend
```
Dragorn421, line 64:
```suggestion
                # the endFrame variable is not used in the implementation of the command
```

fast64_internal/oot/scene/properties.py
```diff
@@ -270,10 +270,9 @@ class OOTSceneHeaderProperty(PropertyGroup):
     # not been removed as doing so would break any existing scenes made with this
     # type of cutscene data.
     csEndFrame: IntProperty(name="End Frame", min=0, default=100)
-    csWriteTerminator: BoolProperty(name="Write Terminator (Code Execution)")
-    csTermIdx: IntProperty(name="Index", min=0)
-    csTermStart: IntProperty(name="Start Frm", min=0, default=99)
-    csTermEnd: IntProperty(name="End Frm", min=0, default=100)
+    csUseDestination: BoolProperty(name="Write Terminator (Code Execution)")
+    csDestination: IntProperty(name="Index", min=0)
+    csDestinationStartFrame: IntProperty(name="Start Frm", min=0, default=99)
```
Dragorn421, line 275: does frm need to be abbreviated?

Dragorn421 (commented):
skimmed through

fast64_internal/oot/oot_upgrade.py
Yanis42: I should rephrase that but we need a proper way to check if we're dealing with the fade out command or the start/stop one, there's a minor difference between these two, the fade out one doesn't take a seq id but the player type (as defined in sequence.h iirc)

all of this is based on the informations inside z64cutscene_commands.h btw

the main issue at the moment is that we have to use "Custom" when upgrading because everything was assumed to have the same parameters, and now we know for sure this is a wrong assumption

fast64_internal/oot/scene/exporter/to_c/scene_cutscene.py
Yanis42: names aren't my strongest point ngl lol

I'll change other names like that too to stay consistent 👍

fast64_internal/oot/scene/properties.py
Yanis42: that's the deprecated code, I haven't updated it tbh I don't even know if everything still works for that

imo we should completely remove that legacy system and upgrade the things (though I don't have a cs in a scene to I can't really try if it'd work properly)

fast64_internal/oot/oot_upgrade.py
Yanis42: I hope the changes I made is what you had in mind

Yanis42:
any updates on this? sorry I just need this to be merged to make things easier with merging zcamedit

Yanis42:
is there anything holding this? except update that branch I guess

fast64_internal/oot/scene/properties.py
Yanis42: should we just remove this thing? I'm not sure anyone still using that now idk if it's worth making the code to update properly

Yanis42:
most of things I wanted to do are done, I improved the importer and the exporter and fixed some issues

I just want to improve the UI and I think I'm done with this, also I want to remove support for embedded cutscenes, I don't think anyone is using it, and if I do that I'll move files under ``motion/exporter``/``motion/importer`` as this is handling everything now (with the necessary class/other rename)

sauraen (changes requested):
Some nits, otherwise looks good.

Tested importing a vanilla scene and previewing the cutscene in Blender. Also tested exporting a basic scene with several different command types. Both worked with no issues.

fast64_internal/oot/cutscene/classes.py
```diff
@@ -0,0 +1,597 @@
...
+    pos: tuple[int, int, int] = field(default_factory=tuple)
+    paramNumber: int = 8
+
+    def __post_init__(self):
+        if self.params is not None:
+            self.continueFlag = self.params[0]
+            self.camRoll = getInteger(self.params[1])
+            self.frame = getInteger(self.params[2])
+            self.viewAngle = float(self.params[3][:-1])
+            self.pos = [getInteger(self.params[4]), getInteger(self.params[5]), getInteger(self.params[6])]
```
sauraen, line 51: Sorry, one more thing! You changed the types to tuples, but what you're actually storing in it is a list. I don't know what code accesses this, but if it needs to be a list rather than a tuple then the type should be `list[int]`. Otherwise change the brackets on line 51 to parentheses. (Same for the other places in this file)

sauraen (changes requested):

sauraen (approved):
Thanks!

PR #209, opened 2023-01-10
[OOT] Actor Collider Import/Export

kurethedead:
This PR allows for importing/exporting actor colliders as a part of the skeleton/DL import/export process. The bulk of this was done a while ago but I never got around to finishing it, mainly because I didn't have the time to figure out how to test exports more extensively with custom actors. There's a couple things I'd like some feedback on:

1. Naming conventions for some of the physics properties (AC, AT, OC), also general UI layout for these properties.
2. The current setup for joint sphere colliders uses the skeleton to store shared collision properties, so only one joint sphere group is allowed per skeleton. I'm not sure if there's a better solution to this other than providing options to import specific named colliders. Right now only the first joint sphere group is imported.
3. I'm not sure how Ganon2's joint sphere colliders work, to me it seems kind of buggy but the code otherwise seems to work on other actors. 

There's some other minor issues I added to the README but those can be handled manually by the user IMO.

### DIFF
--- a/fast64_internal/f3d/f3d_parser.py
+++ b/fast64_internal/f3d/f3d_parser.py
@@ -2156,15 +2156,6 @@ def parseMacroArgs(data):
     return params
 
 
-def getImportData(filepaths):
-    data = ""
-    for path in filepaths:
-        if os.path.exists(path):
-            data += readFile(path)
-
-    return data
-
-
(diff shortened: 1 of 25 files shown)


MNGoldenEagle:
Could we add a checkbox that enables exporting the vertices for a quad collider for custom actors? It probably isn't needed for most actors, but there could be some custom actors that want to use the collider vertices instead.

PR #211, opened 2023-01-13
[OoT] Fixed 4 issues

Yanis42:
Found and fixed 3 issues:
- Fast64 was still using ``LightSettings`` instead of the new type ``EnvLightSettings``, it was fine for normal decomp but on HackerOoT this was an issue as the old type has been removed
- Removed an extra argument inside the ``draw_props()`` function call from ``OOTCutsceneProperty.draw_props``, this was raising an error that didn't show the textbox UI elements for cutscenes
- Added missing ``_scene_`` chars to the filename exported in the spec file for cutscene files

### DIFF
--- a/fast64_internal/oot/scene/exporter/to_c/spec.py
+++ b/fast64_internal/oot/scene/exporter/to_c/spec.py
@@ -101,7 +101,7 @@ def editSpecFile(scene: OOTScene, exportInfo: ExportInfo, sceneC: OOTSceneC):
 
                 if sceneC.sceneCutscenesIsUsed():
                     for i in range(len(sceneC.sceneCutscenesC)):
-                        sceneSegInclude += indent + f'include "{includeDir}_cs_{i}.o"\n'
+                        sceneSegInclude += indent + f'include "{includeDir}_scene_cs_{i}.o"\n'
 
             sceneSegInclude += indent + "number 2\n"
             specEntries.insert(firstIndex, sceneSegInclude)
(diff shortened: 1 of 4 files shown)


fast64_internal/oot/scene/exporter/to_c/spec.py
```diff
@@ -101,7 +101,7 @@ def editSpecFile(scene: OOTScene, exportInfo: ExportInfo, sceneC: OOTSceneC):
 
                 if sceneC.sceneCutscenesIsUsed():
                     for i in range(len(sceneC.sceneCutscenesC)):
-                        sceneSegInclude += indent + f'include "{includeDir}_cs_{i}.o"\n'
+                        sceneSegInclude += indent + f'include "{includeDir}_scene_cs_{i}.o"\n'
```
Dragorn421, line 104: that's weird, it looks like currently it should be correct:
```c
# ootExportSceneToC
os.path.join(levelPath, scene.sceneName() + "_cs_" + str(i) + ".c"),

# editSpecFile
includeDir = f"build/{getSceneDirFromLevelName(sceneName)}/{sceneName}"
...
sceneSegInclude += indent + f'include "{includeDir}_cs_{i}.o"\n'
```

I guess `scene.sceneName()` has `_scene` and `sceneName` doesn't? that's confusing :(

Dragorn421: OK yeah this isn't marvelous code:

```py
# OOTScene
    def sceneName(self):
        return self.name + "_scene"
```

Yanis42: to explain that better, the file is called ``NAME_scene_cs_INDEX.c`` but the spec entry is called ``NAME_cs_INDEX.c``, so you get a file not found error when you compile the game

I used "_scene_" instead of the function as it's consistent with the other filenames earlier in the code

Dragorn421 (approved):

Yanis42:
(we found another type issue, this time with ``CamData``, though it's fixed now)

PR #213, opened 2023-01-27
Fixed save_textures() error when "Save Textures As PNGs" is enabled

kurethedead:


Agent-11:
I don't have write access but I personally approve of this PR and believe to be beneficial to Fast64's existence. 👍

thecozies (approved):

PR #217, opened 2023-02-19
Mat Bleed Geolayout Bugfix

jesusyoshi54:
This time for sure.
Fixes an issue with exporting geo layouts when the area object does not exist.


jesusyoshi54:
it was not this time for sure

PR #219, opened 2023-02-20
[f3d] Material tooltips

jesusyoshi54:
Adds tooltips to the f3d material GUI for RDP settings. Tooltips are provided by the third argument in an enum item, or by the description key word argument to a property.
Tooltips can be viewed by hovering the mouse over the option in the GUI. The enums tooltips can be viewed by opening the enum and hovering the mouse over the given enum.

### DIFF
--- a/fast64_internal/f3d/f3d_enums.py
+++ b/fast64_internal/f3d/f3d_enums.py
@@ -165,20 +165,20 @@ enumRGBDither = [
 ]
 
 enumCombKey = [
-    ("G_CK_NONE", "None", "None"),
-    ("G_CK_KEY", "Key", "Key"),
+    ("G_CK_NONE", "None", "Disables chroma key."),
+    ("G_CK_KEY", "Key", "Enables chroma key."),
 ]
 
 enumTextConv = [
-    ("G_TC_CONV", "Convert", "Convert"),
-    ("G_TC_FILTCONV", "Filter And Convert", "Filter And Convert"),
-    ("G_TC_FILT", "Filter", "Filter"),
+    ("G_TC_CONV", "Convert", "Convert, used for YUV to RGB conversion."),
+    ("G_TC_FILTCONV", "Filter And Convert", "Filter And Convert, used for YUV to RGB conversion."),
+    ("G_TC_FILT", "Filter", "Filter, used for default textures."),
 ]
 
 enumTextFilt = [
-    ("G_TF_POINT", "Point", "Point"),
-    ("G_TF_AVERAGE", "Average", "Average"),
-    ("G_TF_BILERP", "Bilinear", "Bilinear"),
+    ("G_TF_POINT", "Point", "Point filtering."),
+    ("G_TF_AVERAGE", "Average", "Average filter, not recommended except for pixel aligned texrects."),
+    ("G_TF_BILERP", "Bilinear", "Bilinear, standard N64 filtering with 3 point sample."),
 ]
 
 enumTextLUT = [
@@ -188,14 +188,14 @@ enumTextLUT = [
 ]
 
 enumTextLOD = [
-    ("G_TL_TILE", "Tile", "Tile"),
-    ("G_TL_LOD", "LOD", "LOD"),
+    ("G_TL_TILE", "Tile", "Shows selected color combiner tiles"),
+    ("G_TL_LOD", "LoD", "Enables LoD calculations"),
 ]
 
 enumTextDetail = [
-    ("G_TD_CLAMP", "Clamp", "Clamp"),
-    ("G_TD_SHARPEN", "Sharpen", "Sharpen"),
-    ("G_TD_DETAIL", "Detail", "Detail"),
+    ("G_TD_CLAMP", "Clamp", "Clamp, shows base tile for texel0 and texel 1 when magnifying (>1 texel/pixel)"),
+    ("G_TD_SHARPEN", "Sharpen", "Sharpen, sharpens pixel colors when magnifying (>1 texel/pixel)"),
+    ("G_TD_DETAIL", "Detail", "Detail, shows base tile when magnifying (>1 texel/pixel), else shows base tile+1"),
 ]
 
 enumTextPersp = [
@@ -206,8 +206,8 @@ enumTextPersp = [
 enumCycleType = [
     ("G_CYC_1CYCLE", "1 Cycle", "1 Cycle"),
     ("G_CYC_2CYCLE", "2 Cycle", "2 Cycle"),
-    ("G_CYC_COPY", "Copy", "Copy"),
-    ("G_CYC_FILL", "Fill", "Fill"),
+    ("G_CYC_COPY", "Copy", "Copies texture values to framebuffer with no perspective correction or blending"),
+    ("G_CYC_FILL", "Fill", "Uses blend color to fill primitve"),
 ]
 
 enumColorDither = [("G_CD_DISABLE", "Disable", "Disable"), ("G_CD_ENABLE", "Enable", "Enable")]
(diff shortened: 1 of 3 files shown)


fast64_internal/f3d/f3d_enums.py
```diff
@@ -165,20 +165,20 @@
...
-    ("G_CK_KEY", "Key", "Key"),
+    ("G_CK_NONE", "None", "Disables chroma key."),
+    ("G_CK_KEY", "Key", "Enables chroma key."),
 ]
 
 enumTextConv = [
-    ("G_TC_CONV", "Convert", "Convert"),
-    ("G_TC_FILTCONV", "Filter And Convert", "Filter And Convert"),
-    ("G_TC_FILT", "Filter", "Filter"),
+    ("G_TC_CONV", "Convert", "Convert, used for YUV to RGB conversion."),
```
sauraen, line 173: I would suggest not repeating the name of the feature in the description for any of these.

fast64_internal/f3d/f3d_enums.py
```diff
@@ -165,20 +165,20 @@
...
+    ("G_TC_FILTCONV", "Filter And Convert", "Filter And Convert, used for YUV to RGB conversion."),
```
sauraen, line 174: I'm pretty sure this is filter in the first cycle and convert in the second cycle, but you might want to double check.

fast64_internal/f3d/f3d_enums.py
```diff
@@ -219,20 +219,20 @@
...
-    ("G_AC_DITHER", "Dither", "Dither"),
+    ("G_AC_THRESHOLD", "Threshold", "Threshold, writes if alpha is greater than blend color alpha"),
+    ("G_AC_DITHER", "Dither", "Dither, writes if alpha is greater than random value"),
 ]
 
 enumDepthSource = [
-    ("G_ZS_PIXEL", "Pixel", "Pixel"),
-    ("G_ZS_PRIM", "Primitive", "Primitive"),
+    ("G_ZS_PIXEL", "Pixel", "Z value is calculated per primitive pixel"),
+    ("G_ZS_PRIM", "Primitive", "Primitive, use prim depth to set Z value"),
```
sauraen, line 228: I might add a comment that this is not compatible with HLE emulators.

fast64_internal/f3d/f3d_material.py
```diff
@@ -2460,45 +2460,54 @@ class RDPSettings(bpy.types.PropertyGroup):
         name="Z Buffer",
         default=True,
         update=update_node_values_with_preset,
+        description="Turns on/off Z-Buffer.  Z-Buffer set to 0 if disabled."
```
sauraen, line 2463: This is not correct. Not only does this not set the Z buffer itself to zero as is claimed, but it does not set the Z screen coordinates of the triangle to zero--they end up being garbage data in the RDP. What this does is selects whether Z coordinates are computed on the RSP and sent to the RDP or not; it skips sending them if not. I'd suggest something like "Compute and send Z coordinates of tris. Disable if not reading or writing the Z buffer in the blender."

fast64_internal/f3d/f3d_material.py
```diff
@@ -2460,45 +2460,54 @@ class RDPSettings(bpy.types.PropertyGroup):
...
     )
     g_shade: bpy.props.BoolProperty(
         name="Shading",
         default=True,
         update=update_node_values_with_preset,
+        description="Turns on/off shading. Shade register set to 0 if disabled."
```
sauraen, line 2469: Same deal as with Z, I'd suggest something like "Compute and send shade coordinates of tris. Disable if not using vertex colors, lighting, or fog."

fast64_internal/f3d/f3d_material.py
```diff
@@ -2460,45 +2460,54 @@ class RDPSettings(bpy.types.PropertyGroup):
...
     )
     # v1/2 difference
     g_cull_front: bpy.props.BoolProperty(
         name="Cull Front",
         update=update_node_values_with_preset,
+        description="Turns on/off drawing of front faces"
```
sauraen, line 2475: This may be misleading as enabling culling disables drawing front faces. I'd change the description to "Disables drawing of front faces"

fast64_internal/f3d/f3d_material.py
```diff
@@ -2460,45 +2460,54 @@ class RDPSettings(bpy.types.PropertyGroup):
...
     )
     # v1/2 difference
     g_cull_back: bpy.props.BoolProperty(
         name="Cull Back",
         default=True,
         update=update_node_values_with_preset,
+        description="Turns on/off drawing of back faces"
```
sauraen, line 2482: Same as front

fast64_internal/f3d/f3d_material.py
```diff
@@ -2460,45 +2460,54 @@ class RDPSettings(bpy.types.PropertyGroup):
...
     g_lighting: bpy.props.BoolProperty(
         name="Lighting",
         default=True,
         update=update_node_values_with_preset,
+        description="Enables calculation of shade values from lights and vertex normals. Turn off for vertex colors"
     )
     g_tex_gen: bpy.props.BoolProperty(
         name="Texture UV Generate",
         update=update_node_values_with_preset,
+        description="Generates texture coordinates that maps from norm x/y to [0-1]x/y"
```
sauraen, line 2498: This is technically accurate but isn't going to give any information to someone who doesn't already know what this does. I'd suggest something like "Generates texture coordinates for reflective objects, based on vertex normals and lookat directions. In this version, the sky is at the center of your texture image and the ground is a circle inscribed in the border."

fast64_internal/f3d/f3d_material.py
```diff
@@ -2460,45 +2460,54 @@ class RDPSettings(bpy.types.PropertyGroup):
...
     )
     g_tex_gen_linear: bpy.props.BoolProperty(
         name="Texture UV Generate Linear",
         update=update_node_values_with_preset,
+        description="Generates texture coordinates that linearly maps from cos/sin(norm) to [0-1]x/y"
```
sauraen, line 2503: Same as with texgen: "Modifies the texgen mapping; enable with texgen. Use a normal panorama image for the texture, with the sky at the top and the ground at the bottom."

fast64_internal/f3d/f3d_material.py
```diff
@@ -2513,42 +2522,49 @@ class RDPSettings(bpy.types.PropertyGroup):
         items=enumAlphaDither,
         default="G_AD_NOISE",
         update=update_node_values_with_preset,
+        description="Applies your choice dithering type to output frambuffer alpha"
```
sauraen, line 2525: Typo "frambuffer", also I might explain that dithering is how to convert higher precision values to the lower precision for the framebuffer.

fast64_internal/f3d/f3d_material.py
```diff
@@ -2513,42 +2522,49 @@ class RDPSettings(bpy.types.PropertyGroup):
...
     )
     # v2 only
     g_mdsft_rgb_dither: bpy.props.EnumProperty(
         name="RGB Dither",
         items=enumRGBDither,
         default="G_CD_MAGICSQ",
         update=update_node_values_with_preset,
+        description="Applies your choice dithering type to output frambuffer color"
```
sauraen, line 2533: Same as alpha dither

fast64_internal/f3d/f3d_material.py
```diff
@@ -2513,42 +2522,49 @@ class RDPSettings(bpy.types.PropertyGroup):
...
         default="G_CK_NONE",
         update=update_node_values_with_preset,
+        description="Turns on/off the chroma key. Chroma key requires a special setup to work properly"
     )
     g_mdsft_textconv: bpy.props.EnumProperty(
         name="Texture Convert",
         items=enumTextConv,
         default="G_TC_FILT",
         update=update_node_values_with_preset,
+        description="Turns on/off the converter for texture color. Can convert color space from YUV into RGB"
```
sauraen, line 2547: This is not very useful and bordering on incorrect. I'd suggest "Sets the function of the texture convert unit, to do texture filtering, YUV to RGB conversion, or both"

fast64_internal/f3d/f3d_material.py
```diff
@@ -2562,31 +2578,36 @@ class RDPSettings(bpy.types.PropertyGroup):
...
         default="G_CD_ENABLE",
         update=update_node_values_with_preset,
+        description="Applies your choice dithering type to output frambuffer"
     )
     g_mdsft_pipeline: bpy.props.EnumProperty(
         name="Pipeline Span Buffer Coherency",
         items=enumPipelineMode,
         default="G_PM_1PRIMITIVE",
         update=update_node_values_with_preset,
+        description="Changes primitive rasterization timing. For games besides SM64, N-prim is optimal"
```
sauraen, line 2610: "If set to 1 primitive, effectively inserts a pipe sync after every triangle, substantially worsening the performance. Vanilla SM64 has synchronization bugs which may crash if this is not used. In a modern SM64 project with these bugs fixed, or in any other game like OoT, always use N primitive."

fast64_internal/f3d/f3d_material.py
```diff
@@ -2639,37 +2662,47 @@ class RDPSettings(bpy.types.PropertyGroup):
...
         update=update_node_values_with_preset,
+        description="Multiply coverage (amount primitive covers target pixel) with alpha and store result as coverage"
     )
     alpha_cvg_sel: bpy.props.BoolProperty(
         update=update_node_values_with_preset,
+        description="Use coverage (amount primitive covers target pixel) as alpha instead of color combiner alpha"
     )
     force_bl: bpy.props.BoolProperty(
         update=update_node_values_with_preset,
+        description="Always uses blending on. Default blending is conditionally based on depth. Always used when Z Buffering is off"
```
sauraen, line 2705: Default blending is conditionally based on coverage, not depth--if the pixel is fully covered, it doesn't blend, it just writes the new value. More importantly, however, `force_bl` disables the division step in the blender, meaning B must be equal to 1-A or there may be problems.

sauraen (changes requested):

fast64_internal/f3d/f3d_enums.py
jesusyoshi54: This is correct, I have verified it using the RDP command summary.

jesusyoshi54:
By recommendation of cozies, I will also be changing the tooltip/name for the material bleed option. If current changes are SAT, please wait until I make that update before approving.

sauraen (approved):

PR #221, opened 2023-04-12
fixed upgrading certain V1 materials

thecozies:


sauraen (approved):
Have not tested this (don't have any V1 materials!) but the code looks fine.

PR #223, opened 2023-04-16
Mat bleed and SM64 Geo Export Optimizations

jesusyoshi54:
First change I made some changes to the f3d mat writing methodology to correct othermodeH flags in the mat write method write all. This was an simple error that had to be amended for the new bleed logic to work properly.

Next I made it so that when using the mat write method, write all, materials are not reverted for the function `exportF3DCommon`. Realistically this only affects the f3d DL export operator with the chosen mat write method and bleed exports with the sm64 DL export operator.

For sm64 geo layouts, I made a change so that geometry layout reverts are appended to the end of the last DL of each layer instead of appended to the end of the geo layout. This should reduce graph node overhead basically for free.

As far as bleed changes go, the basic principle is that I get world defaults and then add or remove reverts based on what those world defaults are. I also use these reverts to better make decisions on whether or not to bleed a cmd.
I also made some changes to the bleed logic so that "static" cmds are taken out from any potential bleed processing before I start bleeding and added back in at the end. This basically affects sm64 exports which add in extra unnecessary cmds that are handled by the new bleed reset logic.

Overall bleed exports should properly reset other modes, render modes if the default render mode is supplied (so anything with a sm64 layer), and geo modes. It is expected that all other types of f3d setting would be fine if left to bled through, if this proves to be an issue I can amend it.


aglab2:
I tried it on Serene Fusion C1. Results are not great sadly, I assume some CI4 stuff is bleeding too hard
image

jesusyoshi54:
Could you share either the models or code?
For example a test branch would be great. To me it seems like the issue is with palettes being enabled on non palette textures, but I would need to see why that is.

jesusyoshi54:
I believe my latest commit should fix this issue.
I made the mistake of not keeping track of cmds to reset between materials, so it was only resetting the cmds from the last used material.
To fix this, I made the reset cmd dictionary a variable for the bled fmesh rather than a bleed gfx list prop, so that it is always resetting the last set of used cmds.
image

jesusyoshi54:
Based on requests from some people, the following features will be added before merging this PR.
1: Add the option to isolate a model's bleeding per mesh, instead of bleeding between entire geo layouts. This will make it so that if you have plans to edit the geo layout yourself, or make reference to a geo node via some code, you can make sure that the node will draw correctly with full independence.
2: Make room exports always bleed per mesh with independence, instead of over the whole geo. This does not cause an issue with current room implementation, but it does with the shared room changes, so I will make the change now so that all room exports will work for the foreseeable future

fast64_internal/sm64/sm64_geolayout_classes.py
```diff
@@ -467,7 +468,7 @@ def walk(node, last_materials):
                 cmd_list = fMesh.drawMatOverrides.get(base_node.override_hash, None) or fMesh.draw
                 lastMat = last_materials.get(base_node.drawLayer, None)
                 default_render_mode = fModel.getRenderMode(base_node.drawLayer)
-                lastMat = self.bleed_fmesh(fModel.f3d, fMesh, lastMat, cmd_list, default_render_mode)
+                lastMat = self.bleed_fmesh(fModel.f3d, fMesh, lastMat if not base_node.bleed_independently else None, cmd_list, default_render_mode)
```
Yanis42, line 471: it's hard to follow which type ``base_node`` can be there but I'm concerned about it being ``None``, in this case I'd say I'm not ok with using a bool-like condition when the value is None but if you're sure it's never that value and it's true/false then it's fine

Yanis42 (commented):
OoT exports correctly (but I don't think this PR changes this anyway), I have few comments on the code but otherwise as an OoT person it's ok to me

Things I didn't comment about since it's mostly a me-issue:
- the usage of ``while``, I don't really like using it since you can easily fall under infinite loops
- things like ``type(X) == Y``, those could be replaced by ``isinstance(Y, X)`` but this is fine

Otherwise I don't have anything else to say as I lack knowledge in F3D stuff but it seems to work properly from what I can tell

fast64_internal/sm64/sm64_geolayout_classes.py
jesusyoshi54: `base_node` is never `None`.
The construction of geo layout nodes starts with a `TransformNode` class, which holds an attribute `node` always. If `node` is none then the geo layout node is invalid because it has no data.
As for what `base_node` is, it is a subclass of `BaseDisplayListNode` because those are the classes that have a FMesh defined.

fast64_internal/f3d/f3d_bleed.py
```diff
@@ -40,9 +45,73 @@ class BleedGraphics:
...
+
+        self.default_load_geo = SPLoadGeometryMode(setGeo.flagList)
+        self.default_set_geo = setGeo
+        self.default_clear_geo = clearGeo
+
+    def build_default_othermodes(self):
+        is_hardware_v1 = bpy.context.scene.isHWv1
+        defaults = bpy.context.scene.world.rdp_defaults
+
+        othermode_H = SPSetOtherMode("G_SETOTHERMODE_H", 4, 20 - self.is_f3d_old, [])
```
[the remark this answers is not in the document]
jesusyoshi54: I'm aware of this. The way I wrote it is how it is done for all other cases of `SPSetOtherMode` and I didn't bother to change it. I guess I could change this but it seemed unrelated to the PR overall.

fast64_internal/sm64/sm64_geolayout_classes.py
Yanis42: I was talking about ``bleed_independently`` which gets initialised as None (my bad I should have precised that earlier)

jesusyoshi54: This should be valid code if `bleed_independently` is `None` as `None` resolves to False in any test of truthyness.
I can make it init as `False` without really changing things if that makes more sense.

fast64_internal/f3d/f3d_bleed.py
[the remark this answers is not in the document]
Yanis42: it's fine tbh, I said that just in case you didn't know about the field initialiser but if you know about it then I'm cool with it
I guess that would be part of a F3D cleanup

fast64_internal/sm64/sm64_geolayout_classes.py
Yanis42: I commented on that because I remembered reading something Dragorn said in codebase todo (https://discord.com/channels/874816709855440926/887690239911526431/888901683730595891)

I think that would be the case here

jesusyoshi54: I do believe it is always explicitly set but I'll change it to `False` anyway since that is more inline with the expected value.

Lilaa3:
Simple mario export on 3.6 blender produces wrong results when the open hand model switch option is used, this was tested on paraLLEl.
image
This happens because the hand switch option model is not setting its light (shade color)
image

jesusyoshi54:
Latest commit should address switch nodes/function nodes. I verified mats were written on all first mesh children of switch nodes and that it appeared properly in game, though on my test setup it didn't really appear wrong at all previously so that isn't worth that much.

12abz:
[Screencast from 12-31-2023 02:50:05 PM.webm](https://github.com/Fast-64/fast64/assets/109289447/5899828f-a4d7-4981-b392-a6cf214363a3)
It Works! And If Your Wondering Why The Lighting Is Fucked, Is Because Of F3DEX3.

12abz:
Screenshot from 2023-12-31 14-58-49
And Mario's Switch Hands Work.

KazeEmanuar:
Tested and confirmed the visuals look correct on my level. Unfortunately, the performance is worse by a large margin.

12abz:
bleed mario, not the level.

Dragorn421 (approved):
OoT seems not broken :tm: 

About the above comment

> the performance is worse by a large margin

I was told it was a ~~skill~~ config issue

KazeEmanuar:
> OoT seems not broken ™️
> 
> About the above comment
> 
> > the performance is worse by a large margin
> 
> I was told it was a ~skill~ config issue

It's arguably a config issue or an issue in fast64. With mat bleed on, fast64 will insert SetOtherMode commands that reset display lists back to 1PRIM. This will cause the RSP to take a lot longer and is what I had happen on my end. I believe HackerSM64 also uses NPRIM by default, so this will make any WIP project slower unless the user is aware of this quirk and changes all his materials and his world settings explicitly to NPRIM over the default options.

jesusyoshi54:
One day a hacker64 preset option will come out that will forever banish 1-prim from existence. Until then I will ask the good members of the sm64 community to remind all console hackers to set it.

Lilaa3 (approved):
OOT isn't broken (according to dragon, I don´t do oot so I was just waiting on confirmation that it didn´t break anything) and no issues seem to remain.

Dragorn421:
Going to hijack this to save notes on 1PRIM `G_PM_1PRIMITIVE` vs NPRIM `G_PM_NPRIMITIVE`

Relevant manual parts:

- gDPPipelineMode http://n64devkit.square7.ch/n64man/gdp/gDPPipelineMode.htm
- 12.2.3 Span Buffer Coherency http://n64devkit.square7.ch/pro-man/pro12/12-02.htm#03
- 24.4.1 Disable Atomic Primitives http://n64devkit.square7.ch/pro-man/pro24/24-04.htm#01
- glossary "span buffer coherency problem" http://n64devkit.square7.ch/keywords/index/data/system.htm#span%20buffer%20coherency%20problem

Some RDP timing by tharo:

https://discord.com/channels/874816709855440926/874816710639767585/1191456443073839135

<details>

Each was 200 runs drawing 6400 4x3 rectangles that cover the 320x240 framebuffer, numbers are in msec

```
1-CYCLE, IM_RD=0, Z_CMP=0, Z_UPD=0, NPRIM
  Buf:  pruned 4 outliers
  Pipe: pruned 4 outliers
   min_buf   avg_buf            max_buf   min_pipe  avg_pipe           max_pipe
  (6.360368, 6.362571591836734, 6.364752, 6.360816, 6.363043755102042, 6.3652)

1-CYCLE, IM_RD=0, Z_CMP=0, Z_UPD=0, 1PRIM
  Buf:  pruned 3 outliers
  Pipe: pruned 4 outliers
   min_buf   avg_buf             max_buf   min_pipe  avg_pipe            max_pipe
  (10.58064, 10.586511837563451, 10.58976, 10.58064, 10.586500734693878, 10.58976)
```

```
1-CYCLE, IM_RD=1, Z_CMP=1, Z_UPD=1, NPRIM
  Buf:  pruned 4 outliers
  Pipe: pruned 4 outliers
   min_buf    avg_buf             max_buf    min_pipe   avg_pipe            max_pipe
  (24.311552, 24.314218040816325, 24.316576, 24.314944, 24.317647836734693, 24.319968)

1-CYCLE, IM_RD=1, Z_CMP=1, Z_UPD=1, 1PRIM
  Buf:  pruned 3 outliers
  Pipe: pruned 3 outliers
   min_buf    avg_buf            max_buf    min_pipe   avg_pipe            max_pipe
  (25.885872, 25.88944194923858, 25.893808, 25.885872, 25.889446253807105, 25.893808)
```

</details>

from which I conclude "NPRIM is faster" (it was known already and correctly documented in the manual but timing proofs are nice)

OoT already uses NPRIM for everything afaict. (not changed by this PR)

PR #227, opened 2023-04-28
key check the remainder of f3d mat settings for upgrading super ultra old materials

thecozies:
there are no more possible key errors that could occur at this point. gamebun had a material that didnt even have combiner2 or rdp_settings somehow so i just made sure to verify every possible key here before copying


jesusyoshi54 (approved):
looks good to me

PR #230, opened 2023-06-13
Png exporting hotfixes

Lilaa3:
Fixes uncessary renamings for file names which makes using vanilla assets much better.


Dragorn421 (approved):
It seems like this change doesn't affect oot at all, `getNameFromPath` is not called neither on import nor export of the vanilla maps Deku Tree and Hyrule Field, which use a variety of image formats (including CI and rgba16)

Maybe it does change something in specific scenario but that's enough testing for me to say it doesn't affect oot

PR #232, opened 2023-06-22
[OoT] Scene-Table parser fix

HailToDodongo:
Fixes a few issues with the scene-table parser:
- Handle empty lines (they produce empty `DEFINE_SCENE(` otherwise)
- Ignore single-line comments
- Ignore comments after the first data entry (they would be duplicated on save otherwise)


Dragorn421 (approved):
:heavy_check_mark: for the changes
I added comments to not have to understand this twice, looks fine to you?

HailToDodongo:
Looks good 👍

PR #234, opened 2023-07-09
[MM] Basic ActorList.xml

HailToDodongo:
As discussed in discord, this PR adds a basic actor XML file for MM.
Basically all actor params are still missing, but i will make commits from time to time.
The other changes in the oot XML are either small fixes or making a few texts shorter to fit within the UI better.
I also replaced the weird Unicode `’` with regular ones `'`, since i had a few encoding issues in my editor with it.

One difference in the format itself compared to oot is, that i used the enum name instead of the variable name for `ObjectKey`.
This makes it easier for me, since the enum is what gets saved anyways and is just as unique.

### DIFF
--- a/fast64_internal/oot/data/xml/ActorList.xml
+++ b/fast64_internal/oot/data/xml/ActorList.xml
@@ -129,7 +129,7 @@ for each sub element (of <Actor>) mentioned below:
     <Type Index="1" Mask="0xF000">
       <Item Params="0000"> Large</Item>
       <Item Params="1000"> Large, Appears, Clear Flag</Item>
-      <Item Params="2000"> Boss Key’s Chest</Item>
+      <Item Params="2000"> Boss Key's Chest</Item>
       <Item Params="3000"> Large, Falling, Switch Flag</Item>
       <Item Params="4000"> Large, Invisible</Item>
       <Item Params="5000"> Small</Item>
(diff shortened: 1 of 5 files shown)


Yanis42:
I don't think it matters right now but just in case: the idea behind actor "key" is avoid the thing to break if the actor's ID changes, though it's probably fine with MM (it's not even used currently so if we really need it we can always add them later)

fast64_internal/oot/data/xml/ActorList.xml
```diff
@@ -931,12 +931,12 @@ ZROT 0x00FF = Gold Skulltula spawn var high byte (See Gold Skulltula Actor 0095)
...
-    <Property Index="3" Mask="0x0002" Name="Move South"/>
-    <Property Index="4" Mask="0x0004" Name="Move West"/>
-    <Property Index="5" Mask="0x0008" Name="Move East"/>
-    <Property Index="6" Mask="0x0F00" Name="Attack Speed"/>
+    <Property Index="1" Mask="0xF000" Name="Speed" TiedActorTypes="0000,0020"/>
+    <Property Index="6" Mask="0x0F00" Name="Radius"/>
+    <Bool Index="2" Mask="0x0001" Name="Move North"/>
+    <Bool Index="3" Mask="0x0002" Name="Move South"/>
+    <Bool Index="4" Mask="0x0004" Name="Move West"/>
+    <Bool Index="5" Mask="0x0008" Name="Move East"/>
```
Yanis42, line 939: iirc those indices need to be in numerical order, I think it's fine but it should be from 1 to whatever per param type (so here it's ``<Bool>``), here it starts at 2

also the 2nd ``<Property>`` index should be 2 not 6

HailToDodongo:
Ah right, i completely forgot that one.
I don't really want to pull actual names out since i have no need for them, but i guess initially setting this to a lower-case version of the enum would be fine?

Yanis42 (commented):
just a small thing I noticed

Yanis42:
> Ah right, i completely forgot that one. I don't really want to pull actual names out since i have no need for them, but i guess initially setting this to a lower-case version of the enum would be fine?

yeah, the actor key is it's debug name (like ``ACTOR_EN_TEST``) without the actor part and lowercased, I made it that way since we already use those a lot for years

Yanis42:
but again, it won't really matter until we're implementing MM, which should probably be done once OoT is cleaned up (it takes a lot of time unfortunately)

also I originally made the OoT XML file to create a more detailed actor panel on Fast64, that's why there's some useless informations right now

HailToDodongo:
no problem.
I fixed the order and added a key to the mm actors.

Btw, if you're interested or if there where issues implementing parameters in fast64, i made some changes via a conversion script in my project to make it easier to handle in my UI.

Yanis42 (approved):
documentation is cool

PR #235, opened 2023-07-12
[OOT] Adapted zcamedit to Fast64

Yanis42:
This PR takes every feature from [zcamedit](https://github.com/sauraen/zcamedit) to make cutscene motion work with Fast64 natively. The features themselves are the same, what changed is mostly how it works. This is far from perfect but it works well, I'm still putting this PR as a draft while I do more testing, fix issues I find etc, but this is mostly ready to review, I don't except big changes now

Also about the insertion feature, I was completely lost when I ported this, I don't know if what I did is the right thing so let me know if I should change something here (currently it's inserting if exporting the cutscene with the "Export Motion Data Only" checkbox enabled)

<details closed>
<summary>Things that change between this and zcamedit</summary>

- New export/import that supports old command names (the import seems slow when it gets to create camera shot armatures for some reasons)
- re-organised files and renamed most things to """suit Fast64's standards"""
- New object naming standard:
    * the cutscene object is still named ``Cutscene.Name``
    * its direct childs are named ``CS_XX.YYYY ZZ`` where ``XX`` is the cutscene number in the scene, ``YYYY`` is the element (i.e camera shot, actor cue list) and ``ZZ`` the number of this element (i.e the 2nd camera shot, the 14th actor cue list)
    * if these childs have childs objects they are named like their parents except the number is ``X.YY`` where ``X`` is the parent element number and ``YY`` the number of the sub-element (I'll link a screenshot, it will be easier to understand lol)
- a camera is created automatically when creating the cutscene, I used zcamedit before and I was thinking it's annoying that it moves the camera you're using for your map, the camera's name ``Cutscene.Name.Camera``
- added more OoT empty types to make handling actor cues easier (the user shouldn't worry about it)
- added a "dummy" actor cue when you're creating an actor cue list, this is to make it obvious that you need an extra cue to set the real last one's end frame/position
- added an endFrame property for actor cues to make it easier to understand what your end frame will be, this is read-only and it gets it's value from the next actor cue's start frame (nothing new here, it was already doing that at export)
- support for every cutscene commands, though they aren't all processed yet since PR #208 updates most names, my plan is to move the exporter/importer for motion outside of the folder when 208 is merged (this can wait)
- create a basic camera shot when creating the cutscene
- added a button to add a new bone, it's only duplicating the last one (it looks like it works properly at least, I'm aware that bones are weird)
- added a warning text if the user is in pose mode
- added a button to play the camera motion, this is done by selecting the cutscene you want to preview
- preview code is kinda the same, except function renames and other tweaks (this file is hard to understand for me)
- added cutscenes as an option to the scene importer
- added a button to toggle decomp names (useful if you're exporting for a non-decomp use), note that decomp names will be always used if using HackerOoT features
- added a button to export motion data only, this is used for inserting motion data in a file if the cutscene exists, otherwise it makes a new array (as defined by the cutscene's name on Blender)
- changed the old names to the new ones
- I have to admit I haven't tried this since it's unused and not recommended to use but commands 0x7/0x8, aka ``CS_CAM_EYE`` and ``CS_CAM_AT`` should work properly (idk about the preview though)
- added some documentation in the readme, most motion docs are taken from zcamedit and decomp
- upgrade old zcamedit data to the new props
- added a prompt to set the cutscene name when adding one
- locked the location, scale and rotation of the cutscene when it's created, also it resets every value, this is to avoid issues with the camera motion
- actor cue preview objects are automatically created at import if the actor cue has different positions
</details>
<details closed>
<summary>Object Names</summary>

</details>
<details closed>
<summary>Some Panels</summary>

</details>

Let me know if there's anything wrong or if you have questions about this!


Yanis42:
(commit [b53c97e](https://github.com/Fast-64/fast64/pull/235/commits/b53c97e1a10be17ddb7890f0679e9c9ba6720e36) shows how easy it is to add a process implementation for a command, they're all parsed but only a few are currently processed)

Yanis42:
(I'm gonna wait few days to see if there's any feedback from the users and if there's nothing then I'll mark it ready for review I guess)

Yanis42:
I merged my transition/misc commands preview work in this branch, I'll try to get the letterbox working, that would be really nice

Yanis42:
this should be ready to review now, I tested again and fixed the issues I found, also I added some extra features and things I forgot like adding the cutscene to the scene header

sauraen (approved):

PR #239, opened 2023-07-14
[OOT] Applied new organisation to ``oot_level_writer.py`` and ``oot_level_classes.py``

Yanis42:
This PR simply moves the different export functions to their proper folders and files (idk if I'm gonna cleanup the code or make a new exporter yet, ideally I'd like to apply the same system I did for cutscene motion to every exporter/importer, I think it makes the code easy to follow this way but idk I need to think about it, but I'm not satisfied with how I organised the functions here...)

(also I'm probably going to split level_parser outside this PR since there's already a good amount of changes)


Yanis42:
I'm closing this for now as I'm working on a new exporter that will replace this one (hopefully)

PR #240, opened 2023-07-14
[OOT] Applied new organisation to ``oot_level_parser.py``

Yanis42:
This PR simply moves the different import functions to their proper folders and files

also I moved the ``removeComments`` function to the main ``utility.py`` file since it can be used for any game


Dragorn421 (approved):
Importing deku tree not broken :heavy_check_mark: 

Merging this will cause conflicts in #366 but it's a price I'm willing to (make Yanis) pay

kurethedead (commented):
Actually, should the folder structure for the importer mirror the new exporter? (`oot/importer` vs `oot/scene/importer`)

Yanis42:
I moved it outside the scene folder

PR #242, opened 2023-07-16
[OOT] Detailed Actor Panel (Actors, Transition Actors, Entrance Actors)

Yanis002:
Remake of PR #56 but the code is shorter by a lot, also the XML data might not be always accurate so I'll probably do some documentation (which I did a bit)

Anyway, tl;dr this is basically [that tool](https://kibako.gitlab.io/?actor-params) but in Fast64, and this is a draft as most features are done, I just forgot to make a small upgrade logic and also the UI for entrances and transition actors


Yanis42:
this should be ready to review now, issues will most likely be caused by the XML documentation but it *should* be fine

fast64_internal/oot/oot_utility.py
```diff
@@ -874,6 +874,27 @@ def _eval(node):
     return f"0x{_eval(node.body):X}"
 
 
+def getShiftFromMask(mask: int):
+    """Returns the shift value from the mask"""
+
+    # get the shift by subtracting the length of the mask
+    # converted in binary on 16 bits (since the mask can be on 16 bits) with
+    # that length but with the rightmost zeros stripped
+    return int(f"{len(f'{mask:016b}') - len(f'{mask:016b}'.rstrip('0'))}", base=10)
```
Dragorn421, line 883: what is the difference with
```suggestion
    return len(f'{mask:016b}') - len(f'{mask:016b}'.rstrip('0'))
```
or even
```suggestion
    return len(f'{mask:b}') - len(f'{mask:b}'.rstrip('0'))
```

I would also add an assert the mask *is* a mask, i.e. that `f'{mask:b}'.rstrip('0')` looks like "1111" and not "11101" or "":
```
mask_binary_str = f"{mask:b}"
assert set(mask_binary_str.rstrip("0")) == {"1"}, mask_binary_str
```

fast64_internal/oot/oot_utility.py
```diff
@@ -874,6 +874,27 @@ def _eval(node):
...
+
+
+def getFormattedParams(mask: int, value: str, isBool: bool):
```
Dragorn421, line 886: can mask be None ? the function partly does and doesn't handle `mask is None`

fast64_internal/oot/oot_utility.py
```diff
@@ -874,6 +874,27 @@ def _eval(node):
...
+    """Returns the parameter with the correct format"""
+    shift = getShiftFromMask(mask) if mask is not None else 0
+
+    if not int(getEvalParams(value), base=16):
```
Dragorn421, line 890: not sure what this means, you want to check for 0 ? `== 0` would be clearer

also `getEvalParams` should be modified to return an int instead of a string

or do something like:
```py
def getEvalParamsInt(...):
    ... current getEvalParams code minus stringify
def getEvalParams(...):
    i = getEvalParamsInt(...)
    return f"{i:X}"
```

fast64_internal/oot/actor/properties.py
```diff
@@ -23,6 +30,52 @@
...
+        "Type": "type",
+        "Property": "props",
+        "Bool": "bool",
+        "Enum": "enum",
+        "ChestContent": "chestContent",
+        "Collectible": "collectibleDrop",
+        "Message": "naviMsg",
+    }
+    suffix = paramTypeToObjName[paramType] if paramType != "Flag" else flagTypeToObjName[paramSubType]
+    return f"{actorKey}.{suffix}{paramIndex}"  # i.e: ``en_test.props1``
```
Dragorn421, line 45:
```suggestion
    return f"{actorKey}.{suffix}{paramIndex}"  # e.g.: ``en_test.props1``
```

i.e. is for details https://fr.wiktionary.org/wiki/i.e.
e.g. is for examples https://fr.wiktionary.org/wiki/e._g.#fr

fast64_internal/oot/actor/properties.py
```diff
@@ -110,22 +163,189 @@ def draw_props(
...
     rotOverride: BoolProperty(name="Override Rotation", default=False)
-    rotOverrideX: StringProperty(name="Rot X", default="0")
-    rotOverrideY: StringProperty(name="Rot Y", default="0")
-    rotOverrideZ: StringProperty(name="Rot Z", default="0")
+    rotOverrideX: StringProperty(name="Rot X", default="0x0000")
+    rotOverrideY: StringProperty(name="Rot Y", default="0x0000")
+    rotOverrideZ: StringProperty(name="Rot Z", default="0x0000")
+
+    params: StringProperty(
+        name="Actor Parameter",
```
Dragorn421, line 174: could there be a comment somewhere about the difference between `actorParam` and `params`

fast64_internal/oot/actor/properties.py
```diff
@@ -110,22 +163,189 @@ def draw_props(
...
+        default="0x0000",
+        get=lambda self: self.getParamValue("Params"),
+        set=lambda self, value: self.setParamValue(value, "Params"),
+    )
+    isRotX: BoolProperty(name="", default=False, get=lambda self: self.isRotation("XRot"))
+    isRotY: BoolProperty(name="", default=False, get=lambda self: self.isRotation("YRot"))
+    isRotZ: BoolProperty(name="", default=False, get=lambda self: self.isRotation("ZRot"))
+    rotX: StringProperty(
```
Dragorn421, line 182: what do `isRotX` and co. mean?

fast64_internal/oot/oot_upgrade.py
Dragorn421, line 1: there's upgrade code but no `cur_version` bump anywhere?

fast64_internal/oot/oot_level_writer.py
```diff
@@ -741,7 +743,21 @@ def ootProcessLOD(
     DLGroup.addDLCall(transparentLOD.draw, "Transparent")
 
 
-def ootProcessEmpties(scene, room, sceneObj, obj, transformMatrix):
+def getActorRotation(actorProp: OOTActorProperty, blendRots: list[int]):
+    # Figure out which rotation to export, Blender's or the override
+    override = "Override" if actorProp.actorID == "Custom" else ""
+    overrideRots = [getattr(actorProp, f"rot{override}{rot}") for rot in ["X", "Y", "Z"]]
+    exportRots = [f"DEG_TO_BINANG({(rot * (180 / 0x8000)):.3f})" for rot in blendRots]
```
Dragorn421, line 750: so there are three sets of rotations?
- one from the blender object
- two "override", one for non-custom actors and one for custom? why are these two different? why are the properties for "Custom" named "Override"?

fast64_internal/oot/oot_utility.py
Yanis42: shift = shift_from_mask if there's a mask else if there's no mask then there's no shift

fast64_internal/oot/actor/properties.py
Yanis42: ``actorParam`` is used for custom actors (actors with the id custom)

fast64_internal/oot/actor/properties.py
Yanis42: that's an internal prop used to draw the string props for X, Y or Z rots depending on what the actor is using (defined by XML data)

fast64_internal/oot/oot_upgrade.py
Yanis42: yeaaaaahhhhhh I guess, that PR is kinda old now so I probably didn't know yet about that and end up forgetting about it

fast64_internal/oot/oot_level_writer.py
Yanis42: - blender object (which won't be used if the actor can have parameters in its rotation, which makes sense to me)
- string prop rotation for actors without the id "custom"
- string prop rotation for custom actors

iirc the override thing is gone and it all depend on the xml data, I fail to see how using rotations as real rotation and not parameter can be useful if the actor is using rotations

btw this depends on other parameters, some actors doesn't always actually check for the rotation for params, in this case the blender object rotation can be used (aka if you don't see any rotation param field in the fast64 ui)

fast64_internal/oot/oot_utility.py
Yanis42: I wrote that function 2 years ago I can't remember how it works

fast64_internal/oot/oot_utility.py
Yanis42: mask is never ``None``, this is how mask is parsed from the xml: ``elem.get("Mask", "0xFFFF")``, if there's no mask defined it defaults to 0xFFFF ([line's here](https://github.com/Yanis42/fast64/blob/acfe3619860ae0f34f62ee5b01b7d85f9c8cda33/fast64_internal/oot/data/oot_actor_data.py#L97))

I'm not sure why there's ``if mask is not None else 0`` in that function :thonk:

fast64_internal/oot/actor/properties.py
Dragorn421: both are a `StringProperty` without get/set, it doesn't seem there needs to be two?

fast64_internal/oot/actor/properties.py
Dragorn421: so like `isRotXForActorConfig` or smth idk?

fast64_internal/oot/oot_utility.py
Yanis42: I'm not a fan of the second option so I'll keep the ``016b``, I will also add the assert just in case

fast64_internal/oot/actor/properties.py
Yanis42: ``params`` use get/set

fast64_internal/oot/actor/properties.py
Yanis42: when ``isRotX`` is true it means "this actor is using parameters inside this rotation", the name is too confusing, I'm gonna rename it ``isRotXUsedByActor``

fast64_internal/oot/oot_upgrade.py
Yanis42: since I remembered why, ``cur_version`` doesn't exist anymore. we discussed about that a while ago and figured out checking if old props exist in a blend is stronger than a version check (idr the specifics though, iirc it was kure's idea)

Reonu:
@Dragorn421 I'm using this rn and it works great, it makes oot scene creation 100 times easier and it's a crime that it's not merged :(

Yanis42:
(just so I don't forget) todo: look into upgrade issues

Reonu (approved):
Will this get merged soon? I've been using it for quite some time now and it seems to have no issues (the upgrade thing was seemingly a false alarm too). 

Giving it an approving review since I've been using it a lot and I can confirm it works well

Yanis42:
I added this to the next milestone a while ago so it will be merged before september for sure (hopefully)

also I don't remember if it was this PR specifically but I think I had another issue but I can't remember what is was 🥲

fast64_internal/oot/actor/properties.py
```diff
@@ -110,22 +164,193 @@ def draw_props(
...
 
+    def isRotationUsedByActor(self, target: str):
+        actor = ootData.actorData.actorsByID[self.actorID]
+        curType = None
+        for param in actor.params:
+            if param.type == "Type":
+                objName = getObjName(actor.key, param.type, param.subType, param.index)
+                curType = getEvalParamsInt(getattr(self, objName))
+
+            if curType is not None and curType in param.tiedTypes or len(param.tiedTypes) == 0:
```
kurethedead, line 223: Make this `elif`? Or maybe add a comment explaining that the first param is always of type `"Type"`, and that there is only one param like this. This code is a little confusing without this context.

fast64_internal/oot/actor/properties.py
```diff
@@ -110,22 +164,193 @@ def draw_props(
...
+    def draw_params(self, layout: UILayout, objName: str):
+        actor = ootData.actorData.actorsByID[self.actorID]
+        curType = None
+        for param in actor.params:
+            propName = getObjName(actor.key, param.type, param.subType, param.index)
+
+            if param.type == "Type":
+                curType = getEvalParamsInt(getattr(self, propName))
+
+            if curType is not None and curType in param.tiedTypes or param.type == "Type" or len(param.tiedTypes) == 0:
```
kurethedead, line 331: Make this `elif`? Or maybe add a comment explaining that the first param is always of type `"Type"`, and that there is only one param like this. This code is a little confusing without this context.

kurethedead (changes requested):
Seems mostly good. However, I think `Type`/`Enum` should support custom values (ex. you want to add a new Poe type while still using the existing Poe actor), and parameters should be allowed to use string inputs (ex. `MY_CUSTOM_FLAG` instead of `0x01`).

fast64_internal/oot/actor/properties.py
```diff
@@ -110,22 +164,193 @@ def draw_props(
...
+        name="Rot Z",
+        default="0",
+        get=lambda self: self.getParamValue("ZRot"),
+        set=lambda self, value: self.setParamValue(value, "ZRot"),
+    )
+
+    # internal usage only
+    isRotXUsedByActor: BoolProperty(name="", default=False, get=lambda self: self.isRotationUsedByActor("XRot"))
+    isRotYUsedByActor: BoolProperty(name="", default=False, get=lambda self: self.isRotationUsedByActor("YRot"))
+    isRotZUsedByActor: BoolProperty(name="", default=False, get=lambda self: self.isRotationUsedByActor("ZRot"))
```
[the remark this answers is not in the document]
Yanis42: is it gonna be saved in the blend if it's not using a blender prop? I don't remember correctly how it works

fast64_internal/oot/actor/operators.py
```diff
@@ -6,13 +6,55 @@
...
 
 
+class OOT_SearchChestContentEnumOperator(Operator):
+    bl_idname = "object.oot_search_chest_content_enum_operator"
+    bl_label = "Select Chest Content"
+    bl_property = "chestContent"
+    bl_options = {"REGISTER", "UNDO"}
+
+    chestContent: EnumProperty(items=ootData.actorData.ootEnumChestContent, default="item_heart")
+    objName: StringProperty()
```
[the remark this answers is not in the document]
Yanis42: tbh it should even be a pointer to ``OOTActorProperty``

fast64_internal/oot/actor/operators.py
```diff
@@ -6,13 +6,55 @@
...
+
+
+class OOT_SearchNaviMsgIDEnumOperator(Operator):
+    bl_idname = "object.oot_search_navi_msg_id_enum_operator"
+    bl_label = "Select Message ID"
+    bl_property = "naviMsgID"
+    bl_options = {"REGISTER", "UNDO"}
+
+    naviMsgID: EnumProperty(items=ootData.actorData.ootEnumNaviMessageData, default="msg_00")
+    objName: StringProperty()
```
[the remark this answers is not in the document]
Yanis42: for consistency purposes I'm also making this change for ``OOT_SearchActorIDEnumOperator`` (even though it's not modified in this PR)

fast64_internal/oot/actor/operators.py
```diff
@@ -6,13 +6,55 @@
...
 
+class OOT_SearchChestContentEnumOperator(Operator):
+    bl_idname = "object.oot_search_chest_content_enum_operator"
+    bl_label = "Select Chest Content"
+    bl_property = "chestContent"
+    bl_options = {"REGISTER", "UNDO"}
+
+    chestContent: EnumProperty(items=ootData.actorData.ootEnumChestContent, default="item_heart")
+    objName: StringProperty()
+    propName: StringProperty()
```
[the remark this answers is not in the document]
Yanis42: ``propName`` is one of the properties created by fast64 when it loads (see ``initOOTActorProperties``), in this context it will be the name of the property holding the value of the chest item

fast64_internal/oot/actor/operators.py
```diff
@@ -6,13 +6,55 @@
...
+
+class OOT_SearchNaviMsgIDEnumOperator(Operator):
+    bl_idname = "object.oot_search_navi_msg_id_enum_operator"
+    bl_label = "Select Message ID"
+    bl_property = "naviMsgID"
+    bl_options = {"REGISTER", "UNDO"}
+
+    naviMsgID: EnumProperty(items=ootData.actorData.ootEnumNaviMessageData, default="msg_00")
+    objName: StringProperty()
+    propName: StringProperty()
```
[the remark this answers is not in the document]
Yanis42: (see https://github.com/Fast-64/fast64/pull/242#discussion_r1703400687)

fast64_internal/oot/actor/operators.py
```diff
@@ -6,13 +6,55 @@
...
 class OOT_SearchActorIDEnumOperator(Operator):
     bl_idname = "object.oot_search_actor_id_enum_operator"
     bl_label = "Select Actor ID"
     bl_property = "actorID"
     bl_options = {"REGISTER", "UNDO"}
 
-    actorID: EnumProperty(items=ootData.actorData.ootEnumActorID, default="ACTOR_PLAYER")
+    actorID: EnumProperty(items=lambda self, context: ootData.actorData.getItems(self.actorUser))
     actorUser: StringProperty(default="Actor")
     objName: StringProperty()
```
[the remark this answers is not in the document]
Yanis42: I assume you meant ``objName`` like the other ones? if so then I already did that (I forgot I made a small change there)

fast64_internal/oot/actor/properties.py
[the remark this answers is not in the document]
Yanis42: nvm I'm stupid, you meant converting these members into class methods I think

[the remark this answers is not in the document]
Yanis42: in this case I can just call ``isRotationUsedByActor`` since these are only used in the same class

fast64_internal/oot/actor/properties.py
```diff
@@ -110,22 +164,193 @@ def draw_props(
...
+
+    # used for actors with the id "Custom"
     actorParam: StringProperty(name="Actor Parameter", default="0x0000")
     rotOverride: BoolProperty(name="Override Rotation", default=False)
-    rotOverrideX: StringProperty(name="Rot X", default="0")
-    rotOverrideY: StringProperty(name="Rot Y", default="0")
-    rotOverrideZ: StringProperty(name="Rot Z", default="0")
+    rotOverrideX: StringProperty(name="Rot X", default="0x0000")
+    rotOverrideY: StringProperty(name="Rot Y", default="0x0000")
+    rotOverrideZ: StringProperty(name="Rot Z", default="0x0000")
```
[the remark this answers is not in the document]
Yanis42: I mean, sure, but how am I supposed to handle the getter function? you need to return something there, I'm not sure how to handle that

the setter should be doable though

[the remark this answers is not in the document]
Yanis42: btw to clarify, the issue I'm having here is that I think you can't have a getter function that doesn't return anything, it will either raise an error or set a blank value which is not what we want here, for the setter you can just return the value from the parameters and it should be fine

also there's a scenario where I'm just completely stupid and forgot how this works

fast64_internal/oot/actor/operators.py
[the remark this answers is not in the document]
Yanis42: also small clarification here too, ``propName`` depends on the actor id and the prop index (as defined by the XML data ``ActorList.xml``), basically it's ``actor_id.propN`` where ``N`` is the index I mentioned

for these reasons you can't use an hardcoded value since we want to keep this whole system flexible enough so that people can edit the XML data to add their own actor or adjust an actor depending on the OoT mod they do (not mandatory of course)

fast64_internal/oot/actor/properties.py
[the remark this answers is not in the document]
kurethedead: I think I understand, the point of having a separate property is that its a way to get/set actor-specific properties? In that case it should be fine, but you should add a comment explaining the distinction between the duplicate properties.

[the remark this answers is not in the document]
Yanis42: ``rotOverride[X-Y-Z]`` is only used for the actor id "custom" to be able to write any value we want, I'm gonna rename those to ``rot_x_custom`` etc to make it clearer (and I'll add a comment to explain why it's required)

kurethedead (changes requested):
Some issues:
1. For Treasure Chest / Navi Msg, choosing a custom chest content / message value causes an error.
2. If eval params is enabled but the params fail to evaluate, I think the value should still be the unevaluated params. Currently its left blank, which results in an invalid export.
3. For Rot XYZ + Params, if string input is used, I think masking / shifting should still be applied, so that its consistent with numerical input.

Yanis002:
I fixed the issues you mentioned, the only thing left is this: ``For Rot XYZ + Params, if string input is used, I think masking / shifting should still be applied, so that its consistent with numerical input.`` but I don't understand what you mean here

Reonu (approved):
Can this get merged? I've been using it for a while, it works perfectly and the difference in usability for oot scene editing is insane.

kurethedead (approved):
Seems good to me.

PR #243, opened 2023-07-16
[OOT] Use Room objects to set which room you come from/go to

Yanis42:
This avoids having to change the room index on each door and it's easier to keep track of which door do what, also I changed ``dontTransition`` to ``isRoomTransition`` (with a proper name on the UI)

While doing this I found two issues (now fixed), the scene importer/exporter wasn't importing/exporting properly doors that don't make a room change (i.e: the door near the gate in Hyrule Castle)


fast64_internal/oot/actor/properties.py
```diff
@@ -145,15 +151,19 @@ def draw_props(self, layout: UILayout, altRoomProp: OOTAlternateRoomHeaderProper
...
 class OOTTransitionActorProperty(PropertyGroup):
-    roomIndex: IntProperty(min=0)
+    fromRoom: PointerProperty(type=Object, poll=lambda self, object: self.isRoomEmptyObject(object))
+    toRoom: PointerProperty(type=Object, poll=lambda self, object: self.isRoomEmptyObject(object))
     cameraTransitionFront: EnumProperty(items=ootEnumCamTransition, default="0x00")
     cameraTransitionFrontCustom: StringProperty(default="0x00")
     cameraTransitionBack: EnumProperty(items=ootEnumCamTransition, default="0x00")
     cameraTransitionBackCustom: StringProperty(default="0x00")
-    dontTransition: BoolProperty(default=False)
+    isRoomTransition: BoolProperty(name="Is Room Transition", default=False)
```
sauraen, line 160: I think this default needs to flip; the old version was "doors are room transitions unless dontTransition is selected".

fast64_internal/oot/oot_level_parser.py
```diff
@@ -727,12 +727,15 @@ def parseTransActorList(
 
             sharedSceneData.transDict[actorHash] = actorObj
 
-            if roomIndexFront != 255:
-                parentObject(roomObjs[roomIndexFront], actorObj)
-                transActorProp.roomIndex = roomIndexBack
+            fromRoom = roomObjs[roomIndexFront]
+            toRoom = roomObjs[roomIndexBack]
+            if roomIndexFront != roomIndexBack:
```
sauraen, line 732: The old logic for checking for 255 is gone, are you sure this is equivalent? (e.g. `roomIndexFront == validRoomIndex` and `roomIndexBack == 255` is not possible?) I don't know this system so I don't know offhand.

sauraen (changes requested):
One minor change and a question, otherwise looks good.

fast64_internal/oot/oot_level_parser.py
Yanis42: iirc 255 is only used for non-transition doors, idr the specifics though but I do remember noticing a weirdness like this

Yanis42: I'll take a look at this to figure it out

Yanis42: (for the record: I remembered the reason and it's better to check if the room is the same on both sides to tell if a room transitions to another one or not since the camera can be anything you want)

sauraen (approved):
Thanks!

PR #245, opened 2023-07-28
[SM64] Geo Layout Export Node Optimization

jesusyoshi54:
Start nodes are removed unless explicitly placed by the user, now all geo nodes simply start with the root node of the object/bone placed by the user.
I also optimized skinning to combine the skinned mesh into the previous geo layout node if there was no nodes in between the previous animated bone and the child, instead of placing that skinned mesh in its own display list node.

I tested this on mario and a few other object geo layouts and exports did not show any visible errors. Mario experienced about a 20% reduce in line count.

### DIFF
--- a/fast64_internal/sm64/sm64_geolayout_writer.py
+++ b/fast64_internal/sm64/sm64_geolayout_writer.py
@@ -386,16 +386,13 @@ def convertArmatureToGeolayout(
         geolayoutGraph = GeolayoutGraph(name + "_geo")
         if armatureObj.use_render_area:
             rootNode = TransformNode(StartRenderAreaNode(armatureObj.culling_radius))
-        else:
-            rootNode = TransformNode(StartNode())
-        geolayoutGraph.startGeolayout.nodes.append(rootNode)
+            geolayoutGraph.startGeolayout.nodes.append(rootNode)
         meshGeolayout = geolayoutGraph.startGeolayout
 
-    for i in range(len(startBoneNames)):
-        startBoneName = startBoneNames[i]
-        if i > 0:
-            meshGeolayout.nodes.append(TransformNode(StartNode()))
-        processBone(
+    for i, startBoneName in enumerate(startBoneNames):
+        # if i > 0:
+            # meshGeolayout.nodes.append(TransformNode(StartNode()))
+        materialOverrides = processBone(
             fModel,
             startBoneName,
             obj,
@@ -404,7 +401,7 @@ def convertArmatureToGeolayout(
             None,
             None,
             None,
-            meshGeolayout.nodes[i],
+            i,
             [],
             name,
             meshGeolayout,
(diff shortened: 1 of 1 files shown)


fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -433,22 +428,49 @@ def convertArmatureToGeolayout(armatureObj, obj, convertTransformMatrix, camera,
...
+            i,
             [],
             name,
             meshGeolayout,
             geolayoutGraph,
             infoDict,
             convertTextureData,
         )
+
+    def walk(node, fMeshes):
```
Yanis002, line 440: you should add type hints here (if possible, if it's too annoying then it's fine), this applies to the functions that are added in this pr

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -433,22 +428,49 @@ def convertArmatureToGeolayout(armatureObj, obj, convertTransformMatrix, camera,
...
+        base_node = node.node
+        if type(base_node) == JumpNode:
```
Yanis002, line 442: I was wondering, if this better than `isinstance()`? looks like you're using it later, I think one of both should be changed for consistency

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -433,22 +428,49 @@ def convertArmatureToGeolayout(armatureObj, obj, convertTransformMatrix, camera,
...
+            if base_node.geolayout:
```
Yanis002, line 443: I think it'd be better with `if base_node.geolayout is not None` to make the intent clearer

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -433,22 +428,49 @@ def convertArmatureToGeolayout(armatureObj, obj, convertTransformMatrix, camera,
...
+                for node in base_node.geolayout.nodes:
+                    fMeshes = walk(node, fMeshes)
+        fMesh = getattr(base_node, "fMesh", None)
+        if fMesh:
```
Yanis002, line 447: same thing here (`if fMesh is not None`)

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -1037,179 +1059,149 @@ def geoWriteTextDump(textDumpFilePath, geolayoutGraph, levelData):
...
+                child_node.node.drawLayer,
+                0,
+                option_geo_layout,
+                geolayoutGraph,
+                option_geo_layout.name,
+            )
+
+            # if this is a material override, add to the list
+            # if child_node.node.material is not None:
+            # overrideNodes.append(optionGeolayout.nodes[0])
```
Yanis002, line 1125: not sure if this is intended but if it's leftover commented code then it should be removed

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -1669,7 +1661,14 @@ def processBone(
     zeroTranslation = isZeroTranslation(translate)
     zeroRotation = isZeroRotation(rotate)
 
-    # hasDL = bone.use_deform
+    # true when this is the start of the geo layout
+    # if there is no parent, then instead set the node to be the root of our geo layout
+    if type(parentTransformNode) == int:
```
Yanis002, line 1666: `isinstance`?

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -1682,9 +1681,12 @@ def processBone(
             if not zeroRotation:
                 node = DisplayListWithOffsetNode(int(bone.draw_layer), hasDL, mathutils.Vector((0, 0, 0)))
 
-                parentTransformNode = addParentNode(
-                    parentTransformNode, TranslateRotateNode(1, 0, False, translate, rotate)
-                )
+                if parentTransformNode:
```
Yanis002, line 1684: `if parentTransformNode is not None` (there's another ones later in the file)

Yanis002 (commented):
I'm not an expert of this game but I can still review code so here we are, most comments are just function type hints and use `is None`/`is not None` when needed

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -1682,9 +1677,12 @@ def processBone(
...
 
-                parentTransformNode = addParentNode(
-                    parentTransformNode, TranslateRotateNode(1, 0, False, translate, rotate)
-                )
+                if parentTransformNode is not None:
+                    parentTransformNode = addParentNode(
+                        parentTransformNode, TranslateRotateNode(1, 0, False, translate, rotate)
+                    )
+                else:
+                    geolayout.nodes.append(TranslateRotateNode(1, 0, False, translate, rotate))
```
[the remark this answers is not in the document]
jesusyoshi54: I think kaze tested it forever ago but we can get more testers

KazeEmanuar:
This branch is omitting important commands to connect armatures using vertex stripping, breaking most animated actors

PR #248, opened 2023-09-03
Fix default mario.blend

YoshiMilkman:
The default mario.blend has a few errors and inconsistencies with materials and UV's. Changes include recoloring some tris on Mario's upper back and head to match the original model, fixing wing cap hand wing UV's, and adding a dark material on the underside of Mario's held hats (it's there in the original model but hard to see usually). Materials have also been upgraded to V5.


jesusyoshi54 (approved):

Yanis42 (approved):

FrijolesYQueso (approved):
Definitely needs to be merged.

PR #249, opened 2023-09-07
Fixed large uv wrapping

Lilaa3:
Arthur suggested this a solution, he said he could test it later today. Until then this will remain a draft pr.


Dragorn421 (commented):
Does fast64 use the max shift of two textures for exporting? (Can you link to code) (why is this even a per texture setting in fast64)

fast64_internal/f3d/f3d_writer.py
```diff
@@ -140,6 +140,29 @@ def getInfoDict(obj):
...
 
 
+def getSTUVRepeats(tex_prop: "TextureProperty") :
+    SShift, TShift = 2 ** tex_prop.S.shift, 2 ** tex_prop.T.shift
+    sMirrorScale = 2 if tex_prop.S.mirror else 1
+    tMirrorScale = 2 if tex_prop.T.mirror else 1
+    return (SShift * sMirrorScale, TShift * tMirrorScale)
+
+
+def getUVInterval(f3dMat) -> tuple[int, int]:
```
Dragorn421, line 150: More generally this returns float not int (i.e. if shift<0)

Dragorn421, line 150: I don't think this behaves as expected if e.g. tex0 has shift=0 mirror=True and tex1 has shift=1 mirror=False

fast64_internal/f3d/f3d_writer.py
```diff
@@ -169,12 +192,9 @@ def fixLargeUVs(obj):
...
             continue
 
-        f3dMat = material.f3d_mat
-
-        UVinterval = [
-            2 if f3dMat.tex0.S.mirror or f3dMat.tex1.S.mirror else 1,
-            2 if f3dMat.tex0.T.mirror or f3dMat.tex1.T.mirror else 1,
-        ]
+        # To prevent wrong UVs when wrapping UVs into valid bounds,
+        # we need to account for the highest texture shift and if mirroring is active.
```
Dragorn421, line 196: To reiterate on my previous comments why "highest"? There must be a piece of code elsewhere on exporting that handles this similarly? Where

jesusyoshi54:
I have some comments on style but the actual code makes sense and if it is tested then I don't feel the need for any functionality changes

FazanaJ:
Screenshot 1
Screenshot 1
Noticable cutoff from the second texture when using shifting

Yanis42 (commented):
code is ok to me but I still need to get feedback from mario people and I need to see if anything breaks on oot (since this is F3D stuff)

I'll approve once both sides are tested and ok

Lilaa3:
> Screenshot 1 Screenshot 1 Noticable cutoff from the second texture when using shifting

Lilaa3:
Oops closed on accident, what I was going to say is that fazana had incorrect mask values, the pr should be working correctly.

Lilaa3:
> code is ok to me but I still need to get feedback from mario people and I need to see if anything breaks on oot (since this is F3D stuff)
> 
> I'll approve once both sides are tested and ok

Could you test then please? This shouldn´t affect oot differently at all.

jesusyoshi54 (approved):
works on my machine

PR #253, opened 2023-09-20
[OoT] Small cleanup in ``oot_constants.py``

Yanis42:
this wasn't what I wanted to do when I started vscode but here it is lol


sauraen (approved):
Great!

PR #254, opened 2023-09-22
Use object.type instead of object.data to check which type it is

Yanis42:
ngl idk how to name this PR lol, might need testing before merging to make sure everything's fine


Lilaa3:
lgtm, considering it removes the need for a none check and makes the code shorter and cleaner.

jesusyoshi54:
looks good, just needs conflicts fixed.
changes don't really need intensive testing imo since you can statically tell that the one line changes won't make a difference.

Dragorn421 (approved):
Things seem not broken :tm: in my crude testing.
and yeah hopefully this is straightforward enough to not break anything in the first place Prayge

PR #255, opened 2023-09-23
[OOT] Use Room objects to set which room to use for entrances

Yanis42:
Exactly what PR #243 does but applied to entrances

Small note: because of the current exporter it's still require a room as a parent but I'm working on something that, if it goes well, will remove this requirement


sauraen (commented):
Let's let #243 get merged first, cause this has some duplicated code from there.

fast64_internal/oot/oot_level_writer.py
```diff
@@ -819,8 +819,14 @@ def ootProcessEmpties(scene, room, sceneObj, obj, transformMatrix):
...
         elif obj.ootEmptyType == "Entrance":
             entranceProp = obj.ootEntranceProperty
-            spawnIndex = obj.ootEntranceProperty.spawnIndex
-            addActor(scene, OOTEntrance(room.roomIndex, spawnIndex), entranceProp.actor, "entranceList", obj.name)
+            spawnIndex = entranceProp.spawnIndex
+
+            if entranceProp.tiedRoom is not None:
+                roomIndex = entranceProp.tiedRoom.ootRoomHeader.roomIndex
+            else:
+                raise PluginError("ERROR: The room isn't set!")
```
sauraen, line 827: "In scene entrance, tied room empty object not set"

sauraen (changes requested):
Please make this error message clearer, otherwise looks good to me.

sauraen (approved):
Looks good!

PR #257, opened 2023-10-11
[OoT] Improved "Add Path" button

Yanis42:
- Added Shapes (Square, Triangle, Straight Line and Trapezium)
- Button to choose if the shape should be closed or not (adds an extra point)
- Automatically parent the path to the scene object, if selected beforehand


sauraen (approved):
:+1:

PR #264, opened 2023-12-01
Hotfix for Bloxifilmz

sauraen:
Some garbled F3D classes in a SM64 binary export codepath. Looks like this was updated in some previous PR and nobody ran this since.


sauraen:
Overriding merge as this is a small change on an evidently rarely used codepath.

PR #268, opened 2023-12-25
Fix SM64 default light direction for F3DEX3 compatibility

sauraen:
Vanilla SM64 uses a fixed direction of (0x28, 0x28, 0x28) for one directional light for the entire game (in camera space).

This vector is not normalized; its length is about 0.55f. However, F3D through F3DEX2 microcodes normalize light directions after transforming them, so it doesn't matter that this is too short. In contrast, F3DEX3 does not normalize light directions, because it transforms and normalizes vertex normals instead. So, this light direction being too short means that all the lighting is too dark.

The correctly normalized version of (0x28, 0x28, 0x28) is (0x49, 0x49, 0x49). This PR replaces this value. For projects using F3D - F3DEX2, this will change the exported data, but won't change the in-game appearance because the vector gets normalized anyway.

### DIFF
--- a/fast64_internal/f3d/f3d_parser.py
+++ b/fast64_internal/f3d/f3d_parser.py
@@ -513,13 +513,13 @@ class F3DContext:
 
         # Here these are ints, but when parsing the values will be normalized.
         self.lights.l = [
-            Light([0, 0, 0], [0x28, 0x28, 0x28]),
-            Light([0, 0, 0], [0x28, 0x28, 0x28]),
-            Light([0, 0, 0], [0x28, 0x28, 0x28]),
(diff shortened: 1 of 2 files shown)


Lilaa3 (commented):
Obviously doesn't break anything outside of lighting, 0 visual difference, makes sense as the lack of normalization is kinda of a mistake.
I tested on vanilla and hackersm64 (this one matters very little since hacker has fixlights.py which makes lighting directional global) on ParaLLEl, GLideN64, Glide64, Rice and even gln64, I did not test jabo but I'm pretty sure it should be correct there too.
I'd say its ready to be merged as is.

Lilaa3 (approved):

PR #269, opened 2023-12-26
[OoT] Add Quick Import button

Dragorn421:
Add a "quick import" button, which allows to just put in a symbol name for it to be imported by fast64.

It can currently handle DLs, skeletons and animations

button location highlighted

operator props
operator props, boomerang DL
boomerang

Just Works :tm:

### DIFF
--- a/fast64_internal/oot/tools/quick_import.py
+++ b/fast64_internal/oot/tools/quick_import.py
@@ -1,0 +1,115 @@
+from pathlib import Path
+import os
+import re
+
+import bpy
+
+from ..f3d.properties import OOTDLImportSettings
+from ..skeleton.properties import OOTSkeletonImportSettings
+from ..animation.properties import OOTAnimImportSettingsProperty
+
+
+class QuickImportAborted(Exception):
+    def __init__(self, message):
+        super().__init__(message)
+        self.message = message
+
+
+def quick_import_exec(context: bpy.types.Context, sym_name: str):
+    sym_name = sym_name.strip()
+    if sym_name == "":
+        raise QuickImportAborted("No symbol name given")
+    if not all(
+        (
+            "a" <= c <= "z"
+            or "A" <= c <= "Z"
+            or "0" <= c <= "9"
+            or c
+            in {
+                "_",
+            }
+        )
+        for c in sym_name
+    ):
+        raise QuickImportAborted("Symbol names only have characters a-zA-Z0-9_")
+
+    sym_def_pattern = re.compile(rf"([^\s]+)\s+{sym_name}\s*(\[[^\]]*\])?\s*=")
+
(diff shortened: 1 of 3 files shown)


fast64_internal/oot/tools/quick_import.py
```diff
@@ -0,0 +1,115 @@
...
+        raise QuickImportAborted("Symbol names only have characters a-zA-Z0-9_")
+
+    sym_def_pattern = re.compile(rf"([^\s]+)\s+{sym_name}\s*(\[[^\]]*\])?\s*=")
+
+    base_dir_p = Path(context.scene.ootDecompPath)
+    assets_objects_dir_p = base_dir_p / "assets" / "objects"
+
+    all_found_defs: dict[Path, list[tuple[str, str]]] = dict()
+
+    for dirpath, dirnames, filenames in os.walk(assets_objects_dir_p):
```
Yanis42, line 43: looks like dirnames is unused, you can use ``_`` instead if you want to remove the unused thing in vscode (just a suggestion, I don't really mind if you keep the name)

fast64_internal/oot/tools/quick_import.py
```diff
@@ -0,0 +1,115 @@
...
+        raise QuickImportAborted(
+            f"Found several definitions of {sym_name} in {sym_file_p.relative_to(assets_objects_dir_p)}"
+        )
+
+    # We found a single definition of the symbol
+    sym_def_type, sym_def_array_decl = sym_defs[0]
+    is_array = sym_def_array_decl != ""
+    object_name = sym_file_p.relative_to(assets_objects_dir_p).parts[0]
+
+    if sym_def_type == "Gfx" and is_array:
```
Yanis42, line 83: if I want to implement more quick imports in the future I just need to add an if statement in this block? I just wanna avoid dealing with regex if I can lol

Yanis42 (commented):
just minor things, seems to work properly, nice feature!

fast64_internal/oot/tools/quick_import.py
Yanis42: (I'm talking about the ifs from line 83)

Dragorn421: Yeah the regex above matches any type and this if chain is what actually matches the type to know how to import it
You can try importing a texture or collision header for example, there will be an error message with the type of the symbol

fast64_internal/oot/tools/quick_import.py
Dragorn421: I'd rather keep it like this, these names are more or less standard usage of os.walk (copied straight from the docs) so this construct is more recognizable as is

Yanis42 (approved):
Looks good to me!

PR #270, opened 2023-12-31
Check if parent object exists

Yanis42:
should fix the mario.blend error


sauraen (approved):
Thanks! This is not a mario.blend error though. This is a simple bug from the recent OoT cutscene changes.

PR #271, opened 2024-01-01
Rendermode internal simplification

sauraen:
Like some of my other PRs, this started as a simple fix--make the fog controls show up under Sources if fog color or fog alpha are being used in the rendermode, even if fog is not enabled in the geometry mode. This turned into three main changes:

- fast64 now sets all the individual options for the advanced rendermode (while they're hidden) based on which rendermode presets are selected. This conversion happens at .blend file load and when changing the presets. This does not affect export (the strings for the presets are still written if the presets are used).
- Code dependent on the rendermode now is significantly simplified, because it only has to check the individual advanced rendermode settings, rather than have one codepath for that and another to extract the relevant bits from the selected presets.
- A GUI warning is added if the selected rendermode presets are not valid (exactly one must set the rendermode flags). The warnings about RDP silicon bugs are moved from being shown for advanced mode only, to also be shown when the presets are in use. The combination of these warnings will cover most cases of guiding users towards setting up the blender correctly.


fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+    return settings.blend_b1 == mix or (is_one_cycle and settings.blend_b2 == mix)
+
+
+def is_blender_lerp(
+    settings: "RDPSettings",
+    cycle: int,
+    p: str,
+    a: str,
+    m: str,
+    b: str,
```
Yanis42, line 274: idk a lot about this but what's the meaning of those letters here? if with the context it makes sense then ignore this, otherwise I'd like to have comments next to those parameters to say what it is

fast64_internal/oot/oot_upgrade.py
```diff
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
+import bpy
```
Yanis42, line 3: is this intended? looks like it's the only change in this file

Yanis42 (commented):
I still need to try this but code looks ok to me, I don't have a lot to say about it

fast64_internal/f3d/f3d_material.py
jesusyoshi54: these are the arguments used for the blending function. They're referred to with these letter designations in the programming manual

Dragorn421:
Good changes!

While checking the material presets to see if they're consistent with the warnings you added, I noticed
```
oot_shaded_multitexture_lerp_transparent_vertex_alpha
f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_OPA_SURF'
f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'

oot_shaded_texture_transparent_vertex_alpha
f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_OPA_SURF'
f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'

oot_shaded_vcol_multitexture_lerp_transparent_vertex_alpha
f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_OPA_SURF'
f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'

oot_shaded_vcol_texture_transparent_vertex_alpha
f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_OPA_SURF'
f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'
```

And unrelated but this is weird as well? `noop` makes `clr_in` 0 for the next cycle
```
oot_water_lerp_specular_fresnel
f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_NOOP'
f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'

oot_water_mult_specular_fresnel
f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_NOOP'
f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'
```

fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+            return default
+        return getattr(f3d, preset, default)
+    
+    is_one_cycle = settings.g_mdsft_cycletype == "G_CYC_1CYCLE"
+    if is_one_cycle:
+        r = get_with_default(settings.rendermode_preset_cycle_1, f3d.G_RM_AA_ZB_OPA_SURF)
+        r1 = r
+        # For GBL_c2, look at the same bits as for cycle 1. Cycle 1 bits are copied
+        # to cycle 2 bits at export.
+        r2 = r >> 2
```
Dragorn421, line 205: I feel like `r2 = r >> 2` is an unclear hack, and I've been writing more text explaining it than it takes to just make the set-blender-coefficients different based on cycle type

I suggest

```
        # In 1-cycle mode both cycles are the same
        r1 = r
        settings.blend_p1 = f3d.blendColorDict[(r1 >> 30) & 3]
        settings.blend_a1 = f3d.blendAlphaDict[(r1 >> 26) & 3]
        settings.blend_m1 = f3d.blendColorDict[(r1 >> 22) & 3]
        settings.blend_b1 = f3d.blendMixDict[(r1 >> 18) & 3]
        settings.blend_p2 = settings.blend_p1
        settings.blend_a2 = settings.blend_a1
        settings.blend_m2 = settings.blend_m1
        settings.blend_b2 = settings.blend_b1
```

and move the code below for setting pamb12 to inside the 2-cycle else

fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+    mix: str,
+    default_for_no_rendermode: bool = False
+) -> bool:
+    if not settings.set_rendermode:
+        return default_for_no_rendermode
+    is_one_cycle = settings.g_mdsft_cycletype == "G_CYC_1CYCLE"
+    return settings.blend_b1 == mix or (is_one_cycle and settings.blend_b2 == mix)
+
+
+def is_blender_lerp(
```
Dragorn421, line 268: why "lerp"?
If because you can't think of a name, huh, maybe
```suggestion
def is_blender_cycle_equal(
```
since it checks if a blender cycle equals the arguments?

fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+        "G_BL_CLR_FOG",
+        "G_BL_A_SHADE",
+        # While technically it being fog only requires that P and A be fog color
+        # and shade alpha, the only reasonable choice for M and B in this case
+        # is color in and 1-A.
+        "G_BL_CLR_IN",
+        "G_BL_1MA",
+        # if NOT setting rendermode, it is more likely that the user is setting
+        # rendermodes in code, so to be safe we'll enable fog
+        True
```
Dragorn421, line 304: you could make `default_for_no_rendermode` an argument to `is_blender_doing_fog` and move this comment to `update_fog_nodes`, where you may then also add that this is reasonable anyway since there's also the G_FOG check

fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+    )
+
+
+def get_blend_method(material: bpy.types.Material) -> str:
+    settings = material.f3d_mat.rdp_settings
+    if not settings.set_rendermode:
+        return drawLayerSM64Alpha[material.f3d_mat.draw_layer.sm64]
+    if settings.cvg_x_alpha:
+        return "CLIP"
```
Dragorn421, line 313: I have yet to entirely figure out the rdp state about alpha clip and blend (i.e. am not an expert), but I'd expect `g_mdsft_alpha_compare` to play a role more than cvg_x_alpha?

(more of a open-ended rdp question sorry)

Dragorn421:
for the merge conflicts, same suggestion as https://github.com/Fast-64/fast64/pull/280#issuecomment-1880070690

fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+    settings: "RDPSettings",
+    color: str,
+    default_for_no_rendermode: bool = False
+) -> bool:
+    if not settings.set_rendermode:
+        return default_for_no_rendermode
+    is_one_cycle = settings.g_mdsft_cycletype == "G_CYC_1CYCLE"
+    return (
+        settings.blend_p1 == color or settings.blend_m1 == color
+        or (is_one_cycle and (settings.blend_p2 == color or settings.blend_m2 == color))
```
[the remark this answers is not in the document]
sauraen: Good catch! I've made all the checks in the file be `is_two_cycle` now to avoid confusion / making this mistake again.

fast64_internal/oot/oot_upgrade.py
sauraen: Yes, this is a bugfix, `bpy` was already used in this file (not in my code, not related to the rest of this PR) but it was an error when that was executed.

fast64_internal/f3d/f3d_material.py
sauraen: This is not new code by me--this is the function `get_blend_method` originally at line 210. It has been refactored because it no longer needs to check all the details of the rendermode, instead directly using the RDP settings and a helper function.

fast64_internal/f3d/f3d_material.py
sauraen: So first of all, "lerp" is a programming term which means "linear interpolate", this is used by SGI in the N64 context as well as many other places.

I'm referring to the linear interpolation equation part of the blender, as opposed to the other settings within the rendermode. Now the other part of your question is why don't I call this "cycle". Other Fast64 code refers to the two of them as "cycle-dependent rendermode settings", so I guess the real name would be "the settings for one of the cycles within the cycle-dependent rendermode settings". To me "blender cycle" means the timestep within which the blender runs computations, not a subset of the settings for those computations.

With that said I agree the name is a bit unclear, do you have any other suggestions besides `is_blender_cycle_equal`? Maybe `is_blender_equation` or `is_blender_set_to_lerp`?

fast64_internal/f3d/f3d_material.py
```diff
@@ -177,85 +177,144 @@ def update_draw_layer(self, context):
...
+        # Already in advanced mode, don't overwrite this with the preset
+        return
+    
+    def get_with_default(preset, default):
+        # Use the default either if we are not setting rendermode, or if the
+        # preset is not in the GBI. We do want to set the advanced settings
+        # even if not setting rendermode, because they are read for nodes preview.
+        if not settings.set_rendermode:
+            return default
+        return getattr(f3d, preset, default)
```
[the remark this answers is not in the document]
sauraen: I don't think this was handled in a consistent way before this PR. I'm not really a fan of the previewing being affected by hidden properties in any case (including things like prim color), but if you feel strongly about this I'm happy to change it to be consistent. It won't affect the OoT world as materials virtually always set rendermode.

As far as the other question, there are plenty of presets which exist but are not in the GBI, such as all the RA presets--though admittedly I don't know how they're going to get into the preset field here.

fast64_internal/f3d/f3d_material.py
sauraen: Gonna have to disagree on this one--I strongly prefer one line of code which ends up being one instruction to 8 lines of code where four of them are repeats of the above. I've improved the comment though.

fast64_internal/f3d/f3d_material.py
sauraen: Correct, these are their names. P and M probably stand for plus and minus, but A and B are just arbitrary letters.

sauraen:
I apologize for what's going to be a delay--I see you have a couple more comments which didn't show up in the code viewer page for some reason, which I haven't addressed. I'm going to be out of town for a bit soon so it'll be a while before I can get back to this.

fast64_internal/f3d/f3d_material.py
sauraen: Done.

fast64_internal/f3d/f3d_material.py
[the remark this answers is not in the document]
sauraen: I've changed this to work the way you suggested, like prim color.

sauraen:
@Dragorn421 
> f3d_mat.rdp_settings.rendermode_preset_cycle_1 = 'G_RM_OPA_SURF'
> f3d_mat.rdp_settings.rendermode_preset_cycle_2 = 'G_RM_AA_ZB_XLU_SURF2'

From gbi.h:
```
#define RM_OPA_SURF(clk)                                        \
    CVG_DST_CLAMP | FORCE_BL | ZMODE_OPA |                      \
    GBL_c##clk(G_BL_CLR_IN, G_BL_0, G_BL_CLR_IN, G_BL_1)
```
So this is doing `G_RM_PASS` for the first cycle, `CVG_DST_CLAMP` and `ZMODE_OPA` are both 0 so they don't affect anything when being OR'd in, the only thing it actually sets is `FORCE_BL` which is also set by all `XLU_SURF` rendermodes. So practically it is fine, but theoretically it's wrong because it's two rendermodes which set the flags, I've changed all these to `G_RM_PASS` instead.

> noop makes clr_in 0 for the next cycle

No? It sets the four entries to "enum item 0" for each, which is `G_BL_CLR_IN`, `G_BL_A_IN`, `G_BL_CLR_IN`, `G_BL_1MA`. This is not the same as `G_BL_0` which has a value of 3. `RM_NOOP` is functionally equivalent to `RM_PASS` though not actually the same configuration.

Yanis42 (commented):
I reviewed again and tested, import works but fog preview has issues I think

Here's what Fast64 shows me with the intro setup (first cutscene header of hyrule field)
<details>

</details>

Here's how it is in-game:
<details>

</details>

I'm not sure if this was already broken or not but I noticed preview looked different so I wanted to try

Otherwise the code itself is ok to me

sauraen:
The difference between Fast64 and in game is present in main as well, so that won't be addressed in this PR. However, there was a bug where rendermode bits were not updated from rendermode presets when creating a material from a preset in code, which led to collision meshes not having fog on this branch. This has been fixed, now the fog appearance in Blender of an imported scene looks the same between main and this branch.

fast64_internal/f3d/f3d_material.py
Dragorn421: "equation" would be an improvement over lerp yeah
I knew about lerp = linear interpolation, didn't make the link with the blender equation since it can be configured in ways that don't look like lerping (e.g. RM_ADD)

I like `is_blender_equation_equal` ? but if you don't then `is_blender_equation` is fine to me too

sauraen: Changed to `is_blender_equation_equal`

Dragorn421 (approved):
@Yanis42 ?

Yanis42 (approved):
sorry for the delay

PR #272, opened 2024-01-03
[OoT] Refactor ``scene_table_c.py``

Yanis42:
I was too annoyed of people complaining about how the scene table has issues with HackerOoT because of how weird Fast64 handled that file, so basically started over from nothing

This new version is removing the "Enable HackerOoT Features" requirement and should keep any preprocessor things (ifdefs, includes, etc) and any comment (including multi-line ones), basically everything that isn't an actual scene entry

It should work properly but I haven't done extended testing yet and there's 2 minor issues right now (that I will fix soon™, also I need to cleanup a bit):
- custom scenes doesn't update the index in the comment before the entry properly
- exporting multiple times with a vanilla scene adds extra newlines

The benefits of this new code is improved readability and less lines (saved 26 ones)

Note: I know the spec export has minor issues (and I hate it) but that will be part of a dedicated PR


Dragorn421:
Thanks!

Bug reports below about scene numbering, but you seem to also have noticed one so probably will be fixed in one go

<details>

<summary>Deleting a scene breaks numbering</summary>

For example deleting deku tree:

```diff
diff --git a/include/tables/scene_table.h b/include/tables/scene_table.h
index 0337e1252..282dfda32 100644
--- a/include/tables/scene_table.h
+++ b/include/tables/scene_table.h
@@ -9,7 +9,6 @@
  *    - Argument 5: ? (Unknown)
  *    - Argument 6: ? (Unknown)
  */
-/* 0x00 */ DEFINE_SCENE(ydan_scene, g_pn_06, SCENE_DEKU_TREE, SDC_DEKU_TREE, 1, 2)
 /* 0x01 */ DEFINE_SCENE(ddan_scene, g_pn_08, SCENE_DODONGOS_CAVERN, SDC_DODONGOS_CAVERN, 1, 3)
 /* 0x02 */ DEFINE_SCENE(bdan_scene, g_pn_07, SCENE_JABU_JABU, SDC_JABU_JABU, 1, 4)
 /* 0x03 */ DEFINE_SCENE(Bmori1_scene, g_pn_01, SCENE_FOREST_TEMPLE, SDC_FOREST_TEMPLE, 2, 5)
```

then the first scene starts at 0x01

</details>

<details>

<summary>Added scenes start numbering one early</summary>

For example adding two scenes:

```diff
diff --git a/include/tables/scene_table.h b/include/tables/scene_table.h
index 0337e1252..0d2156a29 100644
--- a/include/tables/scene_table.h
+++ b/include/tables/scene_table.h
@@ -120,3 +120,5 @@
 /* 0x6B */ DEFINE_SCENE(hairal_niwa2_scene, g_pn_12, SCENE_HAIRAL_NIWA2, SDC_CASTLE_COURTYARD_GUARDS, 0, 0)
 /* 0x6C */ DEFINE_SCENE(sasatest_scene, none, SCENE_SASATEST, SDC_DEFAULT, 0, 0)
 /* 0x6D */ DEFINE_SCENE(testroom_scene, none, SCENE_TESTROOM, SDC_DEFAULT, 0, 0)
+// Added scenes
+/* 0x6D */ DEFINE_SCENE(mydefnewscen_scene, none, SCENE_MYDEFNEWSCEN, SDC_DEKU_TREE, 0, 0)
+/* 0x6E */ DEFINE_SCENE(againsnne_scene, none, SCENE_AGAINSNNE, SDC_DEKU_TREE, 0, 0)
```

for some reason the first one added uses 0x6D which is already SCENE_TESTROOM

</details>

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -1,283 +1,257 @@
...
+
+    specName: Optional[str] = None
+    titleCardName: Optional[str] = None
+    enumValue: Optional[str] = None
+    drawConfigIdx: Optional[str] = None
+    unk1: Optional[str] = None
+    unk2: Optional[str] = None
+
+    def __post_init__(self):
+        self.paramStart = "DEFINE_SCENE("
```
Dragorn421, line 30:
```suggestion
        macroStart = "DEFINE_SCENE("
```
seems like a better name?
and it can be just a variable

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -1,283 +1,257 @@
...
+        if self.original is not None and self.paramStart in self.original:
+            self.ogIdx = self.original.index(self.paramStart)
+            self.parsed = self.original[self.ogIdx + len(self.paramStart) :][:-1]
+
+            parameters = self.parsed.split(", ")
+            self.setParameters(parameters[0], parameters[1], parameters[2], parameters[3], parameters[4], parameters[5])
```
Dragorn421, line 36:
```suggestion
            self.setParameters(*parameters)
```

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -1,283 +1,257 @@
...
+            + f"{self.drawConfigIdx}, {self.unk1}, {self.unk2})\n"
+            + (self.suffix if self.suffix is not None else "")
+        )
+
+
+@dataclass
+class SceneTable:
+    exportPath: str
+    exportName: Optional[str]
+    selectedSceneEnumValue: str
```
Dragorn421, line 75:
```suggestion
    selectedSceneEnumValue: Optional[str]
```

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -1,283 +1,257 @@
...
-def getSceneIndex(sceneNameList, sceneName):
-    """Returns the index (int) of the chosen scene, returns None if ``Custom`` is chosen"""
-    if sceneName == "Custom":
-        return None
-
-    if sceneNameList is not None:
-        for i in range(len(sceneNameList)):
-            if sceneNameList[i] == sceneName:
+    def getIndexFromEnumValue(self):
+        """Returns the index (int) of the chosen scene, returns -1 if custom and -2 if insertion"""
```
Dragorn421, line 127: not a fan, please return some object instead, maybe
```py
class SomeSceneIndexType(enum.Enum):
    VANILLA = enum.auto()
    CUSTOM = enum.auto()
    INSERTION = enum.auto()

@dataclass
class SomeSceneIndex:
    type: SomeSceneIndexType
    index: Optional[int]  # only if type == SomeSceneIndexType.VANILLA
```

(I don't quite understand what "-2" is)

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
```diff
@@ -1,283 +1,257 @@
...
+    sceneEnumValues: list[str] = field(default_factory=list)  # existing values in ``scene_table.h``
+    isNewCustom: bool = False  # true if the custom scene doesn't exist in the table
+    firstAppend: bool = False
+    selectedSceneIndex: int = -3
+    customSceneIndex: int = 0
+
+    def __post_init__(self):
+        try:
+            with open(self.exportPath) as fileData:
+                # with open(self.exportPath, "r") as fileData:
```
Dragorn421, line 86:
```suggestion
```

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Yanis42: I didn't know you can do that

fast64_internal/oot/scene/exporter/to_c/scene_table_c.py
Yanis42: I used negative numbers since positive ones may be confused with actual indices

Yanis42:
there's still newline issues but the other ones should be fixed

Yanis42:
newline issues should be fixed now, I also found some issues related to exporting and removing custom scenes and they should be fixed too

I also added some comments and descriptions for the missing classes/functions (when the name's not self-explanatory at least)

I have hope in this being good to go

sauraen (approved):
Exporting a scene in HackerOoT works on this branch. I have not tried deleting scenes as I don't want to break my game.

Issue #252, opened 2023-09-20
Oot map importer segments are bugged


ariahiro64:
Oot map importer enables segments A B and C for all materials on import causing crashes without spending 30 minutes figuring out the correct settings by hand.

Yanis42:
can you be more specific? I never got a crash with the importer, ideally provide the errors in the console (but it can be annoying if it's crashing), are you importing a vanilla map when the issue occurs?

ariahiro64:
its when launching the map in game. fast64 doesnt error

ariahiro64:
update it seems to be assuming that everything on the draw config is used on all materials

as an example  in spot00 8 and 9 are used exclusively for xlu on the water and A is only used on a specific window to light it at night

any other use crashes in spot00

ariahiro64:
### Work around###
1. Import your map and fix cutscene related names
2. Uncheck **EVERY** segment in **EVERY** material
3. Apply only sensible ones based on the draw config

Dragorn421:
here is a python script for

> 2. Uncheck EVERY segment in EVERY material

```py
import bpy
for mat in bpy.data.materials:
    for attr in (*(f"segment{n}" for n in "89ABCD"), "customCall0", "customCall1"):
        for propGroup in (mat.ootMaterial.opaque, mat.ootMaterial.transparent):
            setattr(propGroup, attr, False)
```

ariahiro64:
> here is a python script for
> 
> > 2. Uncheck EVERY segment in EVERY material
> 
> ```python
> import bpy
> for mat in bpy.data.materials:
>     for attr in (*(f"segment{n}" for n in "89ABCD"), "customCall0", "customCall1"):
>         for propGroup in (mat.ootMaterial.opaque, mat.ootMaterial.transparent):
>             setattr(propGroup, attr, False)
> ```

that works

PR #274, opened 2024-01-05
Bump version to v2.2.0

Dragorn421:
Can also use this PR to discuss the release (e.g. if we should wait for anything)

Yanies mentioned it may be good to wait for f3dex3 preview, but since that's not anywhere near ready afaik I think it's fine to have a release in between

It's been 2 years since the previous release :sweat_smile: 

If you have permission (I'm not sure who does) there is a release draft for 2.2.0 on the releases page https://github.com/Fast-64/fast64/releases (not visible by everyone apparently)

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -40,7 +40,7 @@ from .fast64_internal.render_settings import (
 # info about add on
 bl_info = {
     "name": "Fast64",
-    "version": (2, 1, 0),
+    "version": (2, 2, 0),
     "author": "kurethedead",
     "location": "3DView",
     "description": "Plugin for exporting F3D display lists and other game data related to Nintendo 64 games.",


Lilaa3 (dismissed):
Sauren said the f3dex3 preview stuff is gonna take awhile.

Lilaa3:
Change-log looks perfect too!

ariahiro64:
there is a major bug with map importer for oot that applies the scene function to every single material causing almost guaranteed crashes. it is NOT ready for a release in this state.

ariahiro64:
#252 the work around is there for the issue so i guess you guys now know where to point people with issues

Dragorn421:
> there is a major bug with map importer for oot that applies the scene function to every single material causing almost guaranteed crashes. it is NOT ready for a release in this state.

Overlooking this because
- it is not a major bug (exporting imported stuff is inherently fragile, and also not the primary goal of fast64)
- though it's a pain you can fix the materials manually
- afaik this isn't a regression, the issue has existed since scene importing was implemented
- the 2.1.0 release already had scene importing (so, including the issue)

I am not trying to start an argument at all, just explaining why releasing despite this comment.

ariahiro64:
> > there is a major bug with map importer for oot that applies the scene function to every single material causing almost guaranteed crashes. it is NOT ready for a release in this state.
> 
> Overlooking this because
> 
>     * it is not a major bug (exporting imported stuff is inherently fragile, and also not the primary goal of fast64)
> 
>     * though it's a pain you can fix the materials manually
> 
>     * afaik this isn't a regression, the issue has existed since scene importing was implemented
> 
>     * the 2.1.0 release already had scene importing (so, including the issue)
> 
> 
> I am not trying to start an argument at all, just explaining why releasing despite this comment.

To add on we've also discussed it on the discord server and we feel just making people aware of the workaround is enough

ariahiro64 (approved):

Yanis42 (commented):
since we have an actual fix I'd like to see #275 merged before releasing 2.2.0, otherwise I don't have any objections

Dragorn421:
> since we have an actual fix I'd like to see #275 merged before releasing 2.2.0, otherwise I don't have any objections

I think it will take a bit more time to find a proper fix, so in agreement with Yanies we're going to release as is

PR #275, opened 2024-01-05
[OOT] Fix wrong segments on import

Yanis42:
Thanks to kenton and cdi-fails for figuring this out


fast64_internal/oot/oot_model_classes.py
```diff
@@ -389,15 +389,21 @@ def processDLName(self, name):
...
                 return None
-            return name
+            else:
+                for attr_name in dir(self.materialContext.ootMaterial.opaque):
+                    if re.match("segment[89ABCD]", attr_name):
+                        setattr(self.materialContext.ootMaterial.opaque, attr_name, False)
+                for attr_name in dir(self.materialContext.ootMaterial.transparent):
+                    if re.match("segment[89ABCD]", attr_name):
+                        setattr(self.materialContext.ootMaterial.transparent, attr_name, False)
+                return name
```
Dragorn421, line 399: would like an explanation

Yanis42: I have to admit, I didn't look at it 4Head

Yanis42: the loop is different tho, should I zip it?

Dragorn421 (commented):
I'm pretty sure this "fix" breaks this comment https://github.com/Fast-64/fast64/blob/b02373cc67b422c4c4405187298497f549795396/fast64_internal/oot/oot_model_classes.py#L423

dynamic material calls aka `self.materialContext.ootMaterial` is exactly what we want to clear under certain conditions (this PR implementing "clear every dlist jump", which is pretty rough imo) but this comment says they shouldn't be cleared ever for skeletons

ariahiro64:
last commit broke importing entirely

Yanis42:
I tried and works on my machine™

Dragorn421 (commented):
`OOTF3DContext.processDLName(name)` has the following purpose to hook into the handling of `gsSPDisplayList({name})`:
if name is a raw segment offset like `0x08000000`, don't try to find a dlist named "0x08000000" and instead mark the material as calling segment 8 (with `materialContext.ootMaterial.{opaque,transparent}.segment8 = True`)

(btw I don't know what is the issue this PR fixes)
The suggested fix here would mean a dlist like
```
gsSPDisplayList(0x08000000),
gsSPDisplayList(gSomeSymbolDL),
(vertices & triangles),
```
would be imported with the resulting material not calling segment 8, which sounds undesirable

Note that this:
```
gsSPDisplayList(gSomeSymbolDL),
gsSPDisplayList(0x08000000),
(vertices & triangles),
```
would have the imported material be set to call segment 8

Dragorn421:
This was attempting to fix #252 btw
Bit more context on fast64 discord: https://discord.com/channels/874816709855440926/874832163483287612/1192887072751616120

PR #276, opened 2024-01-06
Some f3d parser cleanup

Dragorn421:
I did not break importing OoT scenes so it must be fine 4Head


Yanis42 (commented):
idk much about f3d but code looks ok to me (I haven't tried this yet)

jesusyoshi54:
While testing this I found out that sm64 straight up does not work with the f3d importer. I am not surprised at all.

I then tested the f3d importer with some Kirby display lists that I exported and it worked just fine. So therefore I approve :)

jesusyoshi54 (approved):
It seems like sm64 didn't work before, or after this. That will be something to address in a different PR which is probably a long time coming for the f3d importer.

Dragorn421:
Thanks
I'd like to wait for #271 to be in before merging this in case it conflicts

Dragorn421:
merged main and #271, seems to be fine still :) will merge soon tm

PR #281, opened 2024-01-07
Name data in f3d_material_library.blend fast64-specific names

Dragorn421:
When exporting to gltf2 from blender, data from `f3d_material_library.blend` gets included into the export

Here's an example truncated gltf2 export of a scene where I put a cylinder, we can see data from `f3d_material_library.blend` inside YEP:

<details>

```json
{
	"asset":{
		"generator":"Khronos glTF Blender I/O v4.0.44",
		"version":"2.0"
	},
	"scene":0,
	"scenes":[
		{
			"name":"Scene",
			"nodes":[
				0
			]
		},
		{
			"name":"fast64_f3d_material_library_Scene",
			"nodes":[
				1
			]
		}
	],
	"nodes":[
		{
			"mesh":0,
			"name":"Cylinder"
		},
		{
			"mesh":1,
			"name":"fast64_f3d_material_library_PlaneObject"
		}
	],
	"meshes":[
		{
			"name":"Cylinder",
			"primitives":[
				{
					"attributes":{
						"POSITION":0,
						"NORMAL":1,
						"TEXCOORD_0":2
					},
					"indices":3
				}
			]
		},
		{
			"name":"fast64_f3d_material_library_PlaneMesh",
			"primitives":[
				{
					"attributes":{
						"POSITION":4,
						"NORMAL":5,
						"TEXCOORD_0":6
					},
					"indices":7
				}
			]
		}
	],
```

</details>

This can be mitigated by
1) disabling fast64 and unlinking the library manually
2) only exporting the active scene with gltf

However in practice currently only 1) works. 2) is typically broken because the default scene name is `Scene`, which most people don't change, and the fast64 library scene is named the same. Because of implementation details of the gltf2 exporter, this makes the f3d library scene be exported. A workaround could be to rename the scene to be exported to not-`Scene`

Anyway 2) is addressed by also renaming the scene to `fast64_f3d_material_library_Scene` in this PR

And in general renaming all this stuff means it shows up as "fast64_f3d_material_library_*" in the export, which is a good indication of where it comes from

Renames:

- fast64_f3d_material_library_Scene
- fast64_f3d_material_library_Collection
- fast64_f3d_material_library_PlaneObject
- fast64_f3d_material_library_PlaneMesh
- fast64_f3d_material_library_beefwashere

image

----------

Note: this doesn't seem to break previous blends nor creating materials in previous blends, but that's purely from my testing, I don't know anything about how blend libraries work to assess that's expected.


jesusyoshi54:
It seems to me that the reason this isn't a breaking change is because what is linked is the node tree data blocks, not the material itself. When a new f3d material is created, the material data-block is unique and the node tree from the library is linked, a node tree which is not renamed.

jesusyoshi54 (approved):

PR #282, opened 2024-01-07
[Cleanup] object selections and misc.

Dragorn421:
I was trying to do something else and fell into this

### DIFF
--- a/fast64_internal/utility.py
+++ b/fast64_internal/utility.py
@@ -157,18 +157,30 @@ def checkObjectReference(obj, title):
         )
 
 
-def selectSingleObject(obj: bpy.types.Object):
-    bpy.ops.object.select_all(action="DESELECT")
+def getActiveObject():
+    return bpy.context.view_layer.objects.active
+
+
+def setActiveObject(obj: bpy.types.Object):
     obj.select_set(True)
     bpy.context.view_layer.objects.active = obj
 
 
-def parentObject(parent, child):
-    bpy.ops.object.select_all(action="DESELECT")
+def deselectAllObjects():
+    for obj in bpy.data.objects:
+        obj.select_set(False)
+
+
+def selectSingleObject(obj: bpy.types.Object):
+    deselectAllObjects()
+    setActiveObject(obj)
+
+
+def parentObject(parent: bpy.types.Object, child: bpy.types.Object):
+    deselectAllObjects()
 
     child.select_set(True)
-    parent.select_set(True)
-    bpy.context.view_layer.objects.active = parent
+    setActiveObject(parent)
     bpy.ops.object.parent_set(type="OBJECT", keep_transform=True)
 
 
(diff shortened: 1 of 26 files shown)


fast64_internal/utility.py
```diff
@@ -157,18 +157,30 @@ def checkObjectReference(obj, title):
...
+
+
+def parentObject(parent: bpy.types.Object, child: bpy.types.Object):
+    deselectAllObjects()
 
     child.select_set(True)
-    parent.select_set(True)
-    bpy.context.view_layer.objects.active = parent
+    setActiveObject(parent)
     bpy.ops.object.parent_set(type="OBJECT", keep_transform=True)
```
Yanis42, line 184: what is this doing exactly? setting a parent object? if so then why not naming this ``setParentObject`` to avoid any confusion?

Yanis42 (commented):
ah yes, the cleanup rabbit hole

fast64_internal/utility.py
Dragorn421: yeah it's setting `child`'s parent to `parent`

the function name is to be understood as "parent the object" (the "to parent" verb)

but yeah I can change it to `setParentObject(child, parent)` it will be easier to understand

Yanis42: yea with the current name you could think it's doing something to an object's parent

Dragorn421: flink

Dragorn421:
merged main

fast64_internal/utility.py
```diff
@@ -164,18 +164,30 @@ def checkObjectReference(obj, title):
         )
 
 
-def selectSingleObject(obj: bpy.types.Object):
-    bpy.ops.object.select_all(action="DESELECT")
+def getActiveObject():
+    return bpy.context.view_layer.objects.active
```
Yanis42, line 168: is this ``bpy.types.Object`` by default or is it ``Any``? if it's not ``Object`` you should add the return type hint stuff (I mean this: ``def getActiveObject() -> bpy.types.Object``)

fast64_internal/utility.py
```diff
@@ -639,9 +651,7 @@ def highlightWeightErrors(obj, elements, elementType):
     return  # Doesn't work currently
```
Yanis42, line 651: not something you did but :thonk:

fast64_internal/utility.py
```diff
@@ -898,31 +913,26 @@ def get_obj_temp_mesh(obj):
...
 
 def duplicateHierarchy(obj, ignoreAttr, includeEmpties, areaIndex):
     # Duplicate objects to apply scale / modifiers / linked data
-    bpy.ops.object.select_all(action="DESELECT")
+    deselectAllObjects()
     selectMeshChildrenOnly(obj, None, includeEmpties, areaIndex)
-    obj.select_set(True)
-    bpy.context.view_layer.objects.active = obj
+    setActiveObject(obj)
     bpy.ops.object.duplicate()
```
Lilaa3, line 933: should there be an duplicateObj function too?

PR #284, opened 2024-01-09
[SM64] Combined Export Panel & Behavior Script Exporting

jesusyoshi54:
This PR aims to streamline exporting for SM64 all into one panel. This PR is made for C exclusively, and binary features should remain unchanged.

image

* Move Level, Geo, and Collision exporting all into the same panel
* Share properties for all types of exporting so that exports are more streamlined
* Exports of collision, geo layout objects and armatures all from a single button
* Exports of graphics data automatically write level script loads in the correct respective script.c file
* Exports of graphics data will also create a model ID define in model_ids.h
* Exports can have a separate defined collision and graphics object, or export from the shared selected object
* Support for exporting multiple objects at once. All exported objects will export using the name of the blend object
* Fixed level scripts erasing data from other scripts linked inside the existing previous level script

Along with all of the above improvements to exporting. I have also added in behavior script exporting.
Behavior scripts will be exported using the same properties as above, and they also have the option to inherit the following properties from the exported object:
* collision name
* model ID

Behavior scripts come with several preset behaviors and support adding, deleting and editing args of all the existing original sm64 behavior script cmds.


Lilaa3 (commented):
Export level SHOULD NOT be in the same panel as combined object exports, especially if the name of the panel is "SM64 Combined Object Exporter", the positioning of the level prop and the fact that this will cause confusion while exporting for both actor exports and level exports.

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+        else:
+            self.draw_level_path(layout)
+        layout.label(text=f"Geo Layout name will be: {self.geo_name}")
+        layout.label(text=f"Model ID will be: {self.model_id_define}")
+
+    def draw_obj_name(self, layout):
+        split_1 = layout.split(factor=0.55)
+        split_2 = split_1.split(factor=0.45)
+        split_2.label(text="Object Name")
+        split_2.prop(self, "object_name", text="")
```
Lilaa3, line 2027: "Object" in "Object Name" is redundant (small nitpick)

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+        if self.macro not in self._inheritable_macros:
+            return self.macro_args
+        if self.macro == "SET_MODEL":
+            return props.model_id_define
+        if self.macro == "LOAD_COLLISION_DATA":
+            return props.collision_name
+        return self.macro_args
+
+    def draw(self, layout, index):
+        box = layout.box()
```
Lilaa3, line 1504: .column() please

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+            export_path = f"{toAlnum(self.custom_export_path)}/"
+        else:
+            export_path = f"/levels/{toAlnum(self.level_name)}/"
+        layout.label(text=f"Level export location is: {export_path}")
+
+    def draw_gfx_names(self, layout):
+        if self.export_header_type == "Actor":
+            export_path = f"/actors/{toAlnum(self.obj_name_gfx)}/"
+            layout.label(text=f"Object export location is: {export_path}")
+            layout.label(text=f"This will also export to /actors/{toAlnum(self.obj_name_gfx)}(.c, .h, _geo.c)")
```
Lilaa3, line 2017: This label seems pretty irrelevant to me and it doesn´t even reuse export_path

Lilaa3:
I'm unsure if I like stuff like "Export behavior" and "Export collision" to be on by default
Let's say someone is just trying to do a mario export, they maybe watch a tutorial from 1 year ago for 2 minutes and then download fast64 and open up the mario blend and then oops! Behavior must have more than 0 commands?? What does that mean??? And oh no here comes 10 people asking about the same thing on discord!
You will have people exporting collision when they don´t want to on accident too. And this will happen even if you are exporting an armature. People with older files (and even with new files) will have to pay attention and turn off these toggles, it's pretty inconvenient. Just let the users toggle it themselves.

Lilaa3:
All behavior scripts should start with a BEGIN command, maybe don´t make the user add those manually and just add an enum for the object list? Also maybe adding a warning for when the behavior script doesn´t contain an END_LOOP, BREAK, DEACTIVATE, GOTO or RETURN could be a neat addition (only a warning because the user could have their own custom command which exits out of the behavior script). And if you are willing to step into feature creep territory, having toggles for common vanilla object flags would be very convenient. I don´t know what else could be done to help the user not mess up the behaviors.
Adding error checks for empty arguments could be another good addition but to be fair the user will just get a compilation error if they mess up their arguments so it is not really necessary

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+
+        elif self.export_header_type == "Actor":
+            prop_split(col, self, "group_name", "Group Name")
+            if self.group_name == "custom":
+                prop_split(col, self, "custom_export_name", "Group Name")
+
+        if self.export_col and not self.export_all_selected:
+            prop_split(col, self, "collision_object", "Collision Obj")
+        if self.export_gfx and not self.export_all_selected:
+            prop_split(col, self, "graphics_object", "Geo Layout Obj")
```
Lilaa3, line 2081: Is there a reason to abbreviate these?
Also fast64 uses "Geolayout" instead of "Geo Layout" everywhere else

fast64_internal/f3d/f3d_writer.py
```diff
@@ -407,13 +407,7 @@ def addCullCommand(obj, fMesh, transformMatrix, matWriteMethod):
...
     for vertexPos in obj.get("culling_bounds", obj.bound_box):
         fMesh.cullVertexList.vertices.append(
-            F3DVert(
-                Vector(vertexPos),
-                [0, 0],
-                Vector([0, 0, 0]),
-                None,
-                0.0,
-            ).toVtx(
+            F3DVert(Vector(vertexPos), [0, 0], Vector([0, 0, 0]), None, 0.0,).toVtx(
```
Lilaa3, line 410: Hey this might be the thing causing the black test to fail

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+    # load collision,  or set model ID are easy examples
+    inherit_from_export: bpy.props.BoolProperty(name="Inherit From Export")
+
+    _inheritable_macros = {
+        "LOAD_COLLISION_DATA",
+        "SET_MODEL",
+        # add support later maybe
+        # "SET_HITBOX_WITH_OFFSET",
+        # "SET_HITBOX",
+        # "SET_HURTBOX",
```
Lilaa3, line 1475: How would this work? Do you plan on adding an automatic hitbox and hurtbox or something? I do think we should have something like that for culling radius too, I have logic for an auto culling radius in an old branch of mine and it works great!

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+    # elif fast64_sig, insert after that line
+    # else insert after alt_condition
+    def find_export_lines(
+        self, file_lines, match_str=None, fast64_signature=None, alt_condition=None, start_delim=None, end_delim=None
+    ):
+        search_sig = False if start_delim else True
+        insert_line = 0
+        alt_insert_line = 0
+        match_line = 0
+        for j, line in enumerate(file_lines):
```
Lilaa3, line 1557: Is there any reason you use j over i for loops? Typically you use j for a second index and i for the first index

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+        else:
+            export_line = default_line + 1 if default_line else len(model_id_lines)
+            model_id_lines.insert(export_line, f"\n{fast64_sig}\n")
+            model_id_lines.insert(export_line + 1, export_model_id)
+
+        self.write_file_lines(model_ids, model_id_lines)
+
+    def export_group_script_load(self, script_path, props):
+        if not script_path.exists():
+            PluginError(f"could not find {script_path.stem}")
```
Lilaa3, line 1604: A lot of your plugin errors start with lower case, but some of yours and all others in fast64 use normal casing.

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+                self.execute_gfx(props, context, obj, index)
+                # do not export behaviors with multiple selection
+                if props.export_bhv and props.obj_name_bhv and not props.export_all_selected:
+                    self.export_behavior_script(context, props)
+        except Exception as e:
+            raisePluginError(self, e)
+            return {"CANCELLED"}
+
+        props.context_obj = None
+        # you've done it!~
```
Lilaa3, line 1843: Is this comment necessary?

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+            if "Geo" in obj.sm64_obj_type or obj.sm64_obj_type in {"None", "Switch"}:
+                actor_objs.append(obj)
+
+        return actor_objs
+
+    # writes collision.inc.c file, collision_header.h
+    # writes include into aggregate file in export location (leveldata.c/<group>.c)
+    # writes name to header in aggregate file location (actor/level)
+    # var name is: const Collision <props.col_obj>_collision[]
+    def execute_col(self, props, context, obj):
```
Lilaa3, line 1792: context is unused here

fast64_internal/f3d/f3d_writer.py
Dragorn421: absolutely is. just run `black .` to format everything, it really should Just Work :tm: , if it doesn't we want to figure out why. Formatting isn't meant to be a pain

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+            # pass on multiple export, throw on singular
+            if props.export_all_selected:
+                return
+            else:
+                raise Exception(e)
+
+    def execute(self, context):
+        props = context.scene.fast64.sm64.combined_export
+        try:
+            self.verify_context(context, props)
```
Lilaa3, line 1829: I'm personally not a huge fan of not just changing the context mode and then reverting it at the end of the operator, why should I not be able to export my geolayout if I'm in pose or edit mode of the armature, that's a bit dumb

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+    def execute_col(self, props, context, obj):
+        try:
+            if props.export_col and props.obj_name_col:
+                bpy.ops.object.sm64_export_collision(export_obj=obj.name)
+        except Exception as e:
+            # pass on multiple export, throw on singular
+            if props.export_all_selected:
+                return
+            else:
+                raise Exception(e)
```
Lilaa3, line 1801: Why not just do if not props.export_all_selected

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+        if self.obj_name_gfx and self.export_gfx:
+            self.draw_gfx_names(info_box)
+        else:
+            info_box.label(text=f"Gfx will not be exported until gfx obj is defined")
+
+        if self.obj_name_col and self.export_col:
+            info_box.label(text=f"Collision name will be: {self.collision_name}")
+            if self.export_rooms:
+                info_box.label(text=f"Rooms name will be: {self.collision_name}_rooms")
+
```
Lilaa3, line 2117: Extra new line here

fast64_internal/sm64/sm64_collision.py
```diff
@@ -383,6 +383,8 @@ def exportCollisionCommon(obj, transformMatrix, includeSpecials, includeChildren
     tempObj, allObjs = duplicateHierarchy(obj, None, True, areaIndex)
     try:
         addCollisionTriangles(tempObj, collisionDict, includeChildren, transformMatrix, areaIndex)
+        if not collisionDict:
+            raise PluginError("No collision data to export, col export cancelled")
```
Lilaa3, line 387: If there is no collision data to export and it comes out as an error I don´t think you need to mention that the collision export was cancelled.

fast64_internal/sm64/sm64_constants.py
```diff
@@ -1799,6 +2035,29 @@ def __init__(self, geoAddr, level, switchDict):
...
+    ("group10", "group10", "Peach/Yoshi"),
+    ("group11", "group11", "THI Ojbects (Lakitu, Wiggler, Bubba)"),
+    ("group12", "group12", "Bowser/Bowser Bomb"),
+    ("group13", "group13", "Water Objects (Skeeter, Treasure Chest etc.)"),
+    ("group14", "group14", "Ground Objects (Piranha Plant, Chain Chomp etc.)"),
+    ("group15", "group15", "Castle Objects (MIPS, Toad etc.)"),
+    ("group16", "group16", "Ice Objects (Chill Bully, Moneybags)"),
+    ("group17", "group17", "Cave Objects (Swoop, Scuttlebug, Dorrie etc.)"),
+    ("Custom", "Custom", "Custom"),
+]
```
Lilaa3, line 2060: I still think it would be nice to have the description include the levels in which these are loaded. 
For example, while group 4 includes water objects, it does not get loaded in wdw.
Also, group7 and group 16 get loaded in the exact same two levels, you should probably only include group 7

fast64_internal/sm64/__init__.py
```diff
@@ -222,6 +220,8 @@ class SM64_Properties(PropertyGroup):
     exportType: EnumProperty(items=enumExportType, name="Export Type", default="C")
     goal: EnumProperty(items=sm64GoalTypeEnum, name="Export Goal", default="All")
 
+    combined_export: bpy.props.PointerProperty(type=SM64_CombinedObjectProperties)
+
     # TODO: Utilize these across all exports
```
Lilaa3, line 225: You should probably remove that TODO

fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -3333,21 +3240,12 @@ def sm64_geo_writer_register():
...
     bpy.types.Scene.geoInsertableBinaryPath = bpy.props.StringProperty(name="Filepath", subtype="FILE_PATH")
     bpy.types.Scene.geoIsSegPtr = bpy.props.BoolProperty(name="Is Segmented Address")
-    bpy.types.Scene.geoName = bpy.props.StringProperty(name="Directory Name", default="mario")
-    bpy.types.Scene.geoGroupName = bpy.props.StringProperty(name="Name", default="group0")
-    bpy.types.Scene.geoExportHeaderType = bpy.props.EnumProperty(
-        name="Header Export", items=enumExportHeaderType, default="Actor"
-    )
-    bpy.types.Scene.geoCustomExport = bpy.props.BoolProperty(name="Custom Export Path")
-    bpy.types.Scene.geoLevelName = bpy.props.StringProperty(name="Level", default="bob")
-    bpy.types.Scene.geoLevelOption = bpy.props.EnumProperty(items=enumLevelNames, name="Level", default="bob")
```
Lilaa3, line 3350: Since you are removing these, you need some upgrade code.

fast64_internal/sm64/sm64_objects.py
```diff
@@ -14,6 +17,16 @@
     all_values_equal_x,
     checkIsSM64PreInlineGeoLayout,
     prop_split,
+    multilineLabel,
+    applyRotation,
+    applyBasicTweaks,
+    getPathAndLevel,
```
Lilaa3, line 23: Unused imports

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+
+        self.write_file_lines(script_path, script_lines)
+
+    def export_level_specific_load(self, script_path, props):
+        if not script_path.exists():
+            PluginError(f"could not find {script_path.stem}")
+        script_lines = open(script_path, "r").readlines()
+
+        # place model load into custom level script array
+        script_load = f"    LOAD_MODEL_FROM_GEO({props.model_id_define}, {props.geo_name}),\n"
```
Lilaa3, line 1642: Why not \t?

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -296,72 +299,100 @@ def new(cls):
...
-            + "\tCLEAR_LEVEL(),\n"
-            + "\tSLEEP_BEFORE_EXIT(1),\n"
-            + "\tEXIT(),\n};\n"
+            mario_start_macro = "\tMARIO_POS(1, 0, 0, 0, 0),"
+        return "\n".join(
+            filter(
+                None,
+                (
+                    # all the includes
+                    "#include <ultra64.h>" '#include "sm64.h"',
```
Lilaa3, line 360: you missed a comma there

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -482,26 +513,18 @@ def stringToMacros(data):
...
 
-def macroToString(macroCmd, useComma):
-    result = macroCmd[0] + "("
-    for arg in macroCmd[1]:
-        result += arg + ", "
-    result = result[:-2] + ")" + ("," if useComma else "")
-    result += " " + macroCmd[2]
-    return result
+def macroToString(macro_cmd):
+    return f"{macro_cmd.function}({', '.join(macro_cmd.args)}), {macro_cmd.comment}"
```
Lilaa3, line 522: Adding a comma here always causes the game to not compile, DEFINE_LEVEL should not have a comma

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -697,279 +737,272 @@ def __init__(self):
...
-                if f == "area_" + str(index):
+            for index, areaObj in area_dict.items():
+                if folder == f"area_{index}":
                     existingArea = True
             if not existingArea:
-                shutil.rmtree(os.path.join(levelDir, f))
+                shutil.rmtree(os.path.join(level_dir, folder))
+
+    def include_proto(file_name):
+        return f'#include "levels/{level_name}/{file_name}/"\n'
```
Lilaa3, line 936: there should not be a / after the file name (aside from being weird it causes a compilation error so there is that)

fast64_internal/f3d/f3d_writer.py
jesusyoshi54: I did and it didn't format this. Not sure why

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: habits from electrical studies no real reason

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: No

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: I am not sure on the backend if that causes an issue but as far as this PR goes I am not concerned with editing the internals of export operators much, which this restriction is carried over from

fast64_internal/sm64/sm64_constants.py
```diff
@@ -1769,6 +1769,242 @@ def __init__(self, geoAddr, level, switchDict):
...
+    ("ADD_INT", "ADD_INT", "Adds int value to obj field"),
+    ("SET_INT", "SET_INT", "Set int value to obj field"),
+    ("OR_INT", "OR_INT", "Bitwise OR int value to obj field"),
+    ("BIT_CLEAR", "BIT_CLEAR", "Bitwise Clear int value to obj field"),
+    ("SET_INT_RAND_RSHIFT", "SET_INT_RAND_RSHIFT", "Set random float value to obj field with min and R shifted"),
+    ("SET_RANDOM_FLOAT", "SET_RANDOM_FLOAT", "Set random float value to obj field with min"),
+    ("SET_RANDOM_INT", "SET_RANDOM_INT", "Set random int value to obj field with min"),
+    ("ADD_RANDOM_FLOAT", "ADD_RANDOM_FLOAT", "Add random int value to obj field with min"),
+    ("ADD_INT_RAND_RSHIFT", "ADD_INT_RAND_RSHIFT", "Add random int value to obj field with min and R shifted"),
+    ("CMD_NOP_1", "CMD_NOP_1", "No Operation"),
```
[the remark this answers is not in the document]
jesusyoshi54: I support the full sm64 feature set, maybe that is not important anymore but I think it is good to do.

fast64_internal/sm64/sm64_geolayout_writer.py
jesusyoshi54: afaik they're unused so the props will just be vestigial remnants of old classes, and since they aren't called they won't error out

fast64_internal/sm64/sm64_objects.py
Lilaa3: true

fast64_internal/sm64/sm64_objects.py
Lilaa3: it does not cause any significant issues no, in fact my settings pr does it for the tool operators, works great

fast64_internal/sm64/sm64_constants.py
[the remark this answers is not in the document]
Lilaa3: Fine with me, hacker doesn´t remove this btw, it does replace NOP_3 tho but you don´t allow for that one anyways

jesusyoshi54:
> All behavior scripts should start with a BEGIN command
Not necessarily true. Behaviors can exist as partial scripts that are jumped to as an immediate example, and also for objects in the default list the BEGIN cmd is redundant. If the user wants to set up a prototypical behavior quickly they can use a preset, such as a the solid object one.

> add an enum for the object list?
> Also maybe adding a warning for when the behavior script doesn´t contain an END_LOOP, BREAK, DEACTIVATE, GOTO or RETURN could be a neat addition (only a warning because the user could have their own custom command which exits out of the behavior script).

Good ideas I will incorporate these 


> And if you are willing to step into feature creep territory, having toggles for common vanilla object flags would be very convenient.

I am not sure this would work well while not bloating the UI and also being flexible.


> I don´t know what else could be done to help the user not mess up the behaviors. Adding error checks for empty arguments could be another good addition but to be fair the user will just get a compilation error if they mess up their arguments so it is not really necessary

I don't believe in excessively controlling user output. It should just be what they type, otherwise it may be inflexible.

fast64_internal/sm64/sm64_level_writer.py
Lilaa3: This is a major issue that you should address first

fast64_internal/sm64/sm64_level_writer.py
Lilaa3: So is this

fast64_internal/f3d/f3d_writer.py
Dragorn421: Can you run `black . --verbose` and share the output

fast64_internal/sm64/sm64_objects.py
```diff
@@ -2042,57 +2040,61 @@ def draw_bhv_options(self, layout):
...
-        layout.separator()
-        # object exports
         col = layout.column()
+        box = col.box()
+        box.operator("object.sm64_export_level", text="Export Level")
+        prop_split(box, self, "level_name", "Level")
+        if self.level_name == "custom":
+            prop_split(box, self, "custom_export_name", "Level Name")
+        self.draw_level_path(box.box())
+        box.separator()
```
Lilaa3, line 2050:
```
        col = layout.column()

        # level exports
        box = col.box().column()
        box.operator("object.sm64_export_level", text="Export Level")
        prop_split(box, self, "level_name", "Level")
        if self.level_name == "custom":
            prop_split(box, self, "custom_export_name", "Level Name")
        self.draw_level_path(box.box())
 
        col.separator()

```
How about this instead?

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1421,719 @@ def draw_act(self, obj, layout, value):
...
+    arg_3: bpy.props.StringProperty(name="Argument 3")
+    # for gravity, 8 args are used, but everything else only uses a max of 3
+    arg_4: bpy.props.StringProperty(name="Argument 4")
+    arg_5: bpy.props.StringProperty(name="Argument 5")
+    arg_6: bpy.props.StringProperty(name="Argument 6")
+    arg_7: bpy.props.StringProperty(name="Argument 7")
+    arg_8: bpy.props.StringProperty(name="Argument 8")
+
+    # some objects have cmds that make sense to dynamically inherit it from export properties
+    # load collision,  or set model ID are easy examples
```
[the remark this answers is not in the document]
Lilaa3:
```suggestion
    # load collision, or set model ID are easy examples
```

fast64_internal/sm64/sm64_level_writer.py
```diff
@@ -124,12 +127,7 @@ def __init__(self, masks, originalData):
         self.originalData = originalData
 
     def to_c(self):
-        result = ""
-        # result += 'u8 sZoomOutAreaMasks[] = {\n'
-        result += "\n"
```
Lilaa3, line 129: Please add the new line back in the write function for zoom out masks, doesn´t break anything, looks better and there is no reason to cause unnecessary changes like that
image

jesusyoshi54:
I have made the following improvements:

* Added enum for object lists in behaviors
* Added object field enum in behaviors
* Added checkbox to not export model ID and script load, default is False
* Set default model ID to 0xE2, which is the first unoccupied model ID in vanilla SM64
* Small fixes in regards to exporting multiple objects

fast64_internal/sm64/__init__.py
```diff
@@ -228,13 +228,17 @@ def upgrade_changed_props():
...
-        if hasattr(bpy.context.scene, "geoName"):
+        if bpy.context.scene.get("geoName", None):
             combined_props = bpy.context.scene.fast64.sm64.combined_export
-            combined_props.export_header_type = bpy.context.scene.geoExportHeaderType
-            combined_props.custom_export_name = bpy.context.scene.geoLevelName
-            combined_props.custom_export_path = bpy.context.scene.geoExportPath
-            combined_props.level_name = bpy.context.scene.geoLevelOption
-            combined_props.object_name = bpy.context.scene.geoName
+            old_scene_props_to_new = {
+                "geoExportHeaderType": "export_header_type",
```
Lilaa3, line 234: Won´t work, enumss are stored as ints when not registered, see how I do this in my settings rework

fast64_internal/sm64/__init__.py
```diff
@@ -228,13 +228,17 @@ def upgrade_changed_props():
             bpy.context.scene.fast64.sm64.exportType = get_legacy_export_type()
             bpy.context.scene.fast64.sm64.version = SM64_Properties.cur_version
         # props upgrade for combined export panel
-        if hasattr(bpy.context.scene, "geoName"):
+        if bpy.context.scene.get("geoName", None):
```
Lilaa3, line 231: If you are already have defaults for all the props this is very much unimportant and again, I'm pretty sure that unregistered default values are always going to be None so honestly just get rid of this check I guess, and another thing, you aren´t changing the version so this will run after all file loads, lets not do that.

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+        if self.export_bhv and not self.export_all_selected:
+            self.draw_bhv_options(col)
+
+        # info/warnings
+        if self.export_header_type == "Custom":
+            info_box = box.box()
+            info_box.label(text="Export will not write any headers or dependencies", icon="ERROR")
+
+        if self.export_all_selected:
+            info_box = box.box()
```
Lilaa3, line 2161: Should you not use .column() here?

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+                info_box,
+                text="Object name used will be the name of respective selected objects.\n"
+                "Objects will export based on root of parenting hierarchy.\n"
+                "Model IDs will export in order starting from chosen Model ID Num.\n"
+                "Behaviors will not export\n"
+                "Duplicates objects will be exported! Use with Caution.\n",
+                icon="ERROR",
+            )
+
+        info_box = box.box()
```
Lilaa3, line 2172: same here

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+    def draw_bhv_options(self, layout):
+        if self.export_all_selected:
+            return
+        box = layout.box()
+        prop_split(box, self, "preset_behavior_script", "Preset Behavior Script")
+        box.operator("bone.add_behavior_script", text="Add Behavior Cmd").option = len(self.behavior_script)
+        for index, bhv in enumerate(self.behavior_script):
+            bhv.draw(box, index)
+
+    def draw(self, layout):
```
Lilaa3, line 2106: For consistency with oot, maybe this should be renamed to draw_props
> to avoid any confusion with a native draw function blender can have (yanis when we talked about the naming of these functions)

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+    def draw_obj_name(self, layout):
+        split_1 = layout.split(factor=0.45)
+        split_2 = split_1.split(factor=0.45)
+        split_2.label(text="Name")
+        split_2.prop(self, "object_name", text="")
+        if bpy.context.active_object:
+            tmp_obj_name = bpy.context.active_object.name
+        else:
+            tmp_obj_name = ""
+        split_1.label(text=f"or {repr(tmp_obj_name)} if no name")
```
Lilaa3, line 2095: if bpy.context.active_object:
```suggestion
        if bpy.context.active_object:
            tmp_obj_name = bpy.context.active_object.name
            split_1.label(text=f"or {repr(tmp_obj_name)} if no name")
```

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+
+    def execute(self, context):
+        prop = context.scene.fast64.sm64.combined_export
+        prop.behavior_script.remove(self.option)
+        self.report({"INFO"}, "Success!")
+        return {"FINISHED"}
+
+
+class BehaviorScriptProperty(bpy.types.PropertyGroup):
+    expand: bpy.props.BoolProperty(name="Expand")
```
Lilaa3, line 1449: Default to true would be better in my opinion

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
         layout.prop(obj, "sm64_obj_use_act" + str(value), text="")
 
 
+class AddBehavior(bpy.types.Operator):
+    bl_idname = "bone.add_behavior_script"
```
Lilaa3, line 1424: I know it doesn´t matter functionality wise but... bone?

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+    def draw_level_path(self, layout):
+        if self.export_header_type == "Custom":
+            export_path = f"{toAlnum(self.custom_export_path)}/"
+        else:
+            export_path = f"/levels/{toAlnum(self.level_name)}/"
+        layout.label(text=f"Level export location is: {export_path}")
+
+    def draw_gfx_names(self, layout):
+        if self.export_header_type == "Actor":
+            layout.label(text=f"Objects will export to /actors/{toAlnum(self.obj_name_gfx)}(.c, .h, _geo.c)")
```
Lilaa3, line 2080: No, it exports to the name of the mesh of the armature, the mario default blend exports to "skinned" (and the geo ends up being named skinned_geo and for some reason the collision exports to the actual armature name).
The upgrade code doesn´t and can´t account for defaults without being an annoyance, so maybe the mario blend needs a quick update for this pr too. (although the group not defaulting to group0 is a bit odd regardless)

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+        default=False,
+        description="Export collision for linked or selected mesh that have collision data",
+    )
+    export_gfx: bpy.props.BoolProperty(
+        name="Export Graphics",
+        default=False,
+        description="Export geo layouts for linked or selected mesh that have collision data",
+    )
+    export_script_loads: bpy.props.BoolProperty(
+        name="Export Script Loads",
```
Lilaa3, line 1950: is export the right name for this, its more "updating" than an export no?

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,784 @@ def draw_act(self, obj, layout, value):
...
+            box.prop(self, "include_children")
+            box.prop(self, "export_rooms")
+        else:
+            col.prop(self, "export_col")
+        split.prop(self, "export_gfx")
+        if not self.export_all_selected:
+            split.prop(self, "export_bhv")
+        row = layout.row()
+        row.prop(self, "export_all_selected")
+        row.prop(self, "export_script_loads")
```
Lilaa3, line 2068: Shouldn´t this toggle be put in the graphics column like you do for collision toggles

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: It was what warps did iirc

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: Generally speaking you're appending to an existing script which isn't the same but I think a common understanding of appending with new data is an "export". Also considering it is the nomenclature used for all the other operations done I would say it is good.

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: box is a valid sublayout that defines vertical orientation. There is no need to add in an extra `.column()`

> box()[](https://docs.blender.org/api/current/bpy.types.UILayout.html#bpy.types.UILayout.box)
> Sublayout (items placed in this sublayout are placed under each other in a column and are surrounded by a box)
> 
> Returns: Sub-layout to put items in
> 
> Return type: [UILayout](https://docs.blender.org/api/current/bpy.types.UILayout.html#bpy.types.UILayout)

fast64_internal/sm64/sm64_objects.py
Lilaa3: still, shouldn´t it be named scene.add_behavior_script instead?

jesusyoshi54:
yeah I'm not against renaming it. I just added context so I seemed like a
normal programmer.
I generally close the problems I won't address/already have.

fast64_internal/sm64/sm64_objects.py
jesusyoshi54: When adding in a preset behavior script, which I think will be the normal usage, it will add in like 8 or so cmds at once all open. So having it default closed will be better to avoid bloat. If you use the operator to add in a new cmd I will instead make it so that single new one is open.

Lilaa3: That sounds good to me

jesusyoshi54:
The latest commit should address all the review concerns and also added extra control over object naming.

jesusyoshi54:
I have no plans to address a change not yet implemented in a specific repo. All features of this panel can be enabled/disabled independently, if a specific repo does not work with a feature, they can turn it off.
If you turn off the `Export Script Loads` feature, there will be no model ID or level script export, and only the `geo.inc.c` and `collision.inc.c` files will be exported, as would be done with the current existing panel.

If there is support for specific repos and versions of repos I may address this but I don't believe I have any measure currently to handle this.

fast64_internal/sm64/sm64_objects.py
Lilaa3: Unresolving this, but also I personally do this in these kinds of exceptions as much as possible, in here it may seem less useful but having a full traceback is usually good, pylint warns for it
```suggestion
        except Exception as exc:
            # pass on multiple export, throw on singular
            if not props.export_all_selected:
                raise Exception(exc) from exc
```

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,808 @@ def draw_act(self, obj, layout, value):
...
+
+        # pathing for gfx/col exports
+        prop_split(box, self, "export_header_type", "Export Type")
+
+        if self.export_header_type == "Custom":
+            prop_split(box, self, "custom_export_path", "Custom Path")
+
+        elif self.export_header_type == "Actor":
+            prop_split(box, self, "group_name", "Group Name")
+            if self.group_name == "custom":
```
Lilaa3, line 2171: Custom not custom

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,808 @@ def draw_act(self, obj, layout, value):
...
+
+    export_header_type: bpy.props.EnumProperty(
+        name="Header Export",
+        items=[*enumExportHeaderType, ("Custom", "Custom", "No headers are written")],
+        default="Actor",
+    )
+    # level export header
+    level_name: bpy.props.EnumProperty(items=enumLevelNames, name="Level", default="bob")
+    # actor export header
+    group_name: bpy.props.EnumProperty(name="Group Name", default="group0", items=groups_obj_export)
```
Lilaa3, line 1922: How about just Group for the name, since this is an enum and I heavily dislike this kind of repetition personally
image

fast64_internal/sm64/sm64_objects.py
Lilaa3: image
Scutte please.. imagine
image

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,808 @@ def draw_act(self, obj, layout, value):
...
+                prop_split(box, self, "custom_export_name", "Group Name")
+        else:
+            box.label(text="Destination level selection is shared with level export dropdown", icon="PINNED")
+            prop_split(box, self, "level_name", "Level")
+        if self.export_col and not self.export_all_selected:
+            prop_split(box, self, "collision_object", "Collision Obj")
+        if self.export_gfx and not self.export_all_selected:
+            prop_split(box, self, "graphics_object", "Geo Layout Obj")
+        if self.export_gfx and self.export_script_loads:
+            prop_split(box, self, "model_id", "Model ID Num")
```
Lilaa3, line 2181: You should try including these in their respective export boxes, looks really good imo

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,808 @@ def draw_act(self, obj, layout, value):
...
+        else:
+            split.prop(self, "export_col")
+        if self.export_gfx:
+            box = split.box()
+            box.prop(self, "export_gfx")
+            box.prop(self, "export_script_loads")
+        else:
+            split.prop(self, "export_gfx")
+        if not self.export_all_selected:
+            split.prop(self, "export_bhv")
```
Lilaa3, line 2089: Another ui suggestion, have you tried a row with align and then using toggle=1 for the export_ props? I tried it and it looks very clean imo, oot does it a lot and i love it (also using .column())
image

Lilaa3:
bump https://github.com/Fast-64/fast64/pull/284/#discussion_r1445903196

fast64_internal/sm64/sm64_objects.py
```diff
@@ -1403,6 +1420,808 @@ def draw_act(self, obj, layout, value):
...
+        decomp_path = Path(bpy.path.abspath(bpy.context.scene.decompPath))
+        behavior_data = decomp_path / "data" / "behavior_data.c"
+        if not behavior_data.exists():
+            PluginError("could not find behavior_data.c")
+
+        # add at top of bhvs, 3 lines after this is found
+        bhv_data_lines = open(behavior_data, "r").readlines()
+        export_bhv_name = f"const BehaviorScript {props.bhv_name}[] = {{\n"
+        last_bhv_define = "#define SPAWN_WATER_DROPLET(dropletParams)"
+        fast64_sig = "// fast64 object exports get inserted here"
```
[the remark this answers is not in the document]
jesusyoshi54: There is no benefit code wise as far as I know, and I feel like it would be more likely to cause issues if you only did parsing inside a block instead of the whole file. If for example you want to replace an existing model name, and only parse in a block, you would just redefine the model instead of searching the entire file, or insert something outside the block which may confuse users.

Reonu (changes requested):
You forgot to add `common1` to `groups_obj_export`.

Also, the `Custom` option doesn't actually work. It ignores that you type in, and tries to export to a group called `Custom` instead.

fast64_internal/sm64/sm64_constants.py
```diff
@@ -1792,6 +2114,29 @@ def __init__(self, geoAddr, level, switchDict):
     ("Custom", "Custom", "Custom"),
 ]
 
+# groups you can use for the combined object export
+groups_obj_export = [
```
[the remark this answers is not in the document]
Lilaa3: @reonu common1 should not be included

Reonu (approved):
I've been using this PR extensively and everything seems to work well!

fast64_internal/sm64/sm64_collision.py
```diff
@@ -608,41 +604,17 @@ class SM64_ExportCollisionPanel(SM64_Panel):
     bl_label = "SM64 Collision Exporter"
     goal = "Object/Actor/Anim"
 
+    @classmethod
+    def poll(cls, context):
+        return context.scene.fast64.sm64.export_type != "C"
```
Lilaa3, line 609: just overriding the poll doesn´t work and will cause sm64 panels to show in oot, tested this just now by changing to binary and then switching game modes

fast64_internal/sm64/settings/properties.py
```diff
@@ -115,9 +117,28 @@ def upgrade_version_1(self, scene: Scene):
...
+            setattr(combined_props, new, scene.get(old, getattr(combined_props, new)))
+        export_type = scene.get("geoExportHeaderType", None)
+        if export_type is not None:
+            combined_props.export_header_type = enumExportHeaderType[export_type][0]
+
+        level_name = scene.get("geoLevelOption", None)
+        if level_name is not None:
+            combined_props.level_name = enumLevelNames[level_name][0]
+        self.version = 3
+
```
Lilaa3, line 141: When you update to main you can use the upgrade_old_prop function i added, it has the multiple prop get the first one that used to be set functionality
for example
`upgrade_old_prop(self, "object_name", scene, {"geoLevelName", "colLevelName", "animLevelName"})`

Lilaa3 (changes requested):
Went over all the code again, aside from updating to main these are the last things (I'm sorry for dragging this out but I swear I don´t think I missed anything this time, I re-read all the changes and see no issues and I tested all types of exports one more time, so after this I will accept and any leftover issues will be on me lol)

fast64_internal/sm64/settings/properties.py
```diff
@@ -115,6 +123,30 @@ def upgrade_changed_props():
...
+                "level_name",
+                scene,
+                {
+                    "geoLevelOption",
+                    "colLevelOption",
+                    "animLevelOption",
+                },
+            )
+            for old, new in old_export_props_to_new.items():
+                upgrade_old_prop(combined_props, new, scene, old)
```
Lilaa3, line 149:
```suggestion
            for new, old in old_export_props_to_new.items():
                upgrade_old_prop(combined_props, new, scene, old)
```

fast64_internal/sm64/settings/properties.py
```diff
@@ -96,6 +98,12 @@ def upgrade_changed_props():
             "refreshVer": "refresh_version",
             "exportType": "export_type",
         }
+        old_export_props_to_new = {
+            "geoLevelName": "custom_export_name",
+            "geoExportPath": "custom_export_path",
+            "geoName": "object_name",
+            "geoGroupName": "group_name",
+        }
```
Lilaa3, line 106:
```suggestion
        old_export_props_to_new = {
            "custom_export_name": {"geoLevelName", "colLevelName", "animLevelName"},
            "custom_export_path": {"geoExportPath", "colExportPath", "animExportPath"},
            "object_name": {"geoName", "colName", "animName"},
            "group_name": {"geoGroupName", "colGroupName", "animGroupName"},
            "level_name": {"geoLevelName", "colLevelName", "animLevelName"},
            "export_header_type": {"geoExportHeaderType", "colExportHeaderType", "animExportHeaderType"},
        }
```

Lilaa3 (changes requested):
Small quick thing, see if you agree (everything else works)

Lilaa3 (approved):

Lilaa3:
@Reonu im gonna merge in like 2 days max, if you wanna test anymore do it soon please

Reonu:
Since my last comment I've tested exporting behaviors and collision, it all worked fine. My only complaint is that the behavior scripts exported by fast64 are indented differently than the vanilla scripts (no extra indentation after BEGIN_LOOP)

Lilaa3:
I don´t see an issue with that honestly, fast64 doesn´t always stick to what vanilla does for identation

gheskett:
It matters more for behavior script exports than geos because people actually still have good reasons to interface with the behavior script file directly. Geos are exported to their own files and are less necessary to adhere to matching formatting.

Lilaa3:
@jesusyoshi54 do you wanna address this before merging?

jesusyoshi54:
image
It is done

Reonu (approved):
based!

Lilaa3 (approved):
tested and it works even with broken behaviors (multiple end loops for example)

Reonu:
I was about to tell you to merge this and then I realized it was already merged 

Then I was going to tell you to fix wiseguy decals but then I realized you already did

Thank you lila

PR #286, opened 2024-01-09
Fix OoT UI showing in SM64

Yanis42:
oops


Lilaa3 (approved):
Fix works as intended

PR #288, opened 2024-01-09
[OoT] Default ``fromRoom`` to the parented room is it's parented to a room

Yanis42:
This is a feature request I tried to implement, it's not great and the main limitation is the active object requirement, too bad you can't set default values for ``PointerProperty``...

This is a draft until I find a solution I'm happy with, feel free to make suggestions

### DIFF
--- a/fast64_internal/oot/actor/properties.py
+++ b/fast64_internal/oot/actor/properties.py
@@ -1,3 +1,5 @@
+import bpy
+
 from bpy.types import Object, PropertyGroup, UILayout
 from bpy.utils import register_class, unregister_class
 from bpy.props import EnumProperty, StringProperty, IntProperty, BoolProperty, CollectionProperty, PointerProperty
@@ -165,7 +167,22 @@ class OOTTransitionActorProperty(PropertyGroup):
     cameraTransitionBackCustom: StringProperty(default="0x00")
     isRoomTransition: BoolProperty(name="Is Room Transition", default=True)
 
+    # internal usage only
+    isFromRoomSet: BoolProperty(name="isFromRoomSet", default=False, get=lambda self: self.updateFromRoom())
+
     actor: PointerProperty(type=OOTActorProperty)
+
+    def updateFromRoom(self):
+        if self.fromRoom is None:
+            for obj in bpy.data.objects:
+                if (
+                    obj.type == "EMPTY"
+                    and obj.ootEmptyType == "Room"
+                    and bpy.context.view_layer.objects.active in obj.children_recursive
+                ):
+                    self.fromRoom = obj
+                    return True
+        return False
 
     def isRoomEmptyObject(self, obj: Object):
         return obj.type == "EMPTY" and obj.ootEmptyType == "Room"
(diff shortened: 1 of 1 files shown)


Yanis42:
(marked a ready since I enjoy this solution)

fast64_internal/oot/actor/properties.py
```diff
@@ -248,6 +251,17 @@ def draw_props(self, layout: UILayout, obj: Object, altSceneProp: OOTAlternateSc
 )
 
 
+@persistent
+def actorHandler(_: Scene):
+    for parentObj in bpy.data.objects:
```
kurethedead, line 256: You should check if the game editor mode is "OOT" before executing the rest of this code.

fast64_internal/oot/oot_utility.py
```diff
@@ -939,3 +939,41 @@ def getNewPath(type: str, isClosedShape: bool):
...
+    Parameters:
+    - ``emptyType``: the OoT empty type name (Water Box, Transition Actor, Entrance, ...)
+    - ``propName``: the properties name to get or set tied room objects
+    - ``ptrName``: the pointer property name
+    """
+
+    if scene.gameEditorMode != "OOT":
+        return
+
+    def updateTiedRoom(parentObj: Object, isScene: bool):
```
Dragorn421, line 956: please don't use the same function name twice

Dragorn421 (commented):
can you explain what this achieves? (screenshots of relevant props in blender would be appreciated)

fast64_internal/oot/oot_utility.py
Yanis42: oops lol didn't notice that

fast64_internal/oot/oot_utility.py
```diff
@@ -939,3 +939,41 @@ def getNewPath(type: str, isClosedShape: bool):
...
+                                break
+                        if not foundRoom:
+                            setattr(prop, ptrName, None)
+
+    for parentObj in bpy.data.objects:
+        if parentObj.type == "EMPTY":
+            if parentObj.ootEmptyType == "Room":
+                updateTiedRoom(parentObj, False)
+            elif parentObj.ootEmptyType == "Scene":
+                updateTiedRoom(parentObj, True)
```
Dragorn421, line 979: since you use children_recursive, empties will be processed twice, once per scene and once per room (they're recursively children of both)

depending on who comes first between scene and rooms, the final result will be different, which doesn't sound intended

Yanis42:
> can you explain what this achieves? (screenshots of relevant props in blender would be appreciated)

this updates automatically the "tiedRoom" prop (or "fromRoom" and "toRoom" for doors), it's to avoid people updating this manually, for instance for waterboxes it would set the room to none if it becomes parented to the scene empty and vice versa, this idea was something cdi asked: "Would it be possible to still have the shorthand method of having the "from" room be auto assigned if it happens to be parented to a room object?" (https://discord.com/channels/874816709855440926/874832163483287612/1194365695841013923), which lead to this PR (and the similar ones)
image

Lilaa3:
After fixing merge conflicts this works correctly, but I think this method is unnecessary here. I personally would handle the first room being None by using its room parent (instead of just erroring out, tho that would stay for toRoom). This solution is more simplistic but still intuitive (would prob be even more intuitive if there was a warning in the ui for toRoom being None)

Yanis002:
I fixed the conflicts but I remember not being happy with the current implementation (also iirc people told me their concerns about this but idr well), I think it's a good idea but probably needs more input from people

PR #289, opened 2024-01-10
Minor unused stuff cleanup (mostly OoT)

Yanis42:
I used the Python module [deadcode](https://pypi.org/project/deadcode/) to get a list of the unused things that could be removed, there's unused stuff that could be useful in the future so I kept it (like several F3D classes for instance)

### DIFF
--- a/fast64_internal/oot/oot_level_parser.py
+++ b/fast64_internal/oot/oot_level_parser.py
@@ -947,7 +947,7 @@ def parsePathList(
     pathData = getDataMatch(sceneData, pathListName, "Path", "path list")
     pathList = [value.replace("{", "").strip() for value in pathData.split("},") if value.strip() != ""]
     for pathEntry in pathList:
-        numPoints, pathName = [value.strip() for value in pathEntry.split(",")]
+        _, pathName = [value.strip() for value in pathEntry.split(",")]
         parsePath(sceneObj, sceneData, pathName, headerIndex, sharedSceneData)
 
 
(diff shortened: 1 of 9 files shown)


fast64_internal/oot/oot_level_parser.py
```diff
@@ -947,7 +947,7 @@ def parsePathList(
     pathData = getDataMatch(sceneData, pathListName, "Path", "path list")
     pathList = [value.replace("{", "").strip() for value in pathData.split("},") if value.strip() != ""]
     for pathEntry in pathList:
-        numPoints, pathName = [value.strip() for value in pathEntry.split(",")]
+        _, pathName = [value.strip() for value in pathEntry.split(",")]
```
Dragorn421, line 950: I think it's better to keep even unused things named, it serves as documentation / is more readable
```suggestion
        numPoints, pathName = [value.strip() for value in pathEntry.split(",")]
```

fast64_internal/oot/tools/quick_import.py
```diff
@@ -40,7 +40,7 @@ def quick_import_exec(context: bpy.types.Context, sym_name: str):
 
     all_found_defs: dict[Path, list[tuple[str, str]]] = dict()
 
-    for dirpath, dirnames, filenames in os.walk(assets_objects_dir_p):
+    for dirpath, _, filenames in os.walk(assets_objects_dir_p):
```
Dragorn421, line 43: x2
and this is also a super common construct
```suggestion
    for dirpath, dirnames, filenames in os.walk(assets_objects_dir_p):
```

Dragorn421 (approved):

PR #292, opened 2024-01-14
Fix incorrect version checking code 

kurethedead:


Lilaa3 (approved):
This fixes the logic of the function and with that fixes some blender versions having broken vertex colors. lgtm

Issue #293, opened 2024-01-15
[OoT] Unfriendly error when exporting a room without geometry


Dragorn421:
How to reproduce:

- new blender file
- click "Add Scene" tool
- delete the plane mesh object (or unparent it from the room empty)
- export the scene

-> undescriptive error happens

instead fast64 should either print a legible error, or allow exporting rooms without any geometry attached

PR #294, opened 2024-01-15
[OoT] Fix for issue #293

Yanis42:
I tried to find a solution that didn't involve an extra loop but couldn't find something better 🤷


Dragorn421:
I don't think this fixes the corner case where you do have a mesh object but it has no geometry

Also apparently fast64 already tries to check this elsewhere, see this screenshot (using fast64 version not-this-PR) where I try to export a room with a mesh that has no face but does have vertices:
image

Yanis42:
another way do to this would be checking if the ``BoundingBox`` members are still ``None``

Dragorn421:
Needs a better fix, this isn't satisfactory

PR #295, opened 2024-01-15
[OoT] Waterbox update

Yanis42:
- Added a new property to tie a waterbox to a room (sets and updates automatically, you can choose the room if it's parented to one)
- Fixed the issue where it's exporting the same waterbox twice

This shouldn't change anything for users

### DIFF
--- a/fast64_internal/oot/collision/properties.py
+++ b/fast64_internal/oot/collision/properties.py
@@ -1,10 +1,15 @@
+import bpy
 import math
+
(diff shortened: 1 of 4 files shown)


Dragorn421:
> you can choose the room if it's parented to one

so you can parent a waterbox to a room but still put the waterbox in another room using properties? :thinking: that sounds confusing

Yanis42:
I can make it read-only so it only shows which room it is

Yanis42:
closing this for now, it's not really required anymore anyway (and I don't like the automatic update of the property tbh)

PR #296, opened 2024-01-16
Removed tex1 related code from BSDF to F3D converter

kurethedead:


Lilaa3 (approved):
Why Subsurface Color in the first place? I can't even find a color input for the subsurface color in the principled BSDF shader so I assume it was removed in 4.0 along with the other shader changes? Either way using subsurface color for tex1 makes little sense, lgtm.

Lilaa3:
I actually checked discord, it was for avoiding using f3d nodes back in the ancient days, I guess that makes sense

PR #297, opened 2024-01-16
[OoT] Implemented quick cutscene import and single cutscene import

Yanis42:
This also fix a minor issue related to player cues. Since it would be literally the same class I decided to make the "import all cutscenes" a 2-in-1 operator, if the cutscene name for single import is blank then it will import everything (and warn the user with a label)


fast64_internal/oot/tools/quick_import.py
```diff
@@ -66,24 +79,22 @@ def quick_import_exec(context: bpy.types.Context, sym_name: str):
...
     sym_def_type, sym_def_array_decl = sym_defs[0]
     is_array = sym_def_array_decl != ""
-    object_name = sym_file_p.relative_to(assets_objects_dir_p).parts[0]
+    folder_name = sym_file_p.relative_to(found_dir_p).parts[0]
 
     if sym_def_type == "Gfx" and is_array:
         settings: OOTDLImportSettings = context.scene.fast64.oot.DLImportSettings
         settings.name = sym_name
-        settings.folder = object_name
+        settings.folder = folder_name
```
Dragorn421, line 97: this expects an object name afaik, it wouldn't work for importing a dlist from a scene

I think this operator should be extended to more than objects if it can, but not sure how

Yanis42: I renamed that because it would be weird to use ``object_name`` for cutscenes since they're in scenes, that's why I chose ``folder_name`` since it sets ``settings.folder``, but yeah tbh idk how to name that properly, that was the best idea I could find at the moment

fast64_internal/oot/tools/quick_import.py
```diff
@@ -97,19 +108,22 @@ def quick_import_exec(context: bpy.types.Context, sym_name: str):
...
         settings: OOTAnimImportSettingsProperty = context.scene.fast64.oot.animImportSettings
         settings.isCustom = False
         settings.isLink = False
         settings.animName = sym_name
-        settings.folderName = object_name
+        settings.folderName = folder_name
         bpy.ops.object.oot_import_anim()
+    elif sym_def_type == "CutsceneData" and is_array:
+        path = assets_scenes_dir_p / folder_name / sym_file_p
+        bpy.context.scene.ootCSNumber = importCutsceneData(f"{path}", None, sym_name)
```
Dragorn421, line 123: did you test this? `sym_file_p` should be the path to file itself, no need to prepend anything.
If you were lucky and `sym_file_p` is absolute I guess it is equivalent to `path = sym_file_p` and would still work

fast64_internal/oot/tools/quick_import.py
Dragorn421: so, so far this can quick-import
`Gfx[]`
`(Flex)SkeletonHeader`
`AnimationHeader`
this PR adds:
`CutsceneData`

afaik the first three can only be imported by fast64 from object files (at least using the interface it gives)
so probably the code should behave different based on where it found the symbol.
e.g. if it finds a dlist in a room file, it can error with `Can't import Gfx[] from maps`, but if it's in an object then it would be ok.

fast64_internal/oot/tools/quick_import.py
Yanis42: yeah I tried it with that tiny cutscene in hyrule field that plays after the cutscene where you learn song of time and it worked

Dragorn421:
Trying to import `gHyruleFieldGetOoTCs` doesn't work, with or without having the scene imported first (without cutscenes), using QuickImport or the dedicated operator
:(
```
Error: Python: Traceback (most recent call last):
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/fast64_internal/oot/tools/operators.py", line 288, in execute
    quick_import_exec(
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/fast64_internal/oot/tools/quick_import.py", line 131, in quick_import_exec
    bpy.context.scene.ootCSNumber = importCutsceneData(f"{sym_file_p}", None, sym_name)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/fast64_internal/oot/cutscene/importer/functions.py", line 11, in importCutsceneData
    return csMotionImport.setCutsceneData(bpy.context.scene.ootCSNumber)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/fast64_internal/oot/cutscene/importer/classes.py", line 482, in setCutsceneData
    cutsceneList = self.getCutsceneList()
                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/fast64_internal/oot/cutscene/importer/classes.py", line 196, in getCutsceneList
    parsedCutscenes = self.getParsedCutscenes()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/fast64_internal/oot/cutscene/importer/classes.py", line 123, in getParsedCutscenes
    csData[-1] += line
    ~~~~~~^^^^
IndexError: list index out of range
```

Yanis42:
should work now

Dragorn421 (approved):

PR #299, opened 2024-01-18
Prefer RGBA over CI toggle for automatic texture format

Lilaa3:
Suggested by a ton of people.
Kinda of inconvenient for people who care about performance but with repo settings files (see settings rework pr) it will be no issue for anyone who cares.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -280,6 +280,7 @@ class Fast64Settings_Properties(bpy.types.PropertyGroup):
         description="When enabled, fast64 will default colored textures's format to RGBA even if they fit CI requirements, with the exception of textures that would not fit into TMEM otherwise",
     )
 
+
 class Fast64_Properties(bpy.types.PropertyGroup):
     """
     Properties in scene.fast64.
--- a/__init__.py
+++ b/__init__.py
@@ -212,7 +212,9 @@ class Fast64_GlobalSettingsPanel(bpy.types.Panel):
         prop_split(col, context.scene, "gameEditorMode", "Game")
         col.prop(context.scene, "exportHiddenGeometry")
         col.prop(context.scene, "fullTraceback")
-        col.prop(context.scene.fast64.settings, "prefer_ci_over_rgba")
+        col.separator()
+        col.prop(context.scene.fast64.settings, "auto_pick_texture_format")
+        col.prop(context.scene.fast64.settings, "prefer_rgba_over_ci")
         prop_split(col, context.scene.fast64.settings, "anim_range_choice", "Anim Range")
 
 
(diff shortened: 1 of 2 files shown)


__init__.py
```diff
@@ -267,6 +268,7 @@ class Fast64Settings_Properties(bpy.types.PropertyGroup):
         ],
         default="intersect_action_and_scene",
     )
+    prefer_ci_over_rgba: bpy.props.BoolProperty(name="Prefer Color Indexed Over RGBA")
```
jesusyoshi54, line 271: Please add in a description to this prop explaining how it works and when it applies.

jesusyoshi54 (commented):
```python
def getOptimalFormat(tex, curFormat, isMultitexture):
    texFormat = "RGBA16"
    if isMultitexture:
        return curFormat
```
[link](https://github.com/Lilaa3/fast64/blob/37aacff07adb1be9bdfb6fd71855bf595d94e267/fast64_internal/f3d/f3d_material.py#L3369)

If you are adding a multi-texture, and one texture is already set, the optimal texture format of the second one is `curFormat` unless the other texture is in opposition to the current format, in which case the optimal format is CI or whatever this function returns.

A CI multi-texture must use another CI texture and likewise an RGBA/Intensity texture must NOT use a CI texture.

I would also say that it is best if when choosing a new texture in a multi-texture, that it optimally chose the correct format between available options. If I swap between a color or intensity option, I would prefer if the texture format automatically updated. The only restriction is that I can't use a CI.

Lilaa3:
> If you are adding a multi-texture, and one texture is already set, the optimal texture format of the second one is `curFormat` 
Wait why?? curFormat is just the format of the not updated texture format

__init__.py
Lilaa3: Will do

fast64_internal/f3d/f3d_material.py
```diff
@@ -3392,12 +3392,11 @@ def getOptimalFormat(tex, curFormat, isMultitexture):
                 hasAlpha4bit = True
             if color[3] < 0.5:
                 hasAlpha1bit = True
-            pixelColor = getRGBA16Tuple(color)
-            if pixelColor not in pixelValues:
-                pixelValues.append(pixelColor)
+            pixelValues.add(getRGBA16Tuple(color))
 
+    tex_size = tex.size[0] * tex.size[1]
```
Dragorn421, line 3397:
```suggestion
    n_pixels = tex.size[0] * tex.size[1]
```

__init__.py
Dragorn421:
```suggestion
    prefer_ci_over_rgba: bpy.props.BoolProperty(name="Prefer CI Over RGBA")
```
also if the better choice is for this to be on, it should be on by default. it's easy to change and save as default blend

Dragorn421: bump about the default

the description should be clearer that this is about the auto/default format when choosing a texture, not something that overrides what's in the UI

jesusyoshi54:
> Wait why?? curFormat is just the format of the not updated texture format

Currently it returns curFormat universally while making a multi-texture. This is fine but not under the circumstances I laid out before.
On top of that I think you could also just run the format change as needed but just exclude CI or non CI textures as needed

Dragorn421:
Is this PR waiting on anything? (for what specifically?)

If just waiting for lila to answer np I don't mean to rush you. Just want to make sure we're not in a confusingly-waiting-each-other deadlock scenario

also bump again for the `prefer_ci_over_rgba` default

jesusyoshi54:
Shouldn't the `prefer_ci_over_rgba` prop be UI dependent on `auto_pick_texture_format` prop? Maybe having it dynamic kind of sucks idk. I feel like there should be a UI connection of the props though

Lilaa3:
Is there anything else missing @jesusyoshi54 ?

jesusyoshi54:
> Is there anything else missing @jesusyoshi54 ?

I'll review tonight. First glance looks good

jesusyoshi54 (approved):
Ok this is very nice I like a lot good job lila you are the best

__init__.py
Lilaa3: sorry to bother you @Dragorn421 but is it good in its current state? if so can i go ahead and merge?

PR #304, opened 2024-01-31
Solidify checking materials assigned to faces

Dragorn421:
`if material_index >= len(material_slots):` (not sure how to reproduce a blend triggering this, I got one in DMs from someone)
image

`if material is None:` (slot without a material)
image

`if not material.is_f3d:` (material isn't f3d)
image


jesusyoshi54 (approved):
works on my machine

Issue #305, opened 2024-02-02
gc-eu-mq-dbg seems to be an additonal directory that is breaking the decomp build


kevin192291:
It seems the decomp file now has a directory within the build folder called `gc-eu-mq-dbg`
This seems to be breaking fast64's exports. I think this needs to be updated, I am very new to this and
am not sure if I am correct here, but after I export a scene I need to copy the contents of gc-eu-mq-dbg into
build (along side gc-eu-mq-dbg), and everything builds again.

If I am on the correct track, Let me know and I would love to try to fix this :)

Yanis42:
yeah this is a known issue, I'm not making it my top priority as you can use a workaround with the "custom export path" checkbox, it won't edit spec and scene_table for you but at least it will work for sure which the important part here imo

if you wanna fix it you need to change stuff [here](https://github.com/Fast-64/fast64/blob/main/fast64_internal/oot/scene/exporter/to_c/spec.py) I think (I haven't dig into this but afaik the spec file is the only broken thing right now)

Note that I plan to work on a fix once I get the HackerOoT rework done, which should happen soonish™™™, feel free to work on it, just let me know

and of course feel free to ask questions or anything else on [Fast64's Discord](https://discord.gg/ny7PDcN2x8), it will be easier to communicate on this from Discord

Yanis42:
also the fix is just replacing ``build/`` by ``$(BUILD)/`` I believe (in the file I linked)

Dragorn421:
Another workaround is editing the Makefile this way, so `build/` as written by fast64 is replaced with `build/gc-eu-mq-dbg/` on processing the spec:

```diff
diff --git a/Makefile b/Makefile
index a8869f3fe..6da61a684 100644
--- a/Makefile
+++ b/Makefile
@@ -125,10 +125,10 @@ PYTHON     ?= $(VENV)/bin/python3
 
 # Command to replace path variables in the spec file. We can't use the C
 # preprocessor for this because it won't substitute inside string literals.
-SPEC_REPLACE_VARS := sed -e 's|$$(BUILD_DIR)|$(BUILD_DIR)|g'
+SPEC_REPLACE_VARS := sed -e 's|build/|$(BUILD_DIR)/|g' | sed -e 's|$$(BUILD_DIR)|$(BUILD_DIR)|g'
 
 ifeq ($(COMPILER),gcc)
```

Yanis42:
[...]

PR #306, opened 2024-02-07
[OoT] Refactor ``spec.py``

Yanis42:
this fixes the recent issues with decomp and it should prevent future breaks, there's 2 things I need to do but I wanted to PR this anyway so I'm just gonna make it a draft

this PR is in the same spirit as #272, with this spec refactor you can copy paste everything (except ``editSpecFile``) somewhere else, call ``SpecFile`` and use this as a standalone script

also this is a fix for #305 and the other issue HackerOoT has with ifdefs


Yanis42:
just have newline issues left, otherwise it should be ready (keeping as a draft while this is not fixed)

Yanis42:
ok well, no idea where the newline comes from (also it's a kind of edge case tbh), I'll try to see if there's more concerning issues otherwise I'll mark it ready to review

sauraen (approved):

PR #308, opened 2024-02-18
Filled in typehints and doctrings for fast64_internal/utility.py

ThomasJRyan:
What's the best way to jump into a project? Read the documentation!

No documentation? Second best way then, write the documentation!

Docstrings are heavily written by Copilot, typehints are all traced back by myself. I'm reasonably certain of the accuracy of it all.

### DIFF
--- a/fast64_internal/utility.py
+++ b/fast64_internal/utility.py
@@ -59,77 +59,34 @@ enumCompressionFormat = [
 
 
 def isPowerOf2(n: int | float) -> bool:
-    """Check if a number is a power of 2
-
-    Args:
-        n (int | float): Number to check
-
-    Returns:
-        bool: True if n is a power of 2
-    """
+    """Check if a number is a power of 2"""
     return (n & (n - 1) == 0) and n != 0
 
 
 def log2iRoundDown(n: int | float) -> int:
-    """Calculate the log base 2 of a number, rounding down
-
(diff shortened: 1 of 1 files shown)


sauraen:
Thanks for the submission!

There are already some conflicts, but there is another PR or two going in which touch this file too. Once that happens please merge main to your branch and fix the conflicts, then update this branch. Feel free to tag me to review then.

fast64_internal/utility.py
```diff
@@ -137,33 +183,56 @@ def restoreHiddenState(hiddenState):
         layerCol.hide_viewport = True
 
 
-def readFile(filepath):
+def readFile(filepath: str | Path) -> str:
```
Lilaa3, line 186: How about [os.PathLike](https://docs.python.org/3/library/os.html#os.PathLike)
File descriptors should technically also work with this function but fast64 never uses them so meh

fast64_internal/utility.py
```diff
@@ -137,33 +183,56 @@ def restoreHiddenState(hiddenState):
...
+    """Read a file and return its contents"""
```
Lilaa3, line 187: Specify that its expecting a text file with utf-8 encoding

Lilaa3 (changes requested):
I will review more later

PR #317, opened 2024-03-01
add the test of cube export and import

coco875:


Lilaa3:
Sorry for never getting around to reviewing this, but before commenting on the code what porpuse does this serve?
Yes automated testing is always nice, but this tests so very little, basically just makes sure an exception isn´t happening in two methods (and you're not even calling the operators themselves).

Yanis42:
for oot we probably want to import a scene, export one and make sure the DL/Collision/Skeleton/Animation exporters/importers all work

Lilaa3:
> for oot we probably want to import a scene, export one and make sure the DL/Collision/Skeleton/Animation exporters/importers all work

One of my big issues with this is what would count as works?

coco875:
Other solutions will be to use a real rom but can make some legal issues

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
+name: Linux test
+
+on:
+    push:
+        branches: [ "*" ]
+    pull_request:
+        branches: [ "master" ]
```
Lilaa3, line 7: I noticed this only applies to master, do we actually want this?

coco875: Does fast64 have other branch that master that will not make a pr to master ?

Lilaa3: Not currently no, but that's another reason to just apply to all branches (because if its a pr to fast64, it doesn't matter because there is only main, and if its a pr to a branch of a fork it will still run)

coco875: Ok then will change that once I have the possibility

PR #324, opened 2024-04-06
[SM64] Move into the new folders for settings rework

Lilaa3:
Suggested by dragon421 to make reviewing easier


jesusyoshi54 (approved):
Code makes sense and it exported a level and Mario.

Dragorn421 (approved):
doesn't seem to break oot
Merge for next version

PR #325, opened 2024-04-13
Blender Version 4.1 Fix

jesusyoshi54:
Title
Normals are always calculated from version 4.1 onwards, making the use of `calc_normals_split()` unnecessary


Dragorn421 (approved):
Nice, thanks

### On the code side:

The release notes
https://developer.blender.org/docs/release_notes/4.1/python_api/
mention the function was removed and the recomputation of normals is now automatic, so the changes here make sense

### On testing:

Reonu found no issue from a quick test
https://discord.com/channels/874816709855440926/874816710639767585/1227970692826271797

Reonu:
This should be a release btw. The readme explicitly recommends "blender 4.0+". The updater should really take people to this version ASAP after the merge.

Issue #330, opened 2024-04-17
Handle no scene.world in a friendlier way


Dragorn421:
Illustrated by #326

Sometimes (when???) the scene.world (which stores scene lighting settings in vanilla blender and additional related props for fast64) is None

This prevents materials from working (they appear pink according to whatever causes that default behavior) and any exporting fails with a cryptic message

The fix is simply to create a world in the scene world properties

Fast64 should handle a missing world better
(I'm not sure on the specifics of how)

PR #334, opened 2024-04-21
[Codebase] Automatically find or add world

Lilaa3:
Possible solution to #330 
Never changes the actual scene.world, as that should be left to the user, the important logic is in create_or_get_world.
This will almost never create a world because of the f3d nodes library having its own world


Lilaa3:
@Dragorn421 could you look at this at some point?

Lilaa3:
Would an exception or warning popup work better?

fast64_internal/utility.py
```diff
@@ -1868,3 +1868,22 @@ def upgrade_old_prop(
...
+    - The first world in bpy.data.worlds if the current file has a world. (Which it almost always does because of the f3d nodes library)
+    - Create a world named "Fast64" and return it if no world exits.
+    This function does not assign any world to the scene.
+    """
+    if scene.world:
+        return scene.world
+    if bpy.data.worlds:
+        world: World = bpy.data.worlds.values()[0]
+        print(f'No world selected in scene, selected the first one found in this file "{world.name}".')
+        return bpy.data.worlds.values()[0]
```
Dragorn421, line 1886:
```suggestion
        return world
```

Lilaa3: True

Dragorn421 (approved):
great!

Dragorn421:
Will fix #330

fast64_internal/utility.py
```diff
@@ -1868,3 +1868,22 @@ def upgrade_old_prop(
...
+    - The world selected in the scene if the scene has a selected world.
+    - The first world in bpy.data.worlds if the current file has a world. (Which it almost always does because of the f3d nodes library)
+    - Create a world named "Fast64" and return it if no world exits.
+    This function does not assign any world to the scene.
+    """
+    if scene.world:
+        return scene.world
+    if bpy.data.worlds:
+        world: World = bpy.data.worlds.values()[0]
+        print(f'No world selected in scene, selected the first one found in this file "{world.name}".')
```
Dragorn421, line 1885: actually I just realized this severely spams the console. Could you add a counter for this print and only print it the first 10 times or something
use a dirty global variable :)

Dragorn421 (approved):
thanks

PR #336, opened 2024-04-21
[SM64] Sort Level Objects Alphabetically

jesusyoshi54:
added sorting to object exports in sm64. Actors are sorted alphabetically. This will generally make it so the same type of object all export in a row since these will have names like: coin, coin.001, coin.002 etc.

### DIFF
--- a/fast64_internal/sm64/sm64_objects.py
+++ b/fast64_internal/sm64/sm64_objects.py
@@ -262,13 +262,14 @@ enumPuppycamFlags = [
 
 
 class SM64_Object:
-    def __init__(self, model, position, rotation, behaviour, bparam, acts):
+    def __init__(self, model, position, rotation, behaviour, bparam, acts, name):
         self.model = model
         self.behaviour = behaviour
         self.bparam = bparam
         self.acts = acts
         self.position = position
         self.rotation = rotation
+        self.name = name  # to sort by when exporting
 
     def to_c(self):
         if self.acts == 0x1F:
@@ -325,6 +326,7 @@ class SM64_Whirpool:
         self.condition = condition
         self.strength = strength
         self.position = position
+        self.name = "whirlpool"  # for sorting
 
     def to_c(self):
         return (
@@ -453,6 +455,7 @@ class SM64_Mario_Start:
         self.area = area
         self.position = position
         self.rotation = rotation
+        self.name = "Mario"  # for sorting
 
     def to_c(self):
         return (
@@ -500,7 +503,8 @@ class SM64_Area:
         data += "\tAREA(" + str(self.index) + ", " + self.geolayout.name + "),\n"
         for warpNode in self.warpNodes:
             data += "\t\t" + warpNode + ",\n"
-        for obj in self.objects:
+        # export objects in name order
+        for obj in sorted(self.objects, key=(lambda obj: obj.name)):
             data += "\t\t" + obj.to_c() + ",\n"
         data += "\t\tTERRAIN(" + self.collision.name + "),\n"
         if includeRooms:
@@ -840,6 +844,7 @@ def process_sm64_objects(obj, area, rootMatrix, transformMatrix, specialsOnly):
                         behaviour,
                         obj.fast64.sm64.game_object.get_behavior_params(),
                         get_act_string(obj),
+                        obj.name,
                     )
                 )
             elif obj.sm64_obj_type == "Macro":


Lilaa3:
Is there any reason whirlpools and mario starts dont have a seperate list like specials, macros, puppy cam volumes, etc?

jesusyoshi54:
I don't believe there is any special reason, I think it just categorically was grouped as part of the `area` while objects felt deserving of their own separate lists. Probably because they won't be scrutinized more.
It was this way in vanilla so kure probably copied that and it has carried over.

Lilaa3:
wait really? I checked the hacker implementation and they don't create any objects, ao I guess that's hacker specific? sorry about that aha, if that is the case then this looks perfect already, I will do a level export later today and then if everything is correct I will get this merged

Lilaa3:
It seems objects are getting sorted correctly by blender which I don´t remember being the case? In fact this pr broke correct order by putting capitalized names first, if the objective is to have the same types of objects be placed in order then why not sort by model and behavior and presets (for macros)
https://github.com/jesusyoshi54/fast64-KCS/pull/3 here are my suggested changes/implementation, I wrote it a bit carelessly but it does get the job done perfectly

jesusyoshi54:
> 

> It seems objects are getting sorted correctly by blender which I don´t remember being the case? In fact this pr broke correct order by putting capitalized names first, if the objective is to have the same types of objects be placed in order then why not sort by model and behavior and presets (for macros)

The goal

jesusyoshi54:
> wait really? I checked the hacker implementation and they don't create any objects, ao I guess that's hacker specific?

What I meant is for grouping. Mario start and whirlpools aren't objects, but they're chunked together in the same lists as objects in the area level script, so that's why I believe it was coded that way. There's no actual reason it has to be.

jesusyoshi54:
> It seems objects are getting sorted correctly by blender which I don´t remember being the case? In fact this pr broke correct order by putting capitalized names first, if the objective is to have the same types of objects be placed in order then why not sort by model and behavior and presets (for macros)

The goal was to allow the user to sort the objects. I don't know if that is always by model or by behavior. I think the original reason this was brought up was due to some kaze thing where he needed stuff in a specific order.
I don't know what correct order I broke with this, could you explain that?

Lilaa3:
at least in blender 4.1 it seems objects are getting sorted in the UI and in the actual data.
to answer your question (again) your code sorts differently, by capitalization and only then by alphabetic order which is confusing in my opinion.
In any case do you see any use for this where a sort by model, behavior and then just keeping the order that blender has isn't better? If someone wanted to use this for say batching, that would work out of the box. I'd say it makes a lot of sense to sort like that.
This PR might be needed for older blender versions tho, which I seem to remember, if you think model behavior sorting takes away control from the user I can accept that and the only change I'd like you to make then is updating the lambda to sort alphabetically like blender.

jesusyoshi54:
I'd prefer to sort by what the UI has, that's the most intuitive. I'll have to add an ignore case to the sort as it is then, it should match what the blender UI does.
While I don't think it would be common to not sort by model, it can happen, and I would think it is worth allowing customized sorting for any specific weirdness users may want to do; something like a water diamond/water diamond init obj may be such a case.

jesusyoshi54:
Ok from my testing on versions 3.2 and 3.6 of blender, the current method of sorting matches what the UI does. Only blender 4.0+ has an issue where it starts doing some wack stuff. I really have no clue what it really is doing since it doesn't match any sort of ordering I could tell from looking at the console.
image
You can see the order of the 1ups (denoted by name) does not match with what prints out in the console. I believe this is just the blender UI for 4.0 being wack.
image
Here I got an example of a `M *` string coming after a `m *` string, which really doesn't make sense.

Lilaa3:
I really tried understanding how this works but I'm having a really hard time understanding the blender source code

Lilaa3:
Ok from what I did understand the order in the UI isn't achieved with any form of conventional name sorting specifically, and we should not rely on the order of bpy.data.objects, your original implementation is consistent and that's good
I still think the behavior I was seeing was weird, if I two objects like "Blargg" and "amp", blaarg would come first from just the capitalization, is this that common and I'm just not used to seeing it? and does pre 4.0 blender also do it?

jesusyoshi54:
> your original implementation is consistent and that's good I still think the behavior I was seeing was weird, if I two objects like "Blargg" and "amp", blaarg would come first from just the capitalization, is this that common and I'm just not used to seeing it? and does pre 4.0 blender also do it?

My experience from 3.X is the order always matched string ordering. Capital letters First, then lowers, alphabetical order in those sets. It always did this as far as I'm aware, but maybe in some versions past 3.6 it didn't, but I don't use those. This is the same way python sorts strings.

Lilaa3 (approved):

Issue #320, opened 2024-03-21
Rotations involving roll on level objects are not correctly exported to levels (SM64)


yoyotam3:
To replicate:
Create new object in a level
Give it any rotation that involves roll combined with axis rotation? Unsure of exact criterion for whether a rotation isn't exported correctly
Export to level
Enter level ingame 
Object is not at the same rotation that was shown in Blender

yoyotam3:
Seems to be a euler order issue, as I was able to get correct rotations by first rotating the object to the desired position in XYZ euler, then changing to ZXY euler

Lilaa3:
Could you provide a file please? From a quick glance at the object export code I can´t see the issue

yoyotam3:
Sorry, I meant to provide more detail after I figured out some stuff, but forgot. Making a blend file to show the issue rn, but while I was making my hack I figured out that the correct way to convert rotations is to first convert the blender rotations to YXZ (I think you're using ZXY which might be the issue), which I did by changing the rotation type to quaternions then to YXZ, then swapping the Y and Z values, then making the Z negative, then putting at into the levelscript. So if the blender YXZ Euler rotation values are bX, bY,bZ, it's (bX,bZ,-bY).

yoyotam3:
[example_incorrect_rotation.zip](https://github.com/Fast-64/fast64/files/15056588/example_incorrect_rotation.zip)
this blend file should be a good demonstration of the issue.

Lilaa3:
Ok I just woke up, I will step through the export with debug if needed, it's way easier to find issues that way, fast64 reads the actual object matrix and gets the rotation from that so all that should not be an issue but again I will check it out

Lilaa3:
I spent way too long trying to understand why it was still wrong but then i decided to actually check if the orientation we both expected was the actual orientation, it wasn't aha
Add a simple mesh with a clear orientation, parent it to the test object
![image](https://github.com/Fast-64/fast64/assets/87947656/1116d2b7-8ed7-4d7c-8d6a-e4d1ad649dc1)
Anyways I made a fix! It's mentioned right above this comment
![image](https://github.com/Fast-64/fast64/assets/87947656/4ec4fc92-323e-46fb-996a-0d4dfa05b859)

yoyotam3:
Oh I didn't intend for the axes colors to represent actual specific axes if that's what you mean. Sorry for the confusion

yoyotam3:
Should I close the issue? Not super familiar with github.

[...]

PR #339, opened 2024-04-22
[SM64] Fix object rotations

Lilaa3:
Export code was not accounting for the zxy order of object rotations (animations rotate xyz so they don't have this issue), this pr fixes that (see #320)


jesusyoshi54 (approved):
yes

PR #345, opened 2024-05-07
Fixed bug with custom lights

sauraen:
Reported by TrevorBelmondo23, custom lights are not exported correctly, caused by corrupted indentation.


jesusyoshi54 (approved):
Good

PR #350, opened 2024-05-26
[F3D] Fog Export Fix

jesusyoshi54:
Fog now exports only when the `set_fog` boolean is checked and fog is used in the material, which is determined if `g_fog` is enabled, or if fog is an input to any of the blend variables. Previous behavior only checked if `set_fog` was on, which would mistakenly allow fog to be exported when the `set_fog` checkbox was invisible (and left checked from prev config) when material was changed to no longer have any fog inputs.

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -230,24 +230,6 @@ def rendermode_preset_to_advanced(material: bpy.types.Material):
     settings.blend_b2 = f3d.blendMixDict[(r2 >> 16) & 3]
 
 
-def does_blender_use_color(settings: "RDPSettings", color: str, default_for_no_rendermode: bool = False) -> bool:
-    if not settings.set_rendermode:
-        return default_for_no_rendermode
-    is_two_cycle = settings.g_mdsft_cycletype == "G_CYC_2CYCLE"
-    return (
-        settings.blend_p1 == color
-        or settings.blend_m1 == color
-        or (is_two_cycle and (settings.blend_p2 == color or settings.blend_m2 == color))
-    )
-
-
-def does_blender_use_alpha(settings: "RDPSettings", alpha: str, default_for_no_rendermode: bool = False) -> bool:
-    if not settings.set_rendermode:
-        return default_for_no_rendermode
-    is_two_cycle = settings.g_mdsft_cycletype == "G_CYC_2CYCLE"
-    return settings.blend_a1 == alpha or (is_two_cycle and settings.blend_a2 == alpha)
-
-
 def does_blender_use_mix(settings: "RDPSettings", mix: str, default_for_no_rendermode: bool = False) -> bool:
     if not settings.set_rendermode:
         return default_for_no_rendermode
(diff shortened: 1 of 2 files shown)


Lilaa3:
I don´t know if I like this, one could use the alpha in the combiner with a texedge render mode for example, and this would break that ability. Also I don't see the behavior you are describing of fog being hidden in the UI?

jesusyoshi54:
> I don´t know if I like this, one could use the alpha in the combiner with a texedge render mode for example, and this would break that ability.

I'm not sure what will be broken here.

> Also I don't see the behavior you are describing of fog being hidden in the UI?
Then I must be dumb and mistaken. I could have sworn this was an issue, but it might have been fixed sometimes with the render mode changes sauraen made.

Lilaa3:
What will be broken is that, since the material is technically not using fog in render mode, fast64 will just not export the actual fog

jesusyoshi54:
If fog isn't used in the render mode then why would it matter if it is unset.
If you're using shade alpha in the combiner you need g_fog enabled, in which case it will export.

Dragorn421:
Hm I'd like to point out the combiner may use shade alpha without intending for fog, with said alpha being provided by the user in the Vtx data (alongside either verter colors or normals)

Lilaa3:
Ah sorry I misread this, I thought you were checking for both g_fog and a fog render mode, my bad, and also set fog does get hidden still, sorry for so much confusion on my part

fast64_internal/f3d/f3d_material.py
```diff
@@ -3231,6 +3209,29 @@ class RDPSettings(PropertyGroup):
...
+        return (getattr(self, f"blend_{val}") for val in ("p1", "p2", "m1", "m2"))
+
+    @property
+    def blend_alpha_inputs(self):
+        return (getattr(self, f"blend_{val}") for val in ("a1", "a2", "b1", "b2"))
+
+    @property
+    def blend_inputs(self):
+        return self.blend_color_inputs
+        return self.blend_alpha_inputs
```
Lilaa3, line 3227: Did you mean to do something like
```suggestion
        return self.blend_color_inputs + self.blend_alpha_inputs
```

fast64_internal/f3d/f3d_material.py
```diff
@@ -3231,6 +3209,29 @@ class RDPSettings(PropertyGroup):
         update=update_node_values_with_preset,
     )
 
+    @property
+    def using_fog(self):
+        return self.g_fog or any(["FOG" in input for input in self.blend_inputs])
```
Lilaa3, line 3214: I personally would just check for the two blend modes directly for the sake of clarity, color inputs will never get new inputs added there is just those two for fog so its good to make that explicit I think
```suggestion
        return self.g_fog or self.does_blender_use_input("G_BL_CLR_FOG") or self.does_blender_use_input("G_BL_A_FOG")
```

fast64_internal/f3d/f3d_material.py
```diff
@@ -3231,6 +3209,29 @@ class RDPSettings(PropertyGroup):
...
+    @property
+    def blend_inputs(self):
+        return self.blend_color_inputs
+        return self.blend_alpha_inputs
+
+    def does_blender_use_input(self, setting: str) -> bool:
+        for input in self.blend_inputs:
+            if setting == input:
+                return True
+        return False
```
Lilaa3, line 3233: this logic could be one line
```suggestion
        return any(input == setting for input in self.blend_inputs)
```

Lilaa3 (changes requested):
I really like the way you implemented these changes, Just a few small details

fast64_internal/f3d/f3d_material.py
jesusyoshi54: These are generators so I did mean that

Lilaa3 (approved):
The code works perfectly, I like how it is written, merge soon.

PR #352, opened 2024-05-27
[SM64] Fixed Negative Scales Gfx

### LINKED ISSUE #346: Mirroring an object in object mode will cause the visable geometry to desync with the collision
mirroring an object (aka giving it a negative scale value) will cause a desync between the exported collision and the rendered geometry

1. make a basic level with asymmetric geometry
2. (in blender) Object > Mirror > Y Global
3. export level into HackerSM64
4. compile
5. run the rom and run around the custom level.  the collision will not match.

jesusyoshi54:
Graphics at negative scales would have an extra rotation applied to the geo layout node, which would result in the collision and graphics not aligning. With this, graphics and collision should match in game, and also graphics will match the in blender visualization with the exception of normal direction (which is sort of a blender issue).
Once fixed, close #346

### DIFF
--- a/fast64_internal/utility.py
+++ b/fast64_internal/utility.py
@@ -772,7 +772,9 @@ def yield_children(obj: bpy.types.Object):
 def store_original_mtx():
     active_obj = bpy.context.view_layer.objects.active
     for obj in yield_children(active_obj):
-        obj["original_mtx"] = obj.matrix_local
+        # negative scales produce a rotation, we need to remove that since
+        # scales will be applied to the transform for each object
+        obj["original_mtx"] = Matrix.LocRotScale(obj.location, obj.rotation_euler, None)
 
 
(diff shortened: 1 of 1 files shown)


Lilaa3 (approved):
The current methods seem very messy to me and normals being wrong still is really bad, but as a temporary fix it is fully valid, should not affect oot at all
[test bps.zip](https://github.com/user-attachments/files/15517120/test.bps.zip)
And my branch https://github.com/Lilaa3/HackerSM64/tree/fast-352-test

jesusyoshi54:
I can fix the normals to match visuals in blender but it wouldn't be "correct" so I decided not to, but I'm not sure on the philosophy of other people.
I also agree the way exports are done is messy but this was the least intrusive way to fix it without writing a new exporter.

Issue #331, opened 2024-04-17
Broken mat revert displaylists causing vanish cap state to bleed into other displaylists


AgentXLP:
When exporting Mario's model, the vanish cap state visibly bleeds into other transparent displaylists. I theorize the whole introduction of F3DEX3 and mat bleeding may have created this bug. This has never occurred until recently and has caused issues for lots of character pack creators on sm64coopdx.
![image](https://github.com/Fast-64/fast64/assets/44549182/ebb41f0e-8c7a-4566-b7d7-88687516b727)
This screenshot was taken on sm64coopdx, a Super Mario 64 PC Port based project.

```c
Gfx mario_Right_Foot_DL_mesh_layer_1_mat_override_metal_0[] = {
	gsSPDisplayList(mat_mario_metal),
	gsSPDisplayList(mario_Right_Foot_DL_mesh_layer_1_tri_0),
	gsSPDisplayList(mat_revert_mario_metal),
	gsDPPipeSync(),
	gsSPSetGeometryMode(G_LIGHTING),
	gsSPClearGeometryMode(G_TEXTURE_GEN),
	gsDPSetCombineLERP(0, 0, 0, SHADE, 0, 0, 0, ENVIRONMENT, 0, 0, 0, SHADE, 0, 0, 0, ENVIRONMENT),
	gsSPTexture(65535, 65535, 0, 0, 0),
	gsDPSetEnvColor(255, 255, 255, 255),
	gsDPSetAlphaCompare(G_AC_NONE),
	gsDPPipeSync(),
	gsSPSetGeometryMode(G_LIGHTING),
	gsSPClearGeometryMode(G_TEXTURE_GEN),
	gsDPSetCombineLERP(0, 0, 0, SHADE, 0, 0, 0, ENVIRONMENT, 0, 0, 0, SHADE, 0, 0, 0, ENVIRONMENT),
	gsSPTexture(65535, 65535, 0, 0, 0),
	gsDPSetEnvColor(255, 255, 255, 255),
	gsDPSetAlphaCompare(G_AC_NONE),
	gsSPEndDisplayList(),
};
```
I'm not exactly an expert at reading and understanding displaylists, but this one for example doesn't look right. It's running the same list of commands twice after the displaylist calls. Using an older version of Fast64 is currently the only real fix for us.

Lilaa3:
[...]

PR #353, opened 2024-05-27
[SM64] Properly Apply DL Reverts to Last Drawn with Switch Nodes

jesusyoshi54:
This issue is caused by reverts not being applied properly to all switch children. Basically only the last switch child got the revert, when it should be the last drawn one, and if it could be multiple different options, all of those should get a revert.
The first image is the bug in question, and the second is the one with fixes applied.
image

image

Mentioned in #331


Lilaa3 (approved):
Behavior seems to be correct, code looks good, mario exported properly with the needed revert commands PLUS this removed an unnecessary revert in the switch root, awesome work!

PR #356, opened 2024-05-30
[SM64] add LOAD_VANILLA_OBJECTS to list of segment loads cmds

aglab2:
Add `LOAD_VANILLA_OBJECTS` cmd to support PR https://github.com/HackerN64/HackerSM64/pull/796


Lilaa3 (approved):
This will not break any non hackersm64 repo so its an easy approval

PR #357, opened 2024-06-07
Animal Crossing F3DZEX Extensions

Lilaa3:
Used in the hle renderer in Animal Crossing.

### DIFF
--- a/fast64_internal/f3d/f3d_bleed.py
+++ b/fast64_internal/f3d/f3d_bleed.py
@@ -17,6 +17,10 @@ from .f3d_gbi import (
     SPLine3D,
     SPLineW3D,
     SP2Triangles,
+    SPNTrianglesInit_5b,
+    SP5bitTriangles,
+    SPNTrianglesInit_7b,
+    SP7bitTriangles,
     SPCullDisplayList,
     SPSegment,
     SPBranchLessZraw,
@@ -48,6 +52,7 @@ from .f3d_gbi import (
     GbiMacro,
 )
 
+TRI_CMDS = [SP2Triangles, SP1Triangle, SPLine3D, SPLineW3D, SPNTrianglesInit_5b, SP5bitTriangles, SPNTrianglesInit_7b, SP7bitTriangles]
 
 class BleedGraphics:
     # bleed_state "enums"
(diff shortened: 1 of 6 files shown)


Lilaa3:
Existing functionality seems done, all thats missing is looking into pallete references maybe?

__init__.py
```diff
@@ -175,6 +182,10 @@ class Fast64Settings_Properties(bpy.types.PropertyGroup):
         name="Prefer RGBA Over CI",
         description="When enabled, fast64 will default colored textures's format to RGBA even if they fit CI requirements, with the exception of textures that would not fit into TMEM otherwise",
     )
+    point_lit: bpy.props.BoolProperty(
+        name="Supports Point Ligthing",
+        description="When enabled, fast64 will export expecting a gbi that supports point lighting, such as Majora's Mask's ucode",
+    )
```
sauraen, line 188: Just taking a quick look, this is not the right place for this setting, it should be with the microcode version.

Lilaa3: Good point sorry

Lilaa3:
Decal geo modes have a lot more info displayed now to make their behavior clear

Lilaa3:
Branch is now up to date with main

Dragorn421 (commented):
for reference there is a list of f3dex2 ucodes and some games they're used in, in the readme at https://github.com/Mr-Wiseguy/f3dex2 (animal crossing = animal forest)

it may be worthwhile to name the ucodes per their version name, like 2.08J for AC, rather than by the game they're used in.
For example different versions of OoT use different ucodes: n64 releases use 2.06H, gamecube releases use 2.08H which notably adds point lighting. naming one "n64 oot" and the other "gc oot" (or "mm") sounds weird
And one may want to use a different ucode in a hack and it would be weird to eg use the "F3DZEX AC" ucode in OoT

Lilaa3:
> for reference there is a list of f3dex2 ucodes and some games they're used in, in the readme at https://github.com/Mr-Wiseguy/f3dex2 (animal crossing = animal forest)
> 
> it may be worthwhile to name the ucodes per their version name, like 2.08J for AC, rather than by the game they're used in. For example different versions of OoT use different ucodes: n64 releases use 2.06H, gamecube releases use 2.08H which notably adds point lighting. naming one "n64 oot" and the other "gc oot" (or "mm") sounds weird And one may want to use a different ucode in a hack and it would be weird to eg use the "F3DZEX AC" ucode in OoT

Techinically the ucode "is" the same as mm's internaly but that.. does not describe the ucode because it's extended with a lot of commands to assist the hle renderer, I'm gonna confirm with cuyler that there is no internal name nonetheless

Lilaa3:
Culyer confirmed to me that he hasn´t seen any special internal ucode name that could be used, that said if you'd like a better name I'm open to suggestions @Dragorn421 , maybe F3DZEX2 (AC) would at least be closer

Lilaa3:
Drafting, I wanna look at texture export code with numpy and new quantization methods, libragon compat, etc (which this would be using if I get the job done, currently it has an early test impl of numpy texture exporting because i needed to rewrite that for swizzle)

Issue #355, opened 2024-05-29
(Blender 4.0.1+) 'Euler' object has no attribute 'to_euler' Error when exporting a Level


Retro-Hax:
So as the Title implies i have gotten a very weird issue when it came to exporting my Hubworld to ultrasm64 specifically only this File it seems weirdly enough?
As every other File i export works out of the Box without any Issues i even attempted to copy the LevelRoot and AreaRoot of that broken File and copy them to another File which alos has a Level and there it exports without any Issues so my Question would be what exactly is causing this Bug if it is even a Bug?
![image](https://github.com/Fast-64/fast64/assets/69952967/0248d4ce-5f84-41c0-89d1-9400845f93e6)

![image](https://github.com/Fast-64/fast64/assets/69952967/dbbb5c50-7b66-43ae-8d69-89a87b3f0e5e)

My Level Exporter Settings are Refresh 13, YAY0 and the Export Level is set to Castle Grounds tho honestly even changing the Exported Level did not work :(

Romcad82:
I was just having this issue too, but I was able to export again when I removed the only waterbox from my level. Maybe try removing all waterboxes if you have any and see if that works. For now I just reverted to the previous version of Fast64 and it works fine

jesusyoshi54:
Has been fixed with recent PR

s4Ys369:
this fix needs to be applied to camera volumes as well :]

Lilaa3:
Thank you for catching that

PR #358, opened 2024-06-07
[SM64] Fixed Bug In Water Box Exports

jesusyoshi54:
bug introduced in #339 woops lol.
I just did something simple here, the swap between quats and euler objects should be isomorphic (I can't actually prove this it has been a while since maths) so there is no loss in info.


jesusyoshi54:
Fixes #355

Lilaa3 (approved):
Don´t say oops scutte its on me for assuming that was correct lol

PR #360, opened 2024-06-11
[OoT] Occlusion planes system

sauraen:
This adds f3d common and OoT support for a system for managing occlusion plane candidates for F3DEX3. You can create one or more meshes per room consisting of quads representing occlusion plane candidates, and they are exported as part of the room files. An [in-game system](https://github.com/HackerN64/HackerOoT/pull/132) then selects one occlusion plane per frame which is estimated to occlude as much as possible, computes the coefficients for that plane, and sends them to the RSP.

An occlusion plane candidate is a planar, convex quad, painted with vertex colors representing that plane's weight. You can manually weight planes based on the amount of stuff in the scene behind them. Colors black to white represent weights 0 to 1, and they multiply the score assigned to each plane per frame by the occlusion plane system. If you create a mesh marked as occlusion planes which does not meet these requirements, fast64 will give descriptive errors.

There is no change to exports for scenes which don't contain meshes marked as occlusion planes.


fast64_internal/f3d/occlusion_planes/exporter/to_c/occlusion_planes.py
```diff
@@ -0,0 +1,29 @@
+from mathutils import Vector
+
+from ...exporter import OcclusionPlaneCandidate, OcclusionPlaneCandidatesList
+from .....utility import CData
+
+
+def occVertexToC(vertex: Vector):
+    return "\t\t\t{" + ", ".join([str(int(round(a))) for a in vertex]) + "},\n"
```
Yanis42, line 8: now we avoid using tabs for exported stuff, instead you can use the ``indent`` variable from ``utility.py`` (defined as ``indent = " " * 4``)

fast64_internal/f3d/occlusion_planes/exporter/to_c/occlusion_planes.py
Yanis42, line 1: I think it would be better if these functions were a part of the classes it's using, like adding a ``to_c`` method to ``OcclusionPlaneCandidate`` and ``OcclusionPlaneCandidateList``

if you're concerned about the consistency of this don't worry, that's a thing I do a lot in the new exporter I made (#256, though it's named ``getC`` but now I prefer ``to_c``)

Yanis42 (commented):
nice 👀

fast64_internal/f3d/occlusion_planes/exporter/to_c/occlusion_planes.py
sauraen: Fixed.

fast64_internal/f3d/occlusion_planes/exporter/to_c/occlusion_planes.py
sauraen: I've changed this to how you describe.

Yanis42 (approved):

PR #364, opened 2024-06-16
Automatically select correct colorspace

sauraen:
Whenever a material is created or edited, or a scene is loaded, this PR sets the Blender colorspace settings so that preview of vertex colors is accurate (or more accurate than otherwise).

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -1693,7 +1693,7 @@ def get_color_input_update_callback(attr_name="", prefix=""):
 
 
 def update_node_values_of_material(material: Material, context):
-    nodes = material.node_tree.nodes
+    fixBlenderColorspace(context)
 
     update_blend_method(material, context)
     if not has_f3d_nodes(material):
@@ -1704,6 +1704,8 @@ def update_node_values_of_material(material: Material, context):
     update_combiner_connections(material, context)
 
     set_output_node_groups(material)
+
+    nodes = material.node_tree.nodes
 
     if f3dMat.rdp_settings.g_tex_gen:
         if f3dMat.rdp_settings.g_tex_gen_linear:
@@ -2178,8 +2180,9 @@ def has_f3d_nodes(material: Material):
 
 @persistent
 def load_handler(dummy):
-    logger.info("Checking for base F3D material library.")
+    fixBlenderColorspace(bpy.context)
 
+    logger.info("Checking for base F3D material library.")
     for lib in bpy.data.libraries:
         lib_path = bpy.path.abspath(lib.filepath)
 
@@ -2397,6 +2400,7 @@ def addColorAttributesToModel(obj: Object):
 
 
 def createF3DMat(obj: Object | None, preset="Shaded Solid", index=None):
+    fixBlenderColorspace(bpy.context)
     # link all node_groups + material from addon's data .blend
     link_f3d_material_library()
 
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -24,6 +24,7 @@ from bpy.types import (
     UILayout,
     VIEW3D_HT_header,
     World,
+    Event
 )
 from bl_operators.presets import AddPresetBase
 from bpy.utils import register_class, unregister_class
@@ -1693,7 +1694,7 @@ def get_color_input_update_callback(attr_name="", prefix=""):
 
 
 def update_node_values_of_material(material: Material, context):
-    fixBlenderColorspace(context)
+    bpy.ops.dialog.update_colorspace("INVOKE_DEFAULT")
 
     update_blend_method(material, context)
     if not has_f3d_nodes(material):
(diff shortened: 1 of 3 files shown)


Lilaa3:
Maybe an operator popup could work better than just automatically doing it on load? Some people have fast64 installed but have files that are not fast64 related, in which case they shouldn´t have their settings messed with without permission

sauraen:
You mean like something pops up when you load the scene asking whether you want to change it or not? I'm not sure how to do that or if that's possible.

Lilaa3:
See https://github.com/Fast-64/fast64/commit/50a62568c771d7df936a5d1eb1f5369038e8eed4
Only issue here is that i had to add a quick hack for update nodes to prevent mutiple popups, see if you like this

sauraen:
Thank you for the implementation @Lilaa3 . I've accepted that but fixed a bug with the conditions checking, and have it only offer to fix if you're loading a scene which has fast64 materials in it (so not when first opening Blender, which was annoying).

Lilaa3 (approved):
This is just about perfect I think, also that dialog description is way better than the one I suggested aha

fast64_internal/f3d/f3d_material.py
```diff
@@ -4491,6 +4549,7 @@ def mat_register():
 
     Scene.f3dUserPresetsOnly = bpy.props.BoolProperty(name="User Presets Only")
     Scene.f3d_simple = bpy.props.BoolProperty(name="Display Simple", default=True)
+    Scene.dont_ask_colorspace = bpy.props.BoolProperty(name="Don't ask again")
```
Dragorn421, line 4552: Can you move this prop to Fast64_Properties (so access like scene.fast64.dont_ask_colorspace)

fast64_internal/f3d/f3d_material.py
```diff
@@ -2440,6 +2442,61 @@ def reloadDefaultF3DPresets():
...
+            scene.display_settings.display_device = "sRGB"
+            scene.view_settings.view_transform = "Standard"
+            scene.view_settings.look = "None"
+            scene.view_settings.exposure = 0.0
+            scene.view_settings.gamma = 1.0
+            self.report({"INFO"}, "Updated color space.")
+            return {"FINISHED"}
+        except Exception as exc:
+            raisePluginError(self, exc)
+            return {"CANCELLED"}
```
Dragorn421, line 2495:
```suggestion
```
(just don't handle the exception if all that's done is (poorly) reraising it)

fast64_internal/f3d/f3d_material.py
```diff
@@ -2440,6 +2442,61 @@ def reloadDefaultF3DPresets():
...
+            UpdateColorSpacePopup.already_invoked = True
+            return context.window_manager.invoke_props_dialog(self, width=400)
+
+        return {"FINISHED"}
+
+    def draw(self, context: Context):
+        col = self.layout.column()
+        multilineLabel(
+            col,
+            f'The colorspace for the current scene "{context.scene.name}"\nwill cause inaccurate preview compared to N64.\nWould you like to update it?',
```
Dragorn421, line 2475: I would like to suggest simplifying this message in the most likely case of there being only a single Scene in the blend file

This would also help not confusing people in oot with blender scene / oot scene

Something like (ugly pseudocode)

```suggestion
            (f'The colorspace for the current scene "{context.scene.name}"' if len(bpy.data.scenes) >= 2 else "The colorspace settings\n") + '\nwill cause inaccurate preview compared to N64.\nWould you like to update it?',
```

fast64_internal/f3d/f3d_material.py
```diff
@@ -2440,6 +2442,61 @@ def reloadDefaultF3DPresets():
...
+    already_invoked = False  # HACK: used to prevent multiple dialogs from popping up
+
+    def invoke(self, context: Context, event: Event):
+        if UpdateColorSpacePopup.already_invoked:
+            return {"FINISHED"}
+        scene = context.scene
+        view_settings = scene.view_settings
+        # Check if color space is correct
+        if not scene.dont_ask_colorspace and (
+            scene.display_settings.display_device != "sRGB"
```
Dragorn421, line 2460: I would check this before invoking the operator rather than on invoke? Seems more logical and better for performance

fast64_internal/f3d/f3d_material.py
```diff
@@ -2440,6 +2442,61 @@ def reloadDefaultF3DPresets():
             update_preset_manual_v4(material, presetNameToFilename[material.f3d_mat.presetName])
 
 
+class UpdateColorSpacePopup(Operator):
+    bl_label = "Update Color Space"
+    bl_idname = "dialog.update_colorspace"
```
Dragorn421, line 2447: I can't easily look at other fast64 operators to check what convention we use but the bl_idname should be less generic, like throw in "fast64" in there

Dragorn421 (commented):
Nice
I have some small comments on the implementation

fast64_internal/f3d/f3d_material.py
Lilaa3: the point is to report and print traceback, returning CANCELLED also means that no undo entry is added, its done every where else in fast64 so why should it not be done here exactly?

Dragorn421: My bad, I thought raisePluginError raised an exception but it doesn't, only reports it

idk about not wanting to register an undo step, in theory props could have been changed by the operator by then

Dragorn421 (approved):
I pushed a commit addressing my review. So this counts as a half-approval since I'm approving my changes :sunglasses:

PR #366, opened 2024-06-24
[OoT] Fix latest decomp issues

Yanis42:
~~Currently importing mesh doesn't work and I didn't look into it yet~~


Yanis42:
(ready to review)

Dragorn421 (commented):
These changes are not backwards compatible with repos based on decomps before assets moved to extracted/ , right?
As in, if one's mod is based on a decomp revision where assets are in assets/, with these changes fast64 would no longer work in that mod's repo?

Yanis42:
well, yes, I thought of that and I figured people should use something up-to-date because of the better documentation and new features in the case of HackerOoT (and actually both since we pull decomp often)

I guess it shouldn't be that hard to make it backward compatible as the extracted path comes from a function :peepoShrug:

fast64_internal/oot/oot_level_parser.py
```diff
@@ -216,7 +216,14 @@ def parseScene(
     if not settings.isCustomDest and subfolder is None:
         importSubdir = os.path.dirname(getSceneDirFromLevelName(sceneName)) + "/"
 
-    sceneFolderPath = ootGetPath(importPath, settings.isCustomDest, importSubdir, sceneName, False, True)
+    sceneFolderPath = ootGetPath(
+        importPath + f"/{bpy.context.scene.fast64.oot.get_extracted_path()}",
```
sauraen, line 220: If you're using the f-string approach, why use a + here?

fast64_internal/oot/scene/operators.py
```diff
@@ -163,7 +163,9 @@ def execute(self, context):
                 exportInfo = ExportInfo(True, bpy.path.abspath(settings.exportPath), None, levelName)
             else:
                 if option == "Custom":
-                    subfolder = "assets/scenes/" + settings.subFolder + "/"
+                    subfolder = (
+                        context.scene.fast64.oot.get_extracted_path() + "assets/scenes/" + settings.subFolder + "/"
+                    )
```
sauraen, line 168: Change to the f-string approach shown in the next change below?

fast64_internal/oot/scene/panels.py
```diff
@@ -82,7 +82,9 @@ def draw(self, context):
 
         if removeSettings.option == "Custom":
             exportPath = (
-                context.scene.ootDecompPath + f"assets/scenes/{removeSettings.subFolder}/{removeSettings.name}/"
+                context.scene.ootDecompPath
+                + context.scene.fast64.oot.get_extracted_path()
+                + f"assets/scenes/{removeSettings.subFolder}/{removeSettings.name}/"
```
sauraen, line 87: More adds + f-strings?

sauraen (approved):
Some nits, otherwise looks okay. Also I hope you've tested this well.

fast64_internal/oot/scene/panels.py
Yanis42: this one doesn't fit on one line

fast64_internal/oot/scene/operators.py
Yanis42: it's a bit weird to use f-strings with the function as the value returned ends with ``/``

though it's fine for now, I think we should use pathlib more in general anyway so it can be addressed in a general cleanup (though I'll f-string the other parts)

Yanis42:
I tested again with HackerOoT (where I can experiment easier between before and after), everything works for me

fast64_internal/oot/scene/operators.py
Yanis42: I fixed the slash oddities but I didn't use f-strings eveywhere because of the line length limitation of the formatter

Yanis42:
decomp has a PR that lets you use the older asset system for mods, idk if this version change is really necessary now, I need to think about it

Dragorn421:
Checklist:

- nothing must write to or delete from `extracted/`, only from/to `assets/` and `spec`
  - export to `assets/` like previously
  - to delete assets, just remove from `spec` and don't touch `extracted/`
  - to replace assets, just export to `assets/` . A file in `assets/` has priority over a file with the same path under `extracted/version/`
  - to append data to an assets file, like for example to gkeep and link_animetion with exporting link animations, create files in `assets/` and include the corresponding .o in the `spec` (also keeping the existing includes, that will use the `extracted/version/` assets)
- implement importing from `extracted/`
  - there must be a choice of the version to import from (import from `extracted/version/`)
  - there must be an option to use "legacy" behavior of importing from `assets/`, for people using "older" decomp

Dragorn421 (commented):
Works: :clap:
- importing skeletons
- importing Link (child/adult)
- importing scenes (with or without cutscenes)
- exporting scenes
- importing non-link animations
- importing link animations: probably works but this branch lacks the import link anims fix (merge main)

Doesn't work: :chart_with_downwards_trend: 
- exporting skeletons
- exporting Link (child/adult)

Dragorn421:
I pushed a workaround for allowing exporting skeletons/Link

I call it a workaround and not a fix because I'm not super sure the behavior is ideal, but at least it can work

Dragorn421 (approved):
I don't think this PR addresses the numerous problems posed by the evolving assets system in the best way, but at least it works enough to be usable. Better than nothing!

krm01 (approved):
glanced briefly at the code but did not review it with any scrutiny, however tested this locally and everything i tested seemed to work.

PR #375, opened 2024-07-08
[F3D] Upgrade improvements (+related)

Lilaa3:
Changes:
1. Upgrades now copy over the current f3d library node tree onto the existing material, this has a lot of benifits such as no longer needing to purge materials, no naming changes, carying over all props, removal of hacks for material pointer propreties a̶n̶d̶ ̶t̶h̶e̶ ̶s̶p̶e̶e̶d̶ ̶o̶f̶ ̶u̶p̶g̶r̶a̶d̶e̶s̶  and on par speed (sometimes faster and sometimes slower?). Orphan data text is removed from upgrade popup.
2. Removed generateF3DNodeGraph.
3. Added warning for old materials and an early return for such.
image

4. Added "Recreate F3D Shader Nodes" operator, copies over the node tree from the f3d library and updates everything, useful for broken nodes of any kind.
5. Fixed f3d_light upgrading, along with now popping the props instead of just getting them.
6. Fixed recursiveCopyOldPropertyGroup which was interperting non set data as a non group, and trying to set new prop to it, it can also now fail and continue (for some reason not an issue before? maybe I should look into that).
7. Failed material upgrades now print traceback.
8. Fixed all the errors during draw layer updates by locking the updates and passing in a correct material context, fixed auto props error (Same as 6?? It should be having a lot of errors but it was somehow avoiding them previously?).

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -80,7 +80,6 @@ class F3D_GlobalSettingsPanel(bpy.types.Panel):
         prop_split(col, context.scene, "f3d_type", "F3D Microcode")
         col.prop(context.scene, "saveTextures")
         col.prop(context.scene, "f3d_simple", text="Simple Material UI")
-        col.prop(context.scene, "generateF3DNodeGraph", text="Generate F3D Node Graph For Materials")
         col.prop(context.scene, "exportInlineF3D", text="Bleed and Inline Material Exports")
         if context.scene.exportInlineF3D:
             multilineLabel(
@@ -243,22 +242,7 @@ class UpgradeF3DMaterialsDialog(bpy.types.Operator):
         layout = self.layout
         if self.done:
             layout.label(text="Success!")
-            box = layout.box()
-            box.label(text="Materials were successfully upgraded.")
-            box.label(text="Please purge your remaining materials.")
-
-            purge_box = box.box()
-            purge_box.scale_y = 0.25
-            purge_box.separator(factor=0.5)
-            purge_box.label(text="How to purge:")
-            purge_box.separator(factor=0.5)
-            purge_box.label(text="Go to the outliner, change the display mode")
-            purge_box.label(text='to "Orphan Data" (broken heart icon)')
-            purge_box.separator(factor=0.25)
-            purge_box.label(text='Click "Purge" in the top right corner.')
-            purge_box.separator(factor=0.25)
-            purge_box.label(text="Purge multiple times until the node groups")
-            purge_box.label(text="are gone.")
+            layout.label(text="Materials were successfully upgraded.")
             layout.separator(factor=0.25)
             layout.label(text="You may click anywhere to close this dialog.")
             return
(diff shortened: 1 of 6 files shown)


Lilaa3:
There is one very weird bug where materials upgraded via the f3d material panel (added in pr) will have bad nodes, weird

Lilaa3:
Last issue is fixed and I ended up removing my dumb active material hack and just used a temporary override

sauraen (changes requested):
I'm sorry, unfortunately this does not work. I've test-merged this into my F3DEX3 preview branch, which is doing upgrades from materials v5 to v6. When I try to do the upgrade on a test scene, I get a ton of errors like:
```
Execution type: <class 'KeyError'>

Execution value: 'bpy_prop_collection[key]: key "SceneProperties" not found'

Traceback: <traceback object at 0x7fb491811880>
Failed to upgrade VtxColMat
  File "/home/sauraen/.config/blender/3.2/scripts/addons/fast64/fast64_internal/f3d_material_converter.py", line 158, in convertF3DtoNewVersion
    update_all_node_values(material, bpy.context)  # Reload everything
  File "/home/sauraen/.config/blender/3.2/scripts/addons/fast64/fast64_internal/f3d/f3d_material.py", line 1410, in update_all_node_values
    update_node_values_without_preset(material, context)
  File "/home/sauraen/.config/blender/3.2/scripts/addons/fast64/fast64_internal/f3d/f3d_material.py", line 1420, in update_node_values_without_preset
    update_node_values(self, context, update_preset=False)
  File "/home/sauraen/.config/blender/3.2/scripts/addons/fast64/fast64_internal/f3d/f3d_material.py", line 1404, in update_node_values
    update_node_values_of_material(material, context)
  File "/home/sauraen/.config/blender/3.2/scripts/addons/fast64/fast64_internal/f3d/f3d_material.py", line 1791, in update_node_values_of_material
    update_fog_nodes(material, context)
  File "/home/sauraen/.config/blender/3.2/scripts/addons/fast64/fast64_internal/f3d/f3d_material.py", line 1589, in update_fog_nodes
    link_if_none_exist(material, nodes["SceneProperties"].outputs["FogColor"], nodes["FogColor"].inputs[0])
KeyError: 'bpy_prop_collection[key]: key "SceneProperties" not found'
```

And the resulting materials after the upgrade are broken. Some of them seem to work all right, but some are clearly wrong, including ones not using new EX3 features. Selecting one of these and clicking Recreate F3D Shader Nodes gives a very similar error. In that material's node tree, the error message is correct, there indeed is no SceneProperties node. This node is created in code, so I think you're doing something in the wrong order.

Upgrading v5 to v6 works correctly without the changes in this PR.

Lilaa3:
I may need to create scene properties before copying the nodes, sorry

sauraen (approved):
Works correctly now, thank you!

PR #376, opened 2024-07-09
[SM64] Refresh 16

Lilaa3:
Fix MARIO macro parsing, added function map
My regex skills are non existant so please feel free to make suggestions.


Lilaa3:
Issue fixed, now the parser will stop right before the goddard segment's contents like before

Lilaa3:
@jesusyoshi54 could you take a look at this?

jesusyoshi54 (approved):
regex

PR #377, opened 2024-07-10
[F3D] Fix RDP world defaults

Lilaa3:
T̶h̶i̶s̶ ̶i̶s̶ ̶t̶h̶e̶ ̶u̶c̶o̶d̶e̶ ̶d̶e̶f̶a̶u̶l̶t̶,̶ ̶s̶o̶ ̶h̶a̶v̶i̶n̶g̶ ̶t̶h̶i̶s̶ ̶s̶e̶t̶ ̶t̶o̶ ̶N̶O̶I̶S̶E̶ ̶i̶s̶ ̶a̶ ̶w̶e̶i̶r̶d̶ ̶c̶h̶o̶i̶c̶e̶ ̶t̶h̶a̶t̶ ̶a̶l̶s̶o̶ ̶b̶r̶e̶a̶k̶s̶ ̶S̶M̶6̶4̶ ̶(̶p̶l̶e̶a̶s̶e̶ ̶c̶o̶n̶f̶i̶r̶m̶ ̶t̶h̶i̶s̶ ̶d̶o̶e̶s̶n̶´̶t̶ ̶a̶f̶f̶e̶c̶t̶ ̶O̶O̶T̶ ̶p̶l̶e̶a̶s̶e̶)̶
World defaults need to vary across game modes, this pr will try to focus on that from now


gheskett (approved):

Lilaa3:
Draft until a good solution for game mode world defaults is implemented aha

__init__.py
```diff
@@ -355,10 +362,15 @@ def after_load(_a, _b):
...
 def gameEditorUpdate(self, context):
+    world_defaults = {}
     if self.gameEditorMode == "SM64":
-        self.f3d_type = "F3D"
+        self.f3d_type, world_defaults = "F3D", sm64_world_defaults
     elif self.gameEditorMode == "OOT":
-        self.f3d_type = "F3DEX2/LX2"
+        self.f3d_type, world_defaults = "F3DEX2/LX2", oot_world_defaults
+    elif self.gameEditorMode == "Homebrew":
+        self.f3d_type, world_defaults = "F3D", {}  # This will set some pretty bad defaults, but trust the user
```
Yanis42, line 371: nit: I don't really like doing ``var_a, var_b = value_1, value_2``

__init__.py
```diff
@@ -355,10 +362,15 @@ def after_load(_a, _b):
...
+    if self.world:
```
Yanis42, line 372: did you mean ``if self.world is not None`` or ``if len(self.world) > 0``? I think we should avoid doing ``if var`` it can bring a lot of confusion imo

fast64_internal/f3d/f3d_material.py
```diff
@@ -3596,6 +3596,31 @@ def key(self):
         return str(self.to_dict().items())
 
 
+def draw_rdp_world_defaults(layout: UILayout, scene: Scene):
+    world = scene.world
+    if not world:
```
Yanis42, line 3601: same comment about the condition stuff as earlier (hopefully this comment comes after the one I'm talking about lol)

fast64_internal/sm64/sm64_constants.py
```diff
@@ -2005,3 +2005,19 @@ def __init__(self, geoAddr, level, switchDict):
...
 ]
+
+sm64_world_defaults = {
+    "geometryMode": {
+        "zBuffer": True,
+        "shade": True,
+        "cullBack": True,
+        "lighting": True,
+        "shadeSmooth": True,
+    },
```
Yanis42, line 2016: I don't have a lot of knowledge but I think you should also add this to the oot world defaults since it defaults to false for all of these

__init__.py
Lilaa3: I do this relatively often so I'd like to ask why? I assume its a redability thing but I personally can easily following what's going on

fast64_internal/sm64/sm64_constants.py
Lilaa3: Already happening, when not set the from dict functions use the ucode defaults

__init__.py
gheskett: I'm not really a Python dev so can't speak on general consensus here, but in general code should be as readable as possible, as much as possible. Just because one line is easily understood by itself doesn't mean it won't contribute to an overwhelming function if these shortcuts are taken all over the place.

Yanis42:
where is ``world`` defined? from what I can tell it should be a blender prop since it's under bpy's ``Scene`` so it can't be ``None`` right? 🤔

Lilaa3:
> where is `world` defined? from what I can tell it should be a blender prop since it's under bpy's `Scene` so it can't be `None` right? 🤔

it can in fact be none, see #330 and my fix #334

Lilaa3:
I should add that the scene world is picked in the world properties tab
image

Yanis42 (approved):
OoT seems to work properly

Yanis42 (approved):

fast64_internal/sm64/sm64_constants.py
Lilaa3: Wait @Yanis42 I was going to merge just now but by this you meant oot defaults them to false right? Wanted to double check

Yanis42: sorry for the late answer I forgot TWICE in a row to reply lol stoopid brain

the oot defaults for ``geometryMode`` were the same as sm64 afaict so as long as it's stays on ``True`` it should be fine

PR #378, opened 2024-07-10
[SM64/F3D] Area fog improvements

Lilaa3:
Area fog is now a toggle in the area root
Renamed the text for global fog to area fog
Materials now by default show color and position if set fog is on, then in sm64 mode, if they are parented to an area that sets fog the option for area fog shows up with the area name in the text, if it's on a disabled column with the area's fog properties will also show
image
image
And now the toggle for global fog will not show in oot or homebrew!
image

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -2232,7 +2232,7 @@ class FAreaData:
 
 class FGlobalData:
     def __init__(self):
-        # dict of area index : FFogData
+        # dict of area index : FAreaData
         self.area_data = {}
         self.current_area_index = 1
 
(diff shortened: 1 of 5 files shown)


jesusyoshi54:
image
The area fog settings are a bit weird. First they use floats while the material fog uses ints. They should match. Second you say "Set Area Fog" in the hint text but the property box is "Set Fog Settings".
I think the message itself is too wordy and could be shorter. Just say something to the effect of, "Fog properties will be applied to all materials with fog enabled that use area fog"

Lilaa3:
> The area fog settings are a bit weird. First they use floats while the material fog uses ints. They should match. 
Can´t do that without a prop upgrade, using step=100 makes s a fake int and doesn´t require any upgrades so I'm gonna commit that for now, if you think it's worth it to do this properly do say it.
> Second you say "Set Area Fog" in the hint text but the property box is "Set Fog Settings". I think the message itself is too wordy and could be shorter. Just say something to the effect of, "Fog properties will be applied to all materials with fog enabled that use area fog"
I tried making the description shorter, but one of the main things I was trying to change was the whole "global" description which honestly was very missleading

jesusyoshi54 (approved):
yeah

jesusyoshi54:
I made a PR to this branch for conflict, check it out

PR #379, opened 2024-07-12
[F3D] Don´t hide properties that will get exported (specifically geo modes)

Lilaa3:
There was two options in fixing this:
Not exporting hidden modes or always showing all modes
I went with the latter in this pr as its more logical, nonsense example but if a game always defaults to packed normals but doesn´t default to lighting, that should still be setable AND visible in the world settings. Or maybe one could use a hack that revolves around checking for an unused mode being on. In my opinion this is the best way to do it and it doesn´t add much bloat to the UI even in ex3.
I also fixed a bug where if texgen was on without lighting it would still emulate texgen, even though it was neither visible or correct

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -411,14 +411,14 @@ def ui_geo_mode(settings, dataHolder, layout, useDropdown):
     if not useDropdown or dataHolder.menu_geo:
         disable_dependent = False  # Don't disable dependent props in world defaults
 
-        def indentGroup(
-            parent: UILayout, textOrProp: Union[str, "F3DMaterialProperty"], isText: bool, enable=True
-        ) -> UILayout:
+        def indentGroup(parent: UILayout, textOrProp: Union[str, "F3DMaterialProperty"], isText: bool) -> UILayout:
             c = parent.column(align=True)
             if isText:
                 c.label(text=textOrProp)
+                enable = True
             else:
                 c.prop(settings, textOrProp)
+                enable = getattr(settings, textOrProp)
             c = c.split(factor=0.1)
             c.label(text="")
             c = c.column(align=True)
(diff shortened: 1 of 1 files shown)


sauraen:
I don't agree with this PR. When I redesigned the geometry mode UI (which was needed for F3DEX3, but applies to all microcodes), I found it confusing to remember which features depended on which other features, even for features I made myself. Hiding features which you cannot have without other features, e.g. things dependent on lighting, is an easy way to represent this; the alternative would be to have a ton of warnings and no structure in the UI.

It is true that you can import and export a state which is not visible in the UI, e.g. texgen without lighting. However:
1. You can actually edit this state, e.g. by enabling lighting, enabling/disabling texgen, then disabling lighting.
2. While the UI does not correctly reflect the F3D display list contents imported/exported, it does correctly reflect the in-game rendering results on N64. For example, if the display list says texgen without lighting, you will get no texgen and no lighting, which matches what is shown in the UI.

sauraen:
If you absolutely must have them visible--for example, because you want people to know where the texgen option is while they have lighting disabled, otherwise they can't find it--then I'd suggest you change the code so that instead of hiding the currently-invalid options, it grays them out (disables them) instead.

Lilaa3:
I think disabling works quite well, the identation makes whats happening pretty obvious I think, decided to not disable in world defaults and fixed my typo in the node update (g_texture_gen instead of g_tex_gen)

Lilaa3:
Also world defaults ui is a broken in EX3 #370

fast64_internal/f3d/f3d_material.py
```diff
@@ -409,18 +409,20 @@ def ui_geo_mode(settings, dataHolder, layout, useDropdown):
...
+        def indentGroup(
+            parent: UILayout, textOrProp: Union[str, "F3DMaterialProperty"], isText: bool, enable=True
+        ) -> UILayout:
             c = parent.column(align=True)
             if isText:
                 c.label(text=textOrProp)
             else:
                 c.prop(settings, textOrProp)
-                if not getattr(settings, textOrProp):
-                    return None
```
sauraen, line 420: Enable should not be a parameter, enable should be coming from this, that's the whole point of this function, to enable/disable the group based on whether the head of the group is enabled.

fast64_internal/f3d/f3d_material.py
```diff
@@ -458,11 +460,7 @@ def indentGroup(parent: UILayout, textOrProp: Union[str, "F3DMaterialProperty"],
             shadeColorLabel = "Lighting * vertex color"
         else:
             shadeColorLabel = "Lighting"
-        if lightFxPrereq:
-            c = indentGroup(inputGroup, f"Shade color = {shadeColorLabel}:", True)
-            c.prop(settings, "g_fresnel_color")
-        else:
-            inputGroup.column().label(text=f"Shade color = {shadeColorLabel}")
+        inputGroup.label(text=f"Shade color = {shadeColorLabel}:")
```
sauraen, line 463: Please remove the colon from the end of this line

sauraen (approved):
Please fix one tiny issue, otherwise approved. I'm still not sure this is ideal but if it unblocks the world defaults stuff then it's okay.

PR #381, opened 2024-07-17
[F3D] Hack for forced 4.2 node additions

Lilaa3:
Not the prettiest solution but it is safe and fixes blender 4.2, biggest issue is that 4.2 takes a while to compile F3D materials but it is usable after it finishes

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -345,10 +345,28 @@ def upgrade_scene_props_node():
         bpy.ops.dialog.upgrade_f3d_materials("INVOKE_DEFAULT")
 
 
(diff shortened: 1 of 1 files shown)


Lilaa3:
A second option is to use recreate f3d nodes now in main on any file with f3d materials, in version 4.2 or above and with no toggled prop for being already fixed, that prop would be added into settings and be called something like internal_fixed_420
this would be slower however

Lilaa3:
There is more that needs to be done, waiting on sauren's input on discord because the node library needs an update

Lilaa3:
fixed in#382

PR #382, opened 2024-07-19
Improved nodes preview: F3DEX3 and OoT lighting

sauraen:
Preview of materials in the 3D view now supports the following F3DEX3 features:
- Packed normals (vertex colors + lighting)
- Ambient occlusion (ambient and directional)
- Fresnel (color and/or alpha)
- Specular lighting
- Light to alpha

The following F3DEX3 features are still not previewed correctly:
- Cel shading
- Attribute offsets

Also:
- Adds preview of a second directional light, enabling correct OoT time of day lighting
- Fixes OoT fog preview
- Fixes OoT light directions in the preview
- Fixes an error when using OoT light objects

Also includes completely unrelated changes from @Lilaa3 about alpha clip in Blender 4.2, but we wanted to avoid two material upgrade steps for everyone if we separated the PRs.


fast64_internal/render_settings.py
```diff
@@ -10,7 +10,6 @@ def on_update_sm64_render_settings(self, context: bpy.types.Context):
         area: bpy.types.Object = renderSettings.sm64Area
         renderSettings.fogPreviewColor = tuple(c for c in area.area_fog_color)
         renderSettings.fogPreviewPosition = tuple(round(p) for p in area.area_fog_position)
-
```
Lilaa3, line 13: unneeded new line

fast64_internal/utility.py
```diff
@@ -1321,6 +1321,12 @@ def gammaCorrect(linearColor):
     return list(mathutils.Color(linearColor[:3]).from_scene_linear_to_srgb())
 
 
+def gammaCorrectAlpha1Tuple(linearColor):
+    correct = gammaCorrect(linearColor)
```
Lilaa3, line 1325: maybe rename to sRGB or s_rgb color

Lilaa3 (commented):
A more general comment I have is that I think new props, functions, etc should start using snake_case, I'm on mobile rn and want to test packed normals a bit since that is one property I neglected to test last time

Lilaa3:
I know this is not up for review yet (is there a reason that this is in draft?) just had the time to go ahead and comment, I tested this pr again and it's working great except for one more thing i noticed. Currently convertF3DtoNewVersion is still using 5 for its updated version, I'd suggest creating a constant in f3d_material.py
`F3D_MAT_CUR_VERSION = 6`
and then using that in `createF3DMat`, `convertF3DtoNewVersion` and`MatUpdateConvert.version`, you can also replace the use of `MatUpdateConvert.version` for the constant in init.py in the UpgradeF3DMaterialsDialog operator and upgrade_scene_props_node

fast64_internal/render_settings.py
sauraen: I agree, this is a deletion.

fast64_internal/utility.py
Lilaa3: I was refering to `correct` here

fast64_internal/render_settings.py
```diff
@@ -191,30 +184,62 @@ class Fast64RenderSettings_Properties(bpy.types.PropertyGroup):
...
+        default=mathutils.Vector((-1.0, 1.0, -1.0)).normalized(),  # pre normalized
+        update=on_update_render_preview_nodes,
+    )
+    light1SpecSize: bpy.props.IntProperty(
+        name="Light 1 Specular Size",
+        min=1,
+        max=255,
+        default=3,
         update=on_update_render_preview_nodes,
     )
```
Lilaa3, line 236: should these be snake case?

more importantly maybe you should add upgrade code
```
def upgrade_changed_props():
        for scene in bpy.data.scenes:
            upgrade_old_prop(self, "light0Color", self, "lightColor")
            upgrade_old_prop(self, "light0Direction", self, "lightDirection"
)
```

fast64_internal/render_settings.py
```diff
@@ -191,30 +184,62 @@ class Fast64RenderSettings_Properties(bpy.types.PropertyGroup):
...
+        default=3,
+        update=on_update_render_preview_nodes,
+    )
+    light1Color: bpy.props.FloatVectorProperty(
+        name="Light 1 Color",
+        subtype="COLOR",
+        size=4,
+        min=0,
+        max=1,
+        default=(0, 0, 1, 1),
```
Lilaa3, line 218: shouldn´t the default be black? i noticed this before but just thought it didn´t matter but then I remembered that sm64 lighting engine exists aha, by default it has only one global light direction

Lilaa3 (changes requested):
I already tested all of this, just noticed three small things, I'm sorry for not commenting on them earlier and delaying this further

fast64_internal/render_settings.py
sauraen: Fixed

fast64_internal/render_settings.py
sauraen: I have added some upgrade code and checked that it works.

As far as the names, since all the other properties I didn't change in this object are camelCase, I think it's best to keep the new names consistent with the others.

Lilaa3 (approved):
lgtm

Lilaa3:
Tested one last time and merged, great work!

PR #383, opened 2024-07-19
[Codebase] Small generalized upgrade function

Lilaa3:
Makes upgrade code cleaner and easier to read while adding safety checks, handaling popping, never setting if the old prop didn´t exist, cases where there can be multiple props, changed enum definitions and the dreaded forced base 16 int string
Fixes random invalid decomp path exceptions in sm64


sauraen (approved):
Tested that this does eliminate the errors about missing SM64 decomp path in blend files which don't have any SM64 content, including the blender default scene at startup.

I have not tested the SM64 specific stuff.

PR #385, opened 2024-07-21
[F3D] Show tlut othermode in materials and world defaults

Lilaa3:
The mode is still automatically set when using CI textures, however the default can now be changed and it hides what's happening under the hood just a bit less,  I also removed some code from f3d_bleed, specifically I removed reset_command_dict (unused) and DPSetTextureLUT from the cmd list in add_reset_cmd because it is always set by SPSetOtherMode like other othermodes.
image
image

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -2095,7 +2095,7 @@ def update_tex_values_manual(material: Material, context, prop_path=None):
 
     textures = [f3dMat.tex0] if useDict["Texture 0"] else []
     textures += [f3dMat.tex1] if useDict["Texture 1"] else []
-    ci_formats = list(set(tex.ci_format for tex in textures if tex.tex_format.startswith("CI")))
+    ci_formats = list(set(tex.ci_format if tex.tex_format.startswith("CI") else "NONE" for tex in textures))
     if len(ci_formats) != 1:  # if more than one ci format or no ci format textures
         ci_formats = ["NONE"]
     f3dMat.rdp_settings.g_mdsft_textlut = "G_TT_" + ci_formats[0]
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -3554,6 +3554,7 @@ class RDPSettings(PropertyGroup):
         data = self.attributes_to_dict(self.other_mode_h_attributes)
         if lut_format:
             data["lutFormat"] = lut_format
+        return data
 
     def other_mode_h_from_dict(self, data: dict):
         self.attributes_from_dict(data, self.other_mode_h_attributes)
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -2088,7 +2088,7 @@ def get_textlut_mode(f3d_mat: "F3DMaterialProperty", inherit_from_tex: bool = Fa
     tlut_modes = [tex.ci_format if tex.tex_format.startswith("CI") else "NONE" for tex in textures]
     if tlut_modes and tlut_modes[0] == tlut_modes[-1]:
         return "G_TT_" + tlut_modes[0]
-    return None if inherit_from_tex else f3d_mat.rdp_settings.g_mdsft_texlut
+    return None if inherit_from_tex else f3d_mat.rdp_settings.g_mdsft_textlut
 
 
 def update_tex_values_manual(material: Material, context, prop_path=None):
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -546,6 +546,9 @@ def ui_upper_mode(settings, dataHolder, layout: UILayout, useDropdown):
         prop_split(inputGroup, settings, "g_mdsft_combkey", "Chroma Key")
         prop_split(inputGroup, settings, "g_mdsft_textconv", "Texture Convert")
         prop_split(inputGroup, settings, "g_mdsft_text_filt", "Texture Filter")
+        textlut_col = inputGroup.column()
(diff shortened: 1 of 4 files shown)


jesusyoshi54:
If your texture is not set, then you should be able to set the value of the TextureLUT property. This would make it so that the TextureLUT property is properly reverted to normal in the case where a texture is loaded via some function node. This should also fix the bug where unset textures would get defaulted to `G_TT_NONE` while using the mat write method `write_all` (bleed and DL exporter option) or if you set the default to use LUT and had the unset texture apply `G_TT_NONE`

fast64_internal/f3d/f3d_material.py
```diff
@@ -3515,16 +3542,19 @@ def geo_mode_from_dict(self, data: dict):
...
         ("perspectiveCorrection", "g_mdsft_textpersp", "G_TP_NONE"),
         ("cycleType", "g_mdsft_cycletype", "G_CYC_1CYCLE"),
         ("pipelineMode", "g_mdsft_pipeline", "G_PM_NPRIMITIVE"),
     ]
 
-    def other_mode_h_to_dict(self):
-        return self.attributes_to_dict(self.other_mode_h_attributes)
+    def other_mode_h_to_dict(self, lut_format=None):
+        data = self.attributes_to_dict(self.other_mode_h_attributes)
+        if lut_format:
```
jesusyoshi54, line 3555: why is this here? As far as I can tell it isn't used

fast64_internal/f3d/f3d_material.py
```diff
@@ -2070,13 +2081,29 @@ def get_tex_gen_size(tex_size: list[int | float]):
...
+    use_dict = all_combiner_uses(f3d_mat)
+    textures = [f3d_mat.tex0] if use_dict["Texture 0"] and f3d_mat.tex0.tex_set else []
+    textures += [f3d_mat.tex1] if use_dict["Texture 1"] and f3d_mat.tex1.tex_set else []
+    if textures:
+        ci_formats = list(set(tex.ci_format if tex.tex_format.startswith("CI") else "NONE" for tex in textures))
+        if len(ci_formats) == 1:
+            return "G_TT_" + ci_formats[0]
+        elif len(ci_formats) == 0:
+            return "G_TT_NONE"
+    if not none_if_not_set:
```
jesusyoshi54, line 2094: I hate this variable name.

fast64_internal/f3d/f3d_material.py
```diff
@@ -2070,13 +2081,29 @@ def get_tex_gen_size(tex_size: list[int | float]):
     return (tex_size[0] - 1) / 1024, (tex_size[1] - 1) / 1024
 
 
+def get_textlut_mode(f3d_mat: "F3DMaterialProperty", none_if_not_set: bool = False):
+    use_dict = all_combiner_uses(f3d_mat)
+    textures = [f3d_mat.tex0] if use_dict["Texture 0"] and f3d_mat.tex0.tex_set else []
+    textures += [f3d_mat.tex1] if use_dict["Texture 1"] and f3d_mat.tex1.tex_set else []
+    if textures:
```
jesusyoshi54, line 2088: this is a confusing way to pass None if the two ci_format attributes are different between the texture properties but only on the UI. I think you should just call out each case so it is more clear what is going on here. Either you inherit the property when it is coming from a tex, it comes from the property, or you return None. If you spell out each case directly in the if statements it is more clear what is going on and why each thing is returned.

```suggestion
ci_formats = [tex.ci_format if tex.tex_format.startswith("CI") for tex in textures]
if not ci_formats and textures:
    return "G_TT_NONE"
if len(ci_formats) == 1 or ci_formats[0] == ci_formats[1]:
    return "G_TT_" + ci_formats[0]
else:
     # I am changing your var `none_if_not_set` name
     return None if inherit_from_tex else f3d_mat.rdp_settings.g_mdsft_texlut
```

fast64_internal/f3d/f3d_material.py
```diff
@@ -546,6 +546,17 @@ def ui_upper_mode(settings, dataHolder, layout: UILayout, useDropdown):
         prop_split(inputGroup, settings, "g_mdsft_combkey", "Chroma Key")
         prop_split(inputGroup, settings, "g_mdsft_textconv", "Texture Convert")
         prop_split(inputGroup, settings, "g_mdsft_text_filt", "Texture Filter")
+        textlut_col = inputGroup.column()
+        tlut_mode = get_textlut_mode(dataHolder, True) if isinstance(dataHolder, F3DMaterialProperty) else None
+        if tlut_mode:
+            textlut_col.enabled = False
```
jesusyoshi54, line 552: is there a reason you couldn't just do a normal prop split here after disabling it?

jesusyoshi54 (changes requested):

fast64_internal/f3d/f3d_material.py
Lilaa3: would be a good catch but it will be used aha

fast64_internal/f3d/f3d_material.py
Lilaa3: (sorry mobile made me respond to the wrong comment) this will be used later in the gltf pr, wasnt a mistake but if it were it would've been a good catch

fast64_internal/f3d/f3d_material.py
Lilaa3: @jesusyoshi54 your code didn´t have the same functionality, so here is what I did, I renamed ci_formats to tlut_modes and kept the NONE for non ci's, then if there is atleast one mode i compare the first element to the last (I think it keeps the logic explicit without adding another len check?), this shortned the code a lot and still made things more explicit hopefully, if you still dislike it I will change it

jesusyoshi54: yeah coding in the git UI is sort of cringe

jesusyoshi54 (approved):

PR #387, opened 2024-07-21
[F3D Bleed] Fix #386 - Make FImage hashable (now a dataclass)

### LINKED ISSUE #386: [F3D] Mat Bleed reloads textures when using texture reference
If you use the same texture reference on multiple materials in a mesh with mat bleed, it should bleed all the textures, but this does not occur. Instead all textures are loaded each time.

Lilaa3:
Convert FImage to a dataclass for proper `__eq__` method
Fixes #386


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3238,18 +3238,23 @@ def spc(x):
...
+    filename: str
+    data: bytearray = field(init=False, compare=False)
+    startAddress: int = field(init=False, compare=False)
+    isLargeTexture: bool = field(init=False, compare=False)
+    converted: bool = field(init=False, compare=False)
+
+    def __post_init__(self):
+        self.converted = self.isLargeTexture = False
         self.startAddress = 0
         self.data = bytearray(0)
```
Dragorn421, line 3257:
```suggestion
    data: bytearray = field(init=False, compare=False, default_factory=bytearray)
    startAddress: int = field(init=False, compare=False, default=0)
    isLargeTexture: bool = field(init=False, compare=False, default=False)
    converted: bool = field(init=False, compare=False, default=False)
```

Lilaa3: Honestly these should have init true, I will modify the suggestion and commit it

Lilaa3: Btw your code would not use the default, the reason I had it like this was so the values would get initialized but they couldn´t be passed to the constructor

Dragorn421: feel free to change

fwiw afaict my code would have worked?
for example:
```py
from dataclasses import dataclass, field
@dataclass
class Thing:
    foo = field(init=False, default="bar")
print(Thing())  # Thing(foo='bar')
Thing("baz")  # TypeError: Thing.__init__() takes 1 positional argument but 2 were given
```
, it seems to work as expected

Lilaa3: > If init is true, the field will be a parameter to the class's __init__() function.
I guess this doesn´t imply anything about if the value will be initialized, thanks

Dragorn421 (approved):

PR #389, opened 2024-07-24
T3D UI

Lilaa3:
Implement UI for tiny 3d, the fast libdragon 3d microcode


jesusyoshi54:
please provide a description :)

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -133,6 +134,14 @@ def isUcodeF3DEX3(F3D_VER: str) -> bool:
     return F3D_VER == "F3DEX3"
 
 
+def is_ucode_t3d(UCODE_VER: str) -> bool:
+    return UCODE_VER == "T3D"
+
+
+def is_ucode_f3d(UCODE_VER: str) -> bool:
+    return UCODE_VER not in {"T3D"}
```
Yanis42, line 142: you could do ``return not is_ucode_t3d(UCODE_VER)`` but I'd understand if you don't like it

fast64_internal/f3d/f3d_material.py
```diff
@@ -416,7 +488,24 @@ def ui_geo_mode(settings, dataHolder, layout, useDropdown):
     if not useDropdown or dataHolder.menu_geo:
         disable_dependent = False  # Don't disable dependent props in world defaults
 
+        def should_draw(*props):
```
Yanis42, line 491: I'm gonna let Python experts decide if this is okay or not, I think it looks weird but maybe it's fine idk, also fwiw I'm not a fan of nested functions like this

__init__.py
```diff
@@ -87,7 +87,7 @@ def poll(cls, context):
     def draw(self, context):
         col = self.layout.column()
         col.scale_y = 1.1  # extra padding
-        prop_split(col, context.scene, "f3d_type", "F3D Microcode")
+        prop_split(col, context.scene, "f3d_type", "Microcode")
```
Yanis42, line 90: I find it weird to remove "F3D" from the display name there but keep it for the internal stuff, like the prop name ``f3d_type`` should be renamed to ``microcode_type`` or something similar

I guess it's more of a cleanup thing though so feel free to ignore this

Yanis42 (commented):
same here don't understand everything but here's my small code review

fast64_internal/f3d/f3d_gbi.py
Lilaa3: no because in the future we could have more non f3d ucodes, maybe a model64 option for barebones materials

Yanis42: oh yeah good point

__init__.py
Lilaa3: I'd disagree, every internal change to a proprety is just more upgrade code

Yanis42: it would make more sense tbh, but yeah it's not the point of this PR so it can be done later

fast64_internal/f3d/f3d_material.py
Lilaa3: yeah I really want some opinions on this part of the pr especially

Dragorn421: I like how readable it is

Dragorn421 (commented):
The code looks alright, would need some reviews of how it looks in blender

Lilaa3:
Has anyone tested? @HailToDodongo

HailToDodongo:
Tested it again and it works fine.
imo this can be merged, the ongoing discussion about the UI/settings would make their own PRs

Dragorn421 (approved):

PR #391, opened 2024-07-25
[F3D] Fix 4.2 node additions

Lilaa3:
While sauren's pr includes my node fixes for 4.2, it is possible for someone to upgrade a v5 material to v6 AND THEN switch to 4.2, which would result in the black materials we all know and hate


Reonu (approved):
Tested it on a few blends by upgrading them to v6 materials on 4.1 and then opening them on 4.2 with this PR. The PR works perfectly, the materials load correctly.

imagen

Yanis42 (approved):
works on my machine™

PR #394, opened 2024-07-27
[SM64] remove " after ultra64 include

Lilaa3:
Only happens for level data


jesusyoshi54 (approved):
the quotation was removed

PR #395, opened 2024-07-27
[SM64] Fix usage of sm64 scale in animations

Lilaa3:
Happened because of one fix before settings rework, code changed by the fix was never updated to use the new prop


jesusyoshi54 (approved):
animation export works now :)

Issue #291, opened 2024-01-11
[OoT] Importing Player animations causes an error


Yanis42:
To reproduce try to import a player anim, ``gPlayerAnim_link_fighter_upper_pierce_kiru_end`` works for this error, I searched a bit it and looks like negative numbers is the issue? I'm not sure how it is supposed to work though, hence this issue

The error is inside ``getFrameData`` (``oot_anim.py``) in the ``frameData`` list when it tries to convert the str to an int, when it's a negative number (like "-0x00A2") it's reading "x0" for the first int value inside ``from_bytes``

<details close>
<summary>Traceback of the error</summary>

```
Traceback (most recent call last):
  File "C:\Users\Yanis\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\fast64\fast64_internal\oot\animation\operators.py", line 172, in execute
    ootImportAnimationC(armatureObj, settings, actorScale)
  File "C:\Users\Yanis\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\fast64\fast64_internal\oot\animation\operators.py", line 97, in ootImportAnimationC
    ootImportLinkAnimationC(
  File "C:\Users\Yanis\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\fast64\fast64_internal\oot\oot_anim.py", line 528, in ootImportLinkAnimationC
    frameData = getFrameData(animFilepath, animData, frameDataName)
  File "C:\Users\Yanis\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\fast64\fast64_internal\oot\oot_anim.py", line 634, in getFrameData
    frameData = [
  File "C:\Users\Yanis\AppData\Roaming\Blender Foundation\Blender\4.0\scripts\addons\fast64\fast64_internal\oot\oot_anim.py", line 635, in <listcomp>
    int.from_bytes([int(value.strip()[2:4], 16), int(value.strip()[4:6], 16)], "big", signed=True)
ValueError: invalid literal for int() with base 16: 'x0'
```
</details>

Yanis42:
replacing the ``frameData`` generator by the code below seems to fix the issue, but again, not an anim expert so I'd like to get someone with better knowledge than me looking into this
```py
frameData = []
    for value in data.split(","):
        value = value.strip()
        if value != "":
            isNegative = "-0x" in value
            value = value.replace("-0x", "0x") if isNegative else value
[...]

PR #397, opened 2024-07-29
[OoT] Player animation import fix

Yanis42:
idk if this is ideal but it works so... 🤷 (I'm not an anim expert btw), fix for #291


Dragorn421:
I have a better fix, closing

PR #398, opened 2024-07-31
[OoT] Fixed room actors being duplicated

Yanis42:
found by @Reonu, it never happened to me when testing with my usual test blend :)


Reonu (approved):
tested, works well, fixes the bug

Dragorn421 (approved):

PR #399, opened 2024-08-02
[F3D] Pass in shade alpha into fog blender instead of disabling it when fog is off

Lilaa3:
Fixes preview of fogshade with no fog geo mode enabled
image
image

### DIFF
--- a/fast64_internal/f3d/f3d_material.py
+++ b/fast64_internal/f3d/f3d_material.py
@@ -1574,21 +1574,17 @@ def update_node_combiner(material, combinerInputs, cycleIndex):
 def update_fog_nodes(material: Material, context: Context):
     nodes = material.node_tree.nodes
     f3dMat: "F3DMaterialProperty" = material.f3d_mat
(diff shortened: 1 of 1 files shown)


Lilaa3:
To show the change a bit better, here are 4 materials, one with fog and a fog blender, fog with no fog in the blender, no fog but fog in the blender and neither
Main:
image
PR:
image
[Here is the test file, it's in 3.2, upgrade materials in main or in this pr to see how they differ in behavior](https://github.com/user-attachments/files/16473205/Fog.Blender.Test.blend.zip)

Reonu (approved):
I tested it and seems to work well, fixes the issue I had on my blend, lila's blend also works as expected

jesusyoshi54:
image
if you swap global fog it doesn't update the links properly

Lilaa3:
@jesusyoshi54 Issue has been fixed, also added two comments

jesusyoshi54 (approved):
Works well and code looks nice :)

PR #401, opened 2024-08-05
[F3D] Edge case for decals, use hashed for cutout materials instead of blend + more

Lilaa3:
Main change is the title, but I also updated rendermode_preset_to_advanced to actually update the settings based on layer, not only that but the world defaults are used for layers now


Lilaa3:
This should also simplify [this function in f64render](https://github.com/HailToDodongo/f64render/blob/216d359e98c196521e5d6dd7b11dec74030e8f3d/material/parser.py#L60-L93)

fast64_internal/f3d/f3d_material.py
```diff
@@ -180,17 +169,52 @@ def update_draw_layer(self, context):
...
+    else:
+        game_mode = scene.gameEditorMode
+        layer = getattr(f3d_mat.draw_layer, game_mode.lower(), None)
+        if layer is None:  # Game mode has no layer, don´t change anything
+            return
+
+        # If no world defaults for game mode, fallback on get_with_default default arg
+        cycle_1, cycle_2 = default_draw_layers.get(game_mode, {}).get(layer, ("", ""))
+        if scene.world is not None and game_mode in world_layer_defaults_attrs:
+            possible_cycle_1, possible_cycle_2 = world_layer_defaults_attrs[game_mode](scene.world, layer)
```
Yanis42, line 208: this looks really weird to me, I think it would be better if ``world_layer_defaults_attrs`` was a real function

Yanis42 (commented):
can't say I understand what this does but here's my tiny contribution to the review

fast64_internal/f3d/f3d_material.py
```diff
@@ -1428,6 +1455,12 @@ def update_all_node_values(material, context):
     update_rendermode_preset(material, context)
 
 
+def update_all_material_nodes(self, context):
+    for material in bpy.data.materials:
+        with context.temp_override(material=material):
+            update_all_node_values(material, context)
```
Dragorn421, line 1461: this is a bit too trigger-happy, maybe at least check the material is a f3d material before putting it through this

fast64_internal/oot/f3d/properties.py
```diff
@@ -149,12 +150,12 @@ def draw_props(self, layout: UILayout, mat: Object, drawLayer: str):
 
 class OOTDefaultRenderModesProperty(PropertyGroup):
     expandTab: BoolProperty()
-    opaqueCycle1: StringProperty(default="G_RM_AA_ZB_OPA_SURF")
-    opaqueCycle2: StringProperty(default="G_RM_AA_ZB_OPA_SURF2")
-    transparentCycle1: StringProperty(default="G_RM_AA_ZB_XLU_SURF")
-    transparentCycle2: StringProperty(default="G_RM_AA_ZB_XLU_SURF2")
-    overlayCycle1: StringProperty(default="G_RM_AA_ZB_OPA_SURF")
-    overlayCycle2: StringProperty(default="G_RM_AA_ZB_OPA_SURF2")
+    opaqueCycle1: StringProperty(default="G_RM_AA_ZB_OPA_SURF", update=update_all_material_nodes)
```
Dragorn421, line 153: for readability can you add an indirection, like on props put `update=on_world_defaults_change`
and create the function `on_world_defaults_change` in which `update_all_material_nodes` is called

Dragorn421:
btw why not alpha clip for cutout materials

Lilaa3:
> btw why not alpha clip for cutout materials

sure i can do that for pre 4.2, wont make a visual difference tho

Lilaa3:
Currently fast64 main is also assuming fog when no rendermode is set btw, this is was mostly hidden before i removed the requirement for the geo mode to activate the fog blender (i say mostly because it caused unset rendermode sm64 materials to assume the fog blender was on even without setting the rendermode lol), it only showed now but this pr accidentally fixes it

fast64_internal/f3d/f3d_material.py
```diff
@@ -180,17 +169,53 @@ def update_draw_layer(self, context):
...
     """
-    settings = material.f3d_mat.rdp_settings
+    scene = bpy.context.scene
+    f3d_mat = material.f3d_mat
+    settings = f3d_mat.rdp_settings
     f3d = get_F3D_GBI()
 
-    if settings.rendermode_advanced_enabled:
+    if settings.rendermode_advanced_enabled and settings.set_rendermode:
         # Already in advanced mode, don't overwrite this with the preset
```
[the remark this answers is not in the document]
Lilaa3: Could you give me feedback on if the comments are good now?

fast64_internal/oot/f3d/properties.py
Lilaa3: If I'm gonna add something seperate I might as well only call update_rendermode_preset in there

Dragorn421 (approved):

PR #402, opened 2024-08-06
[SM64] Fix string to macros regression

Lilaa3:
I take responsibility for assuming my main merge was correct, sorry
Scut changed macro to function in the combined exporter pr which was merged first


jesusyoshi54 (approved):
not only is this yes but I also tested yes and yes

PR #405, opened 2024-08-06
[SM64] Fix combined exporter level regressions

Lilaa3:
Non decomp relative custom levels are now exportable, custom level names no longer fail to be added to the level define macro lis, export all selected mode will no longer throw errors in the ui when no object is selected, create level layout now uses the correct new prop making it functional again


Dragorn421:
This was caused by merging #284 right? and it didn't conflict as far as git was concerned, but there in fact were incompatible changes made to main meanwhile

any tester in chat (at random, @Reonu ?! :) )

Lilaa3:
> This was caused by merging #284 right? and it didn't conflict as far as git was concerned, but there in fact were incompatible changes made to main meanwhile
> 
> any tester in chat (at random, @Reonu ?! :) )

this was not a merge conflict, you could argue the issue with the simple level layout operator was kinda of a conflict but the other change was just a bug

Reonu:
I can test tonight

jesusyoshi54:
ok honestly I never used custom exports before. I don't really remember what it used to be but shouldn't this string property have a directory subtype so it isn't ass to use. In order for it to export you have to put in the full path you're using, which is kind of weird. You would expect it to branch off of the decomp path.
Also when I did the PR for fixing the name, I used `custom_export_path` not `custom_export_name`. `custom_export_path` is used for the actors not levels -_-
pretty cringe of me

jesusyoshi54:
or rather it seems like the two props are crossed all about. That should get fixed..

jesusyoshi54:
I have some ideas I made a PR for this instead of doing it via review

Dragorn421 (approved):
It is my understanding this has been well tested by the community
The code changes look sensible
I merely tested fast64 enables correctly

PR #407, opened 2024-08-07
[F3D] Fix binary exports

Lilaa3:
If any flag was 0 this would fail

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -4237,7 +4237,7 @@ def geoFlagListToWord(flagList, f3d):
             word += getattr(f3d, name)
         else:
             try:  # Try to cast name to an int instead, if this fails raise an explicit error
-                word += int(word, 0)
+                word += int(name, 0)
             except ValueError as e:
                 raise PluginError("Invalid geometry mode flag " + name) from e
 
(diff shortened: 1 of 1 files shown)


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -4236,7 +4236,10 @@ def geoFlagListToWord(flagList, f3d):
         if name in f3d.allGeomModeFlags:
             word += getattr(f3d, name)
         else:
-            raise PluginError("Invalid geometry mode flag " + name)
+            try:  # Try to cast name to an int instead, if this fails raise an explicit error
+                word += int(word, 0)
+            except:
+                raise PluginError("Invalid geometry mode flag " + name)
```
Dragorn421, line 4242:
```suggestion
            except ValueError as e:
                raise PluginError("Invalid geometry mode flag " + name) from e
```
Catch-all except is bad practice, and chaining exceptions like this gives more helpful context in the trace

Lilaa3: Fair, in this case the only exception that needs to be catched is value error, my bad

Lilaa3:
[minimal example.blend.zip](https://github.com/user-attachments/files/16521900/minimal.example.blend.zip)
Since the user that reported this seems to be offline for now I made a minimal example, use the sm64 insertable binary exporter with ex2 as the microcode and a material with no geo mode flags enabled (there are other ways to trigger this but for the sake of simplicty this works), my file has this set up as is
I already tested and it fixed the issue!

Dragorn421 (approved):
fixes what it ought to fix

PR #408, opened 2024-08-08
[OoT] Fix #404 - Link eyes/mouths import

### LINKED ISSUE #404: [OoT] Link import fails to import eye and mouth textures as flipbook
So the eye/mouth textures are missing and Link has a rgb(0,0,0) face

After investigating this is because https://github.com/zeldaret/oot/pull/1928 changed e.g. `void* sEyeTextures[][8]` to `void* sEyeTextures[][PLAYER_EYES_MAX]` (`8` -> `PLAYER_EYES_MAX`).
But the regex in fast64 for finding these arrays (here https://github.com/Fast-64/fast64/blob/2b44b24d40fd93299ba9697fb5a20b508b7aac47/fast64_internal/oot/oot_texture_array.py#L147 ) assumes a decimal or hexadecimal length, so it "didn't see" those arrays anymore

changing the regex to also allow basically a C identifier like `[0-9a-zA-Z_]` fixes this
full fixed regex `void\s*\*\s*([0-9a-zA-Z\_]*)\s*\[\s*\]\s*\[[0-9a-zA-Z_]*\]\s*=\s*\{(.*?)\}\s*;`
https://regex101.com/r/y5timv/1

<details>

<summary>linkW</summary>

image

</details>

Dragorn421:
Fixes #404

### DIFF
--- a/fast64_internal/oot/oot_texture_array.py
+++ b/fast64_internal/oot/oot_texture_array.py
@@ -144,7 +144,7 @@ def getTextureArrays(actorData: str, flipbookArrayIndex2D: int) -> dict[str, Tex
 
     if flipbookArrayIndex2D is not None:
         for texArray2DMatch in re.finditer(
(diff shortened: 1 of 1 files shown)


Yanis42 (approved):
what a simple fix lol

PR #410, opened 2024-08-08
[OoT] Fix #384 - Custom path DL imports

### LINKED ISSUE #384: [OoT] Importing a DL with the custom path option doesn't work
I remember we talked about this a while ago on discord and it was brought to my attention again recently. The issue occurs when you try to import something with the DL Exporter with the "Use Custom Path" checkbox enabled (and the corresponding value set)

The error is the following: ``FileNotFoundError: [Errno 2] No such file or directory: '\\\\wsl.localhost\\Ubuntu-22.04\\home\\github\\oot\\assets\\objects\\gameplay_keep\\gameplay_keep.c\\gameplay_keepVtx_00D418.vtx.inc'`` (with ``gBoulderFragmentsDL``), basically it keeps the C file part, I remember looking into this but I couldn't find the reason and I just gave up eventually

Dragorn421:
Fixes #384 (loosely, at least it doesn't error and produces a dlist)

See the issue for details

### DIFF
--- a/fast64_internal/oot/f3d/operators.py
+++ b/fast64_internal/oot/f3d/operators.py
@@ -111,7 +111,7 @@ class OOT_ImportDL(Operator):
             folderName = settings.folder
             importPath = abspath(settings.customPath)
             isCustomImport = settings.isCustom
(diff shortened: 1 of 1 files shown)


Yanis42 (approved):
nice

PR #413, opened 2024-08-09
[SM64] Only throw empty gfx exception in actor exports

Lilaa3:
Oh lila LOVESSS breaking main RIGHT SCUT?


fast64_internal/sm64/sm64_geolayout_writer.py
```diff
@@ -978,7 +977,7 @@ def exportGeolayoutObjectBinary(
     camera,
 ):
     geolayoutGraph, fModel = convertObjectToGeolayout(
-        obj, convertTransformMatrix, camera, obj.name, None, None, DLFormat.Static, True
+        obj, convertTransformMatrix, True, obj.name, None, None, DLFormat.Static, True
```
jesusyoshi54, line 980: You should remove the "camera" argument from the functions if it is not used.

jesusyoshi54 (changes requested):
I tested this and it works and the code is simple enough.

On an somewhat unrelated note, the `exportGeolayoutArmature` have camera arguments with some esoteric uses that are lost to time and never used. This should be cleaned up at some point.

fast64_internal/sm64/sm64_geolayout_writer.py
Lilaa3: that is what I did?

jesusyoshi54: yeah remove the `camera` from the function `exportGeolayoutObjectBinary` and others that do not actually use the argument.

jesusyoshi54:
When you call `exportGeolayoutObjectBinaryBank0` it still has the `camera` argument present. Same with `exportGeolayoutObjectBinary` and `exportGeolayoutObjectC`
```python
  File "C:\Users\scut\Desktop\fast64-main\addons\fast64\fast64_internal\sm64\sm64_objects.py", line 1854, in execute
    self.execute_gfx(props, context, obj, index)
  File "C:\Users\scut\Desktop\fast64-main\addons\fast64\fast64_internal\sm64\sm64_objects.py", line 1843, in execute_gfx
    raise Exception(e)
Exception: Error: exportGeolayoutObjectC() takes 13 positional arguments but 14 were given
```

Lilaa3:
sorry about that I will pay more attention next time

jesusyoshi54 (approved):
exported object and area, plus extra empty area. compiles and boots

Lilaa3:
@jesusyoshi54 now that i think about it, would a separate pr for collision be fine?

jesusyoshi54:
what would that PR be for?

Lilaa3:
@jesusyoshi54 basically the same thing i guess

jesusyoshi54:
I don't see a need for it tbh but I guess you can make it. Helpful for menu
type levels maybe. I would make it a new PR though if you want to do it.

PR #414, opened 2024-08-10
[F3D Bleed] Fix #412 - Compare DPSetTileSize tags to last cmd's

### LINKED ISSUE #412: [F3D] Mat bleed issues with tile scrolling
In a level model with multiple materials using tile scrolling and multiple materials not using scrolling, exporting with mat bleed caused some mats that shouldnt scroll to scroll, and some mats that should scroll to not

Lilaa3:
This fixes materials with no tags following a material with a tag being bleeded even though they should not.
Fixes #412


fast64_internal/f3d/f3d_bleed.py
```diff
@@ -438,6 +437,12 @@ def bleed_individual_cmd(self, cmd_list: GfxList, cmd: GbiMacro, bleed_state: in
         if not last_cmd_list:
             return self.bleed_self_conflict
 
+        # Don´t bleed if the cmd tags are not the same (these are not hashed)
+        if cmd in last_cmd_list:
```
jesusyoshi54, line 441: I think it would be better if this behavior was done inside the cmd specific bleed logic. There is no reason it couldn't be done there since really this only affects set tile size cmds.
In the case that tags have some general use to every type of gfx cmd it would make sense to have it here, but if it is for set tile size put it in the set tile size function

Lilaa3: I was thinking of future proofing I guess, but if that happens it would be better to just hash the tags

fast64_internal/f3d/f3d_bleed.py
```diff
@@ -479,9 +478,13 @@ def bleed_SPSetOtherMode(self, cmd_list: GfxList, cmd: GbiMacro, bleed_state: in
             else:
                 return cmd == self.default_othermode_L
 
-    # bleed if there are no tags to scroll and cmd was in last list
+    # Don´t bleed if the cmd tags are not the same (these are not hashed)
     def bleed_DPSetTileSize(self, cmd_list: GfxList, cmd: GbiMacro, bleed_state: int, last_cmd_list: GfxList = None):
-        return cmd.tags != GfxTag.TileScroll0 and cmd.tags != GfxTag.TileScroll1 and cmd in last_cmd_list
+        if cmd in last_cmd_list:
+            last_size_cmd = last_cmd_list[last_cmd_list.index(cmd)]
+            if last_size_cmd.tags == cmd.tags:
```
jesusyoshi54, line 485: If you have two materials with scrolls and the same cmd values but different scroll values they will bleed when they shouldn't. A scroll should never bleed.
This does prevent a cmd from bleeding after a scroll which is good and I have tested that that works. Just change this to something like, `cmd.tags != GfxTag.TileScroll0 and cmd.tags != GfxTag.TileScroll1 and cmd in last_cmd_list` and then do the same conditional you do here

jesusyoshi54 (changes requested):

jesusyoshi54 (approved):
works :)

PR #415, opened 2024-08-11
[SM64] Check actual bone type instead of group or collections

Lilaa3:


jesusyoshi54 (approved):
works on mario and makes sense code wise.

PR #417, opened 2024-08-12
Fix gMtxClear reference for OOT scene importing

kurethedead:
When importing a scene that references `gMtxClear` (ex. inside Jabu Jabu), currently the code fails since `F3DContext.createMesh()` clears out old matrix data. This fixes that.


Dragorn421 (approved):
can import jabu (up to the point where it errors because of the transition actor room id thing)
does not break importing other things
:heavy_check_mark:

PR #419, opened 2024-08-13
[F3D] Fix issues in the GBI macros

Lilaa3:
Add size() of 2 for SPAmbOcclusion and SPLightColor, fixes SPPerspNormalize in ex3, DPSetOtherMode and DPSetTileSize.is_LOADTILE (was that one ever used?).
Also cleans up SPSetOtherMode to use getattr default instead of checking with hasattr.

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -4826,9 +4826,9 @@ class DPSetOtherMode(GbiMacro):
     def to_binary(self, f3d, segments):
         mode0 = mode1 = 0
         for mode in self.mode0:
-            mode0 |= getattr(f3d, mode) if hasattr(f3d, str(mode)) else mode
+            mode0 |= getattr(f3d, str(mode), mode)
         for mode in self.mode1:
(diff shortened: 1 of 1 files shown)


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -4818,7 +4824,12 @@ class DPSetOtherMode(GbiMacro):
     mode1: list
 
     def to_binary(self, f3d, segments):
-        words = _SHIFTL(f3d.G_RDPSETOTHERMODE, 24, 8) | _SHIFTL(self.mode0, 0, 24), self.mode1
+        mode0 = mode1 = 0
+        for mode in self.mode0:
+            mode0 |= getattr(f3d, mode) if hasattr(f3d, str(mode)) else mode
```
Dragorn421, line 4829: why is getattr given `mode` and `hasattr` is given `str(mode)` ?

`getattr` can also take a default if it simplifies things:
`getattr(f3d, mode, mode)`

Lilaa3: because I referenced SPSetOtherMode, should I update it too?

Dragorn421 (approved):

PR #423, opened 2024-08-15
[SM64] Fix SM64 binary geolayout exports

Lilaa3:
Don't pass in dl reference when not an object empty (need confirmation this is fine)
Update size and export binary functions for transform nodes
Check for export tag in material binary saves (fixes bleed exporting unused data)


Lilaa3:
Last commit accidentaly fixed an issue with bleed in binary!

Lilaa3:
It turns out my last issue was just from world defaults being wonky in the file I was testing, everything is fixed sorry for the wait

ThePerfectMario64:
LET'S FUCKING GO

jesusyoshi54 (approved):
exported successfully Mario in RM test ROM.
worked with and without bleed.
also exported to decomp and it still worked
image

Issue #424, opened 2024-08-19
[SM64] Area 2, 3 etc are messed up ingame.


FrijolesYQueso:
Using the lastest commits, Areas anything other than 1 are extremely stretched all scales, transforms and the like were applied. Collision is not affected.
![image-1370](https://github.com/user-attachments/assets/5faa1fbd-461d-4910-a64e-ca9cc572c968)

Lilaa3:
broken by https://github.com/Fast-64/fast64/pull/352

Dragorn421:
Partially fixed by #425
No test case to debug any issue left, closing

PR #425, opened 2024-08-19
[SM64] Fix #352 regression

Lilaa3:
It is not enough to rely on an object's location and rotation as we need a relative position.
[broken graphics parented.blend.zip](https://github.com/user-attachments/files/16657743/broken.graphics.parented.blend.zip)
[Code diff](https://github.com/Lilaa3/sm64/commit/321dedfe3d4d114e0f32cf28343e67f9b76d4069)

### DIFF
--- a/fast64_internal/utility.py
+++ b/fast64_internal/utility.py
@@ -788,7 +788,8 @@ def store_original_mtx():
     for obj in yield_children(active_obj):
         # negative scales produce a rotation, we need to remove that since
         # scales will be applied to the transform for each object
-        obj["original_mtx"] = Matrix.LocRotScale(obj.location, obj.rotation_euler, None)
+        loc, rot, _scale = obj.matrix_local.decompose()
+        obj["original_mtx"] = Matrix.LocRotScale(loc, rot, None)
 
 
 def rotate_bounds(bounds, mtx: mathutils.Matrix):


Reonu (approved):
Exported my RHDC.ppt level which is a custom level added with fast64, works fine. tested the things mentioned in the PR and saw no problems

jesusyoshi54 (approved):
ahh yes, the classic gaffe.
confirmed bug fix. First before
image
Then after
image
on another note, the instant warp with object offsets should make use of the same thing. They should account for relative locations so that you can parent each object to an area.

Lilaa3:
I don´t have confirmation that this fixed #424 just yet, @FrijolesYQueso I know this fixed at least one issue but what else still happens?

FrijolesYQueso:
I ended up deleting the mesh and started with a new one. Seemingly worked again.

Lilaa3:
> I ended up deleting the mesh and started with a new one. Seemingly worked again.

Well that's not really helpful to figuring out what was wrong, if you still have your old file send it over

FrijolesYQueso:
Probably should have worded that better, but I mean't that it did fix the offset issue. However the old mesh still had a few broken polygons, and I had added in a plane as well to test said issue, and the PR you gave me made that appear properly. Unfortunately I overwrote the blend long since then

Lilaa3:
Didn't fix the weird behavior in the issue, but did fix one bug

PR #426, opened 2024-08-19
[SM64] Custom draw layers

Lilaa3:
Gives the user full control over their available draw layers, allowing for custom draw layers!


Anoliah:
Not sure what the plans are for this one, but right now it's giving the error `8 is not a valid draw layer` when trying to export something using a newly added layer. Would be nice if this got finished eventually.

PR #427, opened 2024-08-20
[F3D] Don´t update format to reload node values

Lilaa3:
Fixes the auto picker trying to run in material upgrades


Dragorn421 (approved):
sure

PR #428, opened 2024-08-20
[OoT] Fix scene custom path import

Dragorn421:
also fixes my faulty merge from #366 which added back in deleted file oot_level_parser.py


Reonu (approved):
it worke

Dragorn421:
I'll skip the code side review as it's a half-line change

PR #429, opened 2024-08-20
Speed up render settings update and add toggle for doing it manually

Dragorn421:
This is also designed in a way that allows e.g. f64render to do `bpy.context.scene.fast64.renderSettings.enableAutoUpdatePreview = False` to prevent (most) callbacks from doing anything


krm01 (approved):
been testing this for several days with no issues, only bug was something unrelated that's already been fixed in main.

Lilaa3 (approved):
Code looks good and it works, currently turning off auto requires switching to eevee first, but if htd plans on just setting the property in the f64 renderer then I see no problem in that.

Dragorn421:
> if htd plans on just setting the property in the f64 renderer

That's what I planned for yeah

PR #431, opened 2024-08-22
[SM64] Fix #355 - Fix unaddressed object types in #358

### LINKED ISSUE #355: (Blender 4.0.1+) 'Euler' object has no attribute 'to_euler' Error when exporting a Level

Lilaa3:
The previous fix was incomplete but the issue was never noticed from how uncommon camera volumes are. 
Previous fix was a fix to a #339 regression.

### DIFF
--- a/fast64_internal/sm64/sm64_objects.py
+++ b/fast64_internal/sm64/sm64_objects.py
@@ -854,7 +854,7 @@ def process_sm64_objects(obj, area, rootMatrix, transformMatrix, specialsOnly):
                     SM64_Whirpool(obj.whirlpool_index, obj.whirpool_condition, obj.whirpool_strength, translation)
                 )
             elif obj.sm64_obj_type == "Camera Volume":
-                checkIdentityRotation(obj, rotation, True)
+                checkIdentityRotation(obj, rotation.to_quaternion(), True)
                 if obj.cameraVolumeGlobal:
                     triggerIndex = -1
                 else:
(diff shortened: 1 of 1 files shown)


Dragorn421 (approved):
code's obviously ok, someone test?

jesusyoshi54 (commented):
Camera volumes gave error object has no attribute to euler before, now exports without error.
puppycam volumes do not work though, but that is a separate issue. You can approve if you want or you can also fix that. Here is the error
```python
File "C:\Users\\Desktop\fast64-main\addons\fast64\fast64_internal\sm64\sm64_objects.py", line 936, in process_sm64_objects
    levelIDNames[bpy.data.scenes["Scene"].levelOption]
```

Lilaa3:
Sounds like something for combined exporter regressions pr to me, review that instead I will add it rn

jesusyoshi54:
Well it isn't really really related to combined export but sure do it there

Lilaa3:
> Well it isn't really really related to combined export but sure do it there

It was broken by combined exporter pr, levelOption was removed in it

jesusyoshi54:
Oh I was talking about the fact that it just assumes the scene name is "scene"

Lilaa3:
> Oh I was talking about the fact that it just assumes the scene name is "scene"
Oh I'm very sorry, is it out of scope to fix it in regressions pr 😅

PR #432, opened 2024-08-22
Replace hackerFeaturesEnabled with featureSet 

Pepe20129:


fast64_internal/oot/__init__.py
```diff
@@ -58,6 +58,13 @@
     oot_operator_unregister,
 )
 
+
+feature_set_enum = (
+    ("decomp", "Decomp", "Decomp"),
+    ("hacker_oot", "HackerOOT", "Hacker OOT"),
```
Dragorn421, line 64: the correct spelling is exactly "HackerOoT" https://github.com/HackerN64/HackerOoT

fast64_internal/oot/cutscene/exporter/functions.py
```diff
@@ -44,6 +44,6 @@ def getNewCutsceneExport(csName: str, motionOnly: bool):
     # this allows us to change the exporter's variables to get what we need
     return CutsceneExport(
         getCutsceneObjects(csName),
-        bpy.context.scene.fast64.oot.hackerFeaturesEnabled or bpy.context.scene.fast64.oot.useDecompFeatures,
+        (bpy.context.scene.fast64.oot.feature_set == "hacker_oot") or bpy.context.scene.fast64.oot.useDecompFeatures,
```
Dragorn421, line 47:
```suggestion
        (bpy.context.scene.fast64.oot.feature_set == "hacker_oot") or (bpy.context.scene.fast64.oot.feature_set == "decomp" and bpy.context.scene.fast64.oot.useDecompFeatures),
```

Dragorn421 (commented):
code looks good, I'll test later

fast64_internal/oot/__init__.py
Pepe20129: Done.

fast64_internal/oot/__init__.py
```diff
@@ -58,6 +58,13 @@
     oot_operator_unregister,
 )
 
+
+feature_set_enum = (
+    ("decomp", "Decomp", "Decomp"),
+    ("hacker_oot", "HackerOoT", "Hacker OoT"),
```
Yanis002, line 64: I'm not sure about these names, how about "OoT" instead of "Decomp" instead? technically both repos are decomps so it bothers me to name one "decomp" when they both are, 

now that I think more about it, this is more concerning if the plan is add SoH support since this is also a decomp of OoT, so it should be like "Default", "HackerOoT" and "Ship" (or "SoH" or whatever name you like)

also ``hacker_oot`` should be changed to ``hackeroot`` imo but it's probably fine if this is to help people reading this

fast64_internal/oot/file_settings.py
```diff
@@ -24,9 +24,9 @@ def draw(self, context):
             prop_split(col, context.scene.fast64.oot, "oot_version_custom", "Custom Version")
 
         col.prop(context.scene.fast64.oot, "headerTabAffectsVisibility")
-        col.prop(context.scene.fast64.oot, "hackerFeaturesEnabled")
+        col.prop(context.scene.fast64.oot, "feature_set")
 
-        if not context.scene.fast64.oot.hackerFeaturesEnabled:
+        if bpy.context.scene.fast64.oot.feature_set == "decomp":
```
Yanis002, line 29:
```suggestion
        if context.scene.fast64.oot.feature_set == "decomp":
```
no need to use bpy there (it's generating an error actually, and we don't need to import bpy since we already have access to context)

fast64_internal/oot/scene/panels.py
```diff
@@ -43,7 +43,7 @@ def draw(self, context):
             self.drawSceneSearchOp(exportBox, settings.option, "Export")
         settings.draw_props(exportBox)
 
-        if context.scene.fast64.oot.hackerFeaturesEnabled:
+        if bpy.context.scene.fast64.oot.feature_set == "hacker_oot":
```
Yanis002, line 46: same thing about bpy there, it should be removed

fast64_internal/oot/scene/operators.py
```diff
@@ -155,7 +155,7 @@ def execute(self, context):
             option = settings.option
 
             bootOptions = context.scene.fast64.oot.bootupSceneOptions
-            hackerFeaturesEnabled = context.scene.fast64.oot.hackerFeaturesEnabled
+            hackerFeaturesEnabled = bpy.context.scene.fast64.oot.feature_set == "hacker_oot"
```
Yanis002, line 158: useless bpy thing

Yanis002 (changes requested):

fast64_internal/oot/__init__.py
Pepe20129: Done.

fast64_internal/oot/scene/panels.py
Pepe20129: Done.

fast64_internal/oot/scene/operators.py
Pepe20129: Done.

fast64_internal/oot/__init__.py
```diff
@@ -77,7 +84,7 @@ class OOT_Properties(bpy.types.PropertyGroup):
     """Global OOT Scene Properties found under scene.fast64.oot"""
 
     version: bpy.props.IntProperty(name="OOT_Properties Version", default=0)
-    hackerFeaturesEnabled: bpy.props.BoolProperty(name="Enable HackerOOT Features")
+    feature_set: bpy.props.EnumProperty(name="Feature Set", default="decomp", items=feature_set_enum)
```
Yanis002, line 87:
```suggestion
    feature_set: bpy.props.EnumProperty(name="Feature Set", default="default", items=feature_set_enum)
```

Yanis002 (commented):
just this and it should be good

Yanis002:
actually there's other places where it checks for "decomp" instead of "default" (changing enums can be annoying)

PR #433, opened 2024-08-23
[UI] [OOT] More informative text regarding culling

Reonu:
Changes some text labels that might be misleading. 

This one only shows up when you select the Cullable room type which leads you to believe it sets up CPU culling automatically. It doesn't. The only thing fast64 sets up automatically is RSP culling, and it does so regardless of whether the room type is "Cullable" or "Normal".
imagen

Therefore I changed it to: 
imagen

This text which shows up when you select  "Custom Cull Group" empty type is also misleading. It can lead the user to believe that what they're doing is unnecessary since fast64 is already doing it:
imagen

So I changed it to:
imagen

### DIFF
--- a/fast64_internal/oot/props_panel_main.py
+++ b/fast64_internal/oot/props_panel_main.py
@@ -229,7 +229,7 @@ class OOTCullGroupProperty(bpy.types.PropertyGroup):
         col.prop(self, "sizeControlsCull")
         if not self.sizeControlsCull:
             prop_split(col, self, "manualRadius", "Radius (OOT Units)")
-        col.label(text="RSP culling is automatic. This is for CPU culling.", icon="INFO")
+        col.label(text="RSP culling is automatic. The 'Custom Cull Group' empty type is for CPU culling.", icon="INFO")
         col.label(text="This will create custom cull group shape entries to be used in Cullable rooms.")
         col.label(text="Use Options -> Transform -> Affect Only -> Parent ", icon="INFO")
         col.label(text="to move object without affecting children.")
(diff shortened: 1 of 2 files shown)


Lilaa3:
You already know I don´t do oot but you need to black format first, reonu.
See the readme

Dragorn421:
I offered to do it for him and I will

fast64_internal/oot/room/properties.py
```diff
@@ -172,7 +172,8 @@ def draw_props(self, layout: UILayout, dropdownLabel: str, headerIndex: int, obj
                 if self.roomShape == "ROOM_SHAPE_TYPE_IMAGE":
                     self.drawBGImageList(general, objName)
                 if self.roomShape == "ROOM_SHAPE_TYPE_CULLABLE":
-                    general.label(text="Cull regions are generated automatically.", icon="INFO")
+                    general.label(text="This is for CPU culling, and requires meshes to be parented to Custom Cull Group empties.", icon="INFO")
```
Dragorn421, line 175:
```suggestion
                    general.label(text="The 'Cullable' room shape type is for CPU culling, and requires meshes to be parented to Custom Cull Group empties.", icon="INFO")
```
otherwise it's unclear what "this" refers to imo

fast64_internal/oot/props_panel_main.py
```diff
@@ -229,8 +229,8 @@ def draw_props(self, layout: bpy.types.UILayout):
         col.prop(self, "sizeControlsCull")
         if not self.sizeControlsCull:
             prop_split(col, self, "manualRadius", "Radius (OOT Units)")
-        col.label(text="Meshes generate cull groups automatically.", icon="INFO")
-        col.label(text="This is only for custom cull group shapes.")
+        col.label(text="RSP culling is automatic. This is for CPU culling.", icon="INFO")
```
Dragorn421, line 232:
```suggestion
        col.label(text="RSP culling is automatic. The 'Custom Cull Group' empty type is for CPU culling.", icon="INFO")
```

Dragorn421:
image
image

Dragorn421 (approved):

PR #438, opened 2024-08-25
[SM64] Update included mario models to v6

Lilaa3:
Replace lowpolyskinnedmario by its v5 variant (upgraded to v6), there shouldn´t be two.
I think it's also worth considering removing the low poly blend in place of yoshi milkman's skinned mario, or remove the mario models all together to be safe (would ideally be done post c geolayout importing).


Dragorn421:
How about we remove all the blends and create a new repo for them

Lilaa3:
> How about we remove all the blends and create a new repo for them

Repo where exactly? I think this just has to be removed all together once importing is really easy

Dragorn421:
Fast-64/fast64-models, and yeah we can remove them from there later. It'll still be in the git history of course but oh well so is it here

Lilaa3:
> Fast-64/fast64-models, and yeah we can remove them from there later. It'll still be in the git history of course but oh well so is it here

If you aren´t busy you could create that and I'll pr (I think I can do it too now?), but it's still not safe aha

PR #440, opened 2024-08-30
[F3D/MK64] Fix #366 regression

Lilaa3:
Remove game requirement for importMesh, mk64 uses it.


coco875:
try in association with #441 and it work

Dragorn421 (approved):
nice fix, I didn't know how that should be addressed

PR #445, opened 2024-09-04
[SM64] Fix zoomout masks

Lilaa3:
I went over the file again to make sure there are no other old macro appends that don´t used the new named tuple but couldn´t find any other, weird that this never triggered for me even when using the test file provided by @Reonu


Reonu (approved):
me when the PR works:

Dragorn421 (approved):

Reonu:
me when the PR is merged:

PR #447, opened 2024-09-06
more conservative game editor update

Lilaa3:
Noticed that when kaze updated his microcode was set when it shouldn´t have, this is bad but we need to update existing world defaults nonetheless, added repo settings load to make this even less of an issue, pre-release should update the decomp path causing the path to be correct as well (if there is a repo file of course)

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -328,9 +328,10 @@ def upgrade_changed_props():
     SM64_ObjectProperties.upgrade_changed_props()
     OOT_ObjectProperties.upgrade_changed_props()
     for scene in bpy.data.scenes:
-        if scene.fast64.settings.internal_game_update_ver != 1:
-            gameEditorUpdate(scene, bpy.context)
-            scene.fast64.settings.internal_game_update_ver = 1
+        settings: Fast64Settings_Properties = scene.fast64.settings
+        if settings.internal_game_update_ver != 1:
+            set_game_defaults(scene, False)
+            settings.internal_game_update_ver = 1
         if scene.get("decomp_compatible", False):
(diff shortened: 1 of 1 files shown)


Dragorn421:
This should be merged in 2.3.0?

Lilaa3:
It's definitly a bug since it messes up files

__init__.py
```diff
@@ -363,23 +364,34 @@ def after_load(_a, _b):
...
-        self.f3d_type = "F3D"
+    elif scene.gameEditorMode == "MK64":
+        f3d_type = "F3DEX/LX"
+    elif scene.gameEditorMode == "Homebrew":
+        f3d_type = "F3D"
         world_defaults = {}  # This will set some pretty bad defaults, but trust the user
-    if self.world is not None:
-        self.world.rdp_defaults.from_dict(world_defaults)
+    if set_ucode:
+        scene.ucode = f3d_type
```
Dragorn421, line 388:
```suggestion
        scene.f3d_type = f3d_type
```

```
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/__init__.py", line 394, in gameEditorUpdate
    set_game_defaults(scene)
  File "/home/dragorn421/.config/blender/4.2/scripts/addons/fast64_gitlinked/__init__.py", line 388, in set_game_defaults
    scene.ucode = f3d_type
    ^^^^^^^^^^^
AttributeError: 'Scene' object has no attribute 'ucode'
```

Dragorn421 (approved):
This shou

PR #448, opened 2024-09-07
[SM64] fix exporting the levels/scripts.c write for common1 and group0

s4Ys369:
currently the `LOAD_MODEL_FROM_GEO` command gets written incorrectly, before the segment load commands. The game will compile, but crashes.


Lilaa3 (approved):
Works on the machine of someone very close to me

PR #449, opened 2024-09-08
[SM64] Smarter includes

Lilaa3:
I'd really like some opinions on write_includes's complexity and specifically the fact that it's so specific, I think its a good step foward as is tho.

Added new functions, write_includes and update_actor_includes.

write_includes is a smarter version of writeIfNotFound specifically for includes and externs, it removes comments prior to searching (but doesn't remove comments from the actual file of course) and allows whitespace (where possible, a path including whitespace won´t be found when searching for a file with no whitespace), futhermore it adds a trailing newline to the file instead of adding a leading newline for each include, the logic that finds the `#endif` of .h files is also improved like the logic to find includes and externs (whitespace is tolerated, comments are ignored) but it also allows no endif for files that use the GCC `#pragma once` macro

update_actor_includes standardizes group and level includes into one place for ease of editing, one of the major differences is that level exports no longer use the full path (levels/actor/my_file.c would instead be /actor/my_file.c) this should allow level header type actors to actually update headers on custom non decomp levels but that is not done in this PR.

### DIFF
--- a/fast64_internal/sm64/sm64_utility.py
+++ b/fast64_internal/sm64/sm64_utility.py
@@ -1,8 +1,19 @@
+from pathlib import Path
 import os
+import re
+
 import bpy
 from bpy.types import UILayout
 
-from ..utility import PluginError, filepath_checks, run_and_draw_errors, multilineLabel, prop_split
+from ..utility import (
+    PluginError,
+    filepath_checks,
+    removeComments,
+    run_and_draw_errors,
+    multilineLabel,
+    prop_split,
+    COMMENT_PATTERN,
+)
 from .sm64_function_map import func_map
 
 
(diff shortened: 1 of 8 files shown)


Lilaa3:
While I like my current idea, I think I can iron out some stuff and actually update writeIfNotFound, looking into that right now

Lilaa3:
I'm actually done this time, just needed to change stuff for animations rework

jesusyoshi54:
can you add function descriptions to the more complex funcs, it is something we should probably do more often

fast64_internal/sm64/sm64_utility.py
```diff
@@ -122,3 +136,305 @@ def convert_addr_to_func(addr: str):
...
+    to_remove: Optional[list[ModifyFoundDescriptor]] = None,
+    path_must_exist=False,
+    create_new=False,
+    error_if_no_header=False,
+    header: Optional[ModifyFoundDescriptor] = None,
+    error_if_no_footer=False,
+    footer: Optional[ModifyFoundDescriptor] = None,
+    ignore_comments=True,
+):
+    """
```
jesusyoshi54, line 281: You should write what the header and footer are meant to be since that is sort of the important part of this

jesusyoshi54 (approved):

PR #450, opened 2024-09-09
[SM64] Fix/Improve multi-actor UI (not export all selected)

Lilaa3:
Doesn´t add empties to the text, uses parenthesis, show similar text for level and custom


fast64_internal/sm64/sm64_objects.py
```diff
@@ -2166,12 +2179,24 @@ def draw_level_path(self, layout):
         return True
 
     def draw_actor_path(self, layout):
-        actor_path = Path(bpy.context.scene.fast64.sm64.decomp_path) / "actors"
-        if not filepath_ui_warnings(layout, (actor_path / self.actor_group_name).with_suffix(".c")):
+        if self.export_locations is None:
             return
```
Dragorn421, line 2183:
```suggestion
            return False
```
and other returns in this function
would be cleaner

fast64_internal/sm64/sm64_objects.py
```diff
@@ -2156,8 +2156,21 @@ def draw_export_options(self, layout):
...
 
+    @property
+    def actor_names(self) -> list:
+        return list(dict.fromkeys(filter(None, [self.obj_name_col, self.obj_name_gfx])).keys())
+
+    @property
+    def export_locations(self) -> str | None:
+        names = self.actor_names
+        if len(names) > 1:
+            return f"({'/'.join(names)})"
```
Dragorn421, line 2167: Note `{a,b,c}` is a more common syntax for indicating "a or b or c", as it's shell syntax. For example
```c
$ echo foo/{a,b,c}
foo/a foo/b foo/c
```

jesusyoshi54 (approved):
Good. Definitely more clear that a manually selected gfx and col object will go to different places.

PR #452, opened 2024-09-10
[SM64/F3D] Fix and update Ui Image Exporter

### LINKED ISSUE #451: Exporting Hud results in Error
the Error is  "TextInfo.writeall() takes Four Arguments, but Five were given."  Screenshot:
Screen Shot 2024-09-09 at 7 24 39 PM

Lilaa3:
Fixes #451 

- Updated: Tex prop is now drawn using ui_image, ui_image has been updated to optionally hide low and high props (are handled by size cmd, does not exist for texture rectangles) 
- materialless_setup func for TexInfo, equivelant to moreSetupFromModel for bpy material-less contexts, obviously cannot handle flipbooks 
- setup_single_tex func for TexInfo, sets up a single texture without the context of the multi texture manager
- ignore_tex_set bool for fromProp
- Use only fromProp and updated writeAll
- Use saveModeSetting (and world defaults) instead of manullay appending each DP command (except for blend color)
- Removed syncs except for the sync after the texrect cmd to do the reverts
- Removed deprecated arg for png exporting in save_textures (assumed from fimage)
- Removed the ) part of the delimiter checked when exporting "Menu", refresh 13 made it `(void)` instead of `()`
- Some basic ui reordering to match other panels
- Added 1 cycle variant for non rgba16 exports
- Fix dynamic displaylists adding an extra comma (caused a compiler error)
- Fix overwritedata not handeling the aligner and made it so palletes are handeled correctly as well

### DIFF
--- a/fast64_internal/f3d/f3d_gbi.py
+++ b/fast64_internal/f3d/f3d_gbi.py
@@ -3262,6 +3262,10 @@ class FImage:
     isLargeTexture: bool = field(init=False, compare=False, default=False)
     converted: bool = field(init=False, compare=False, default=False)
 
+    @property
+    def aligner_name(self):
+        return f"{self.name}_aligner"
+
     def size(self):
         return len(self.data)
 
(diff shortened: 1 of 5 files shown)


sauraen (approved):
Non SM64 stuff looks okay. Please make sure to test export with some advanced cases like materials with two CI textures before merging.

Lilaa3:
> Non SM64 stuff looks okay. Please make sure to test export with some advanced cases like materials with two CI textures before merging.

Actual non-sm64 logic is unaffected besides one thing i changed now because i didnt notice the dynamic commands were broken aha (ive just been making sure displaylists were correct but today i made sure to test everything since I didn´t hear back from the person having the issues in the first place, yes i should've been more careful but I was expecting to hear back.. sorry)

Lilaa3:
Is there anyone else interested in testing this real quick? @Reonu ?

jesusyoshi54:
Worked during my testing with complicated materials

PR #455, opened 2024-09-11
Create black-suggest.yml

coco875:
will suggest what black found wrong automatically.

### DIFF
--- a/.github/workflows/black-suggest.yml
+++ b/.github/workflows/black-suggest.yml
@@ -1,0 +1,16 @@
+name: black Suggest
+on: [pull_request]
+jobs:
+  formatting-check:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Dependance
+        run: |
+          pip install black
+      - name: Run black style check.
+        run: |
+          black .
+      - uses: parkerbxyz/suggest-changes@v1
+        with:
+          comment: 'Please commit the suggested changes from black.'


coco875:
I will make an example of PR later, but also don't know if it's useful because black are pretty easy to install and are easy to format with it

Lilaa3:
Not an expert on github actions but shouldn't you be using our specific version of black? the check is failing

coco875:
The fail is because of I don't setup the install correctly and will try to do how your action do

coco875:
Ok can't because he use the source directly but I will hard code the version That you use

Lilaa3:
@Dragorn421 I think this could be neat, didn't you have to black format for reonu just a while ago?

coco875:
(I can also make an action who will PR black format when he detect it)

Dragorn421:
This is neat
Is there a way to combine the two github actions in one? Both the lint check and suggestions are doing the same thing (black formatting)

coco875:
in one file or in one job ?

coco875:
and black suggest need to be trigger only for pull request not for push so it will be even hard to combine in one file

Lilaa3:
I don´t know how this could be merged into a single action, don´t see anything on the docs. Is it possible to check if there was a difference and use that to block merging? @coco875

Lilaa3:
I tried to help coco implement a single job for both formatting suggestion and linting, see if you like it!

Lilaa3:
@Dragorn421 I'm going to wait a bit longer on your input before approving if that's ok.

PR #457, opened 2024-09-13
[SM64] Custom level script and special commands and improvements to custom geo commands

Lilaa3:
Custom geo (in levels) replaced with custom object type which can export as collision, level or geo, logic is shared across these.
New custom type is more advanced with inheritable translation, rotation, scale, matrix, displaylist, layer, alongside these there are types such as parameter, color, number, enum and boolean. These are all highly customizable, to a degree even I feel is overkill.
The new custom command can export to binary, facilitated by an eval field and value type for the exported primitive types.

Adds depth arg to geolayouts/level/collision commands so bigger custom command are readable, along with removing commas and new lines and adding them in the respective loops for consistency.
Adds do_export_checks and group_children fields for geolayout cmds, these are currently optional but in the future I think other classes should use these methods.

With the data driven, extremely customizable and feature complete nature of these, I wonder if future commands could use this system, similar to oot using xml for actors. I don´t just mean lol make vanilla commands use this but instead a repo like hackersm64 being able to define essential presets that would show up in the usual enum. This would only need a way of enforcing the presets instead of them going back to NO_PRESET but that's easy to implement after the PR. Sweet sweet light commands tm soon


Lilaa3:
Removed hacks, world is better now

fast64_internal/sm64/sm64_objects.py
```diff
@@ -719,11 +726,80 @@ def to_c(self):
...
+            if self.rot is not None:
+                arg_groups.append(f"/*rot ({self.rot_type.lower()})*/ {', '.join([f'{x}f' for x in self.rot])}")
+            if self.scale is not None:
+                arg_groups.append(f"/*scale*/ {', '.join([f'{x}f' for x in self.scale])}")
+        if self.color is not None:
+            arg_groups.append(f"/*color*/ {', '.join([str(x) for x in self.color])}")
+        if self.parameter is not None:
+            arg_groups.append(f"/*param*/ {self.parameter}")
+        if len(str(arg_groups)) > 100:
+            seperator = ",\n" + ("\t" * (depth + 1))
```
jesusyoshi54, line 796: it makes no sense to add a `_depth` arg to all the associated macro functions and then only use it in this one specific instance. There likely is a better way of determining the proper amount of tabs to be used.

fast64_internal/sm64/sm64_geolayout_classes.py
```diff
@@ -380,9 +380,9 @@ def to_binary(self, segmentData):
 
     def to_c(self, depth):
         if self.node is not None:
-            nodeC = self.node.to_c()
+            nodeC = self.node.to_c(depth)
```
jesusyoshi54, line 383: why do this, it seems to me that the `depth` argument is never used by any "node" type object

fast64_internal/sm64/sm64_objects.py
Lilaa3: Open to suggestions, I'm not checking  the function signature that's for sure

jesusyoshi54: is there ever a case in which the depth is different than 1 and could not be adjusted for in the calling function?
you can return unformatted f strings, e.g. a "{tabs} macro(args)" then run `str.format` on the returned value so that it inserts the tabs later, allowing you to defer indentation until the root caller.
Imo this is better than carrying args to full depth

fast64_internal/sm64/sm64_geolayout_classes.py
Lilaa3: just for consistency

fast64_internal/sm64/sm64_objects.py
Lilaa3: I think formatting every string for depth is much more hacky than passing in depth, are you sure about this?

jesusyoshi54:
It would be good to get some rounding here. The arguments also don't seem to match between export (I guess due to some parenting memes) and GUI.
Tbh I don't see much of a reason to go beyond a few decimal points of precision, but I guess if you want to keep a lot then 6 would be good. I think it would be best to just avoid the negative exponentials.
image

jesusyoshi54:
Custom commands that are direct children of the level root and not the area root should be written as level commands, but currently they just don't export. For example the setup I have as pictured will result in nothing exported, when the user would probably expect the custom command to execute before the AREA cmds do
image

jesusyoshi54:
The `remove or add custom preset cmd` button seems to only ever add custom presets. There is no place to edit the property of the operator to change the case. I mean it has been a while for me but afaik this is something you can only do for operators that bring up a menu when you execute them or via a redo
```python
    index: IntProperty(default=-1)
    op_name: StringProperty()
    example_name: StringProperty(default="")

    def execute_operator(self, context):
        presets = context.scene.fast64.sm64.custom_cmds
        custom, owner = get_custom_prop(context)
        conf_type = get_conf_type(context)
        match self.op_name:
```
So in the code here there is no way to have any case other than the default and no name other than the default

jesusyoshi54:
I'll also once again request some class and function hints, this is a lot of code. This will be helpful for the future of the project because there is no way in heck any of this code is being re-used by anyone besides you at this rate, and we don't want to be constantly re-writing code

Lilaa3:
> The `remove or add custom preset cmd` button seems to only ever add custom presets. There is no place to edit the property of the operator to change the case. I mean it has been a while for me but afaik this is something you can only do for operators that bring up a menu when you execute them or via a redo
> 
> ```python
>     index: IntProperty(default=-1)
>     op_name: StringProperty()
>     example_name: StringProperty(default="")
> 
>     def execute_operator(self, context):
>         presets = context.scene.fast64.sm64.custom_cmds
>         custom, owner = get_custom_prop(context)
>         conf_type = get_conf_type(context)
>         match self.op_name:
> ```
> 
> So in the code here there is no way to have any case other than the default and no name other than the default

It also removes them in the presets tab.

Lilaa3:
I got a bit excited in making this feature complete- so feel free to also call me out on unnecessary or convoluted stuff

PR #459, opened 2024-09-15
[SM64] Bring binary dl exporter up-to-date

Lilaa3:
- Add bleed and inline support.
- Assert for fmeshes len
- No longer use getDrawLayerV3, yes it doesn´t matter here but it was the last bit of code to use it, with permission I'd like to remove the function


jesusyoshi54 (changes requested):
I tested with insertable binary and direct binary export and they work as expected.
Besides the requested changes I see no issue with the code.

jesusyoshi54 (approved):

PR #460, opened 2024-09-15
[F3D] Backport point lit ucode from AC pr

Lilaa3:
internal_game_update_ver now checks for less than 2, we need to update defaults again lol


fast64_internal/f3d/f3d_material.py
```diff
@@ -570,8 +573,10 @@ def indentGroup(parent: UILayout, textOrProp: Union[str, "F3DMaterialProperty"],
 
         c = indentGroup(inputGroup, "Not useful:", True)
         c.prop(settings, "g_lod")
-        if isUcodeF3DEX1(bpy.context.scene.f3d_type):
+        if f3d.F3DEX_GBI:
```
Dragorn421, line 576: Note this also passes for f3dex2, whereas the previous `isUcodeF3DEX1()` check is specifically f3dex1

Dragorn421 (changes requested):
I can't explain it but loading a blend that has F3DEX3 set as ucode (scene.f3d_type) with this branch, sets the ucode to F3DEX2
It must have something to do with the `enumF3D` change but idk

What's the reasoning for setting the world defaults for positional lighting to true for oot and sm64 when they don't use it in the first place?

fast64_internal/f3d/f3d_material.py
Lilaa3: Good catch! Sorry

Lilaa3:
> I can't explain it but loading a blend that has F3DEX3 set as ucode (scene.f3d_type) with this branch, sets the ucode to 

F3DEX2 It must have something to do with the `enumF3D` change but idk
Going to look at that, shouldn't be happening? 

> What's the reasoning for setting the world defaults for positional lighting to true for oot and sm64 when they don't use it in the first place?

If you're using a point lit ucode in your repo that is most likely how you are implementing it, that's the case for sm64 ligthing engine for example

Lilaa3:
Merged upstream!

PR #461, opened 2024-09-15
[SM64/F3D] Use WriteDifferingAndRevert for bleed

Lilaa3:
This required implementing some new functionality in bleed to be able to keep track of what needs to be reverted for geo modes, othermodes should work by default but testing should still be done


Lilaa3:
Noticed something othermode is doing *sigh* back to vs code I go

Reonu (approved):
This fixed my issues with point lights, thank you

Lilaa3:
I feel like I could improve this further in terms of how resets are handled, will look into it later today, functionality shouldn't change tho unless I find a bug

fast64_internal/f3d/f3d_bleed.py
```diff
@@ -243,12 +263,45 @@ def bleed_mat(self, cur_fmat: FMaterial, last_mat: FMaterial, bleed_state: int):
             commands_bled = copy.copy(gfx)
             commands_bled.commands = copy.copy(gfx.commands)  # copy the commands also
             last_cmd_list = last_mat.mat_only_DL.commands
+
+            # handle write diff, save pre bleed cmds
+            is_rendermode_cmd = lambda cmd: isinstance(cmd, DPSetRenderMode) or (
+                isinstance(cmd, SPSetOtherMode) and cmd.sets_rendermode
+            )
```
jesusyoshi54, line 270: don't use lambda for a named function, make this a def.

fast64_internal/f3d/f3d_bleed.py
```diff
@@ -243,12 +263,45 @@ def bleed_mat(self, cur_fmat: FMaterial, last_mat: FMaterial, bleed_state: int):
...
+            othermode_cmds = [cmd for cmd in commands_bled.commands if isinstance(cmd, SPSetOtherModeSub)]
+            rendermode_cmds = [cmd for cmd in commands_bled.commands if is_rendermode_cmd(cmd)]
+
             for j, cmd in enumerate(gfx.commands):
                 if self.bleed_individual_cmd(commands_bled, cmd, bleed_state, last_cmd_list):
                     commands_bled.commands[j] = None
             # remove Nones from list
             while None in commands_bled.commands:
                 commands_bled.commands.remove(None)
+
```
jesusyoshi54, line 281: I'm not exactly sure what, but something funky is going on here. I inspected some display lists exported using bleed and they were not properly setting the render modes.
the test in question was to have two materials that are the same, but with different geo/render modes. the geo modes were properly set to match the material, but the render modes were not. In fact the render mode setting for the second material was completely missing

jesusyoshi54 (changes requested):
Export using bleed, image differs from expected output.
image
Export without bleed, image matches expectation
image

For reference, I've taken a look at one pair of the offending display lists (the bottom two pictured in the test grid): https://hastebin.com/share/fokuqototo.markdown
You can see that without bleed, I am setting the render mode and othermode settings, and reverting them each time. With bleed, the othermode settings are bled as expected, but so is the render mode

as far as the middle test mats, I am not exactly sure where they went wrong, but they definitely are not right.

jesusyoshi54:
actually let me add that something extra extra wrong is going on with the material logic.
this is the test export on main
w/ bleed:

jesusyoshi54:
The exporter is giving the wrong render mode, please compare:
`gsDPSetRenderMode(GBL_c1(G_BL_CLR_IN, G_BL_A_FOG, G_BL_CLR_MEM, G_BL_1MA) | GBL_c2(G_BL_CLR_IN, G_BL_A_FOG, G_BL_CLR_MEM, G_BL_1MA), AA_EN | Z_CMP | Z_UPD | IM_RD | CVG_DST_CLAMP | ZMODE_OPA | FORCE_BL),`
to settings:
image

KazeEmanuar:
MR has a bug, material bleeding no longer works ("bleed_cmd_list() missing 1 required positional argument: 'bleed_state')

Lilaa3:
> MR has a bug, material bleeding no longer works ("bleed_cmd_list() missing 1 required positional argument: 'bleed_state')

sorry about that, fixed I think

Lilaa3:
Final result with that last commit (wanted to add revert options instead of assuming its fine to skip reverts, like bleed independently but for bones) is a 320 commands reduction on a well configured mario export (disabling reverts in functions, should functions just be revert off by default?). Thank you, everyone!

KazeEmanuar:
I have tested the MR again and found it to fix a bug without introducing any new issues; seems good!

fast64_internal/f3d/f3d_writer.py
```diff
@@ -1706,48 +1673,42 @@ def saveOtherModeHDefinitionIndividual(fMaterial, settings, tlut, defaults):
...
+            fMaterial.revert.commands.append(revert_cmd)
+        flagList, blender = getRenderModeFlagList(settings, fMaterial)
+        cmd.flagList.update(flagList)
+        if blender is not None:
+            cmd.flagList.add(blender)
 
     fMaterial.mat_only_DL.commands.append(cmd)
 
-    if settings.g_mdsft_zsrcsel == "G_ZS_PRIM":
-        fMaterial.mat_only_DL.commands.append(DPSetPrimDepth(z=settings.prim_depth.z, dz=settings.prim_depth.dz))
```
Dragorn421, line 1736: what happened to this? I think it's no longer anywhere writing a DPSetPrimDepth in the "WriteAll" code path

fast64_internal/f3d/f3d_writer.py
```diff
@@ -1706,48 +1673,42 @@ def saveOtherModeHDefinitionIndividual(fMaterial, settings, tlut, defaults):
...
-            cmd.flagList.extend(
-                [
-                    "GBL_c1(" + blendList[0] + ", " + blendList[1] + ", " + blendList[2] + ", " + blendList[3] + ")",
-                    "GBL_c2(" + blendList[4] + ", " + blendList[5] + ", " + blendList[6] + ", " + blendList[7] + ")",
-                ]
-            )
+        if defaultRenderMode:
+            revert_cmd = copy.deepcopy(cmd)
+            revert_cmd.flagList.update(defaultRenderMode)
+            fMaterial.revert.commands.append(revert_cmd)
```
Dragorn421, line 1692: shouldn't `revert_cmd.flagList` be cleared before taking in defaultRenderMode? or if that's intended to keep the g_mdsft_alpha_compare and g_mdsft_zsrcsel settings from above, could you elaborate on why?

fast64_internal/f3d/f3d_bleed.py
```diff
@@ -45,9 +61,61 @@
...
     GbiMacro,
+    get_F3D_GBI,
 )
 
 
+def get_geo_cmds(
+    clear_modes: set[str], set_modes: set[str], is_ex2: bool, matWriteMethod: GfxMatWriteMethod
+) -> tuple[
+    list[SPLoadGeometryMode | SPGeometryMode | SPSetGeometryMode | SPClearGeometryMode]
+    | list[SPGeometryMode | SPSetGeometryMode | SPClearGeometryMode]
```
Dragorn421, line 72: I think you meant `tuple[A, B]` ?
```suggestion
    , list[SPGeometryMode | SPSetGeometryMode | SPClearGeometryMode]
```
\+ formatting

fast64_internal/f3d/f3d_bleed.py
```diff
@@ -45,9 +61,61 @@
...
+]:
+    set_modes, clear_modes = set(set_modes), set(clear_modes)
+    if len(clear_modes) == 0 and len(set_modes) == 0:
+        return ([], [])
+    if is_ex2:
+        if matWriteMethod == GfxMatWriteMethod.WriteAll:
+            return ([SPLoadGeometryMode(set_modes)], tuple())
```
Dragorn421, line 79:
```suggestion
            return ([SPLoadGeometryMode(set_modes)], [])
```

fast64_internal/f3d/f3d_gbi.py
```diff
@@ -4291,8 +4291,14 @@ def geoFlagListToWord(flagList, f3d):
...
-    clearFlagList: list
-    setFlagList: list
+    clearFlagList: set[str] = field(default_factory=set)
+    setFlagList: set[str] = field(default_factory=set)
+
+    def extend(self, clear_list: set, set_list: set):
+        clear_list = self.clearFlagList | clear_list
+        set_list = self.setFlagList | set_list
+        self.setFlagList = set_list - clear_list
+        self.clearFlagList = clear_list - set_list
```
Dragorn421, line 4301: per gbi.h,

> \*	One gSPGeometryMode(pkt,c,s) GBI is equal to these two GBIs.
> \*
> \*		gSPClearGeometryMode(pkt,c)
> \*		gSPSetGeometryMode(pkt,s)

that means the "set" flags take precedence, so shouldn't this be
```suggestion
    def extend(self, clear_list: set, set_list: set):
        clear_list = self.clearFlagList | clear_list
        set_list = self.setFlagList | set_list
        self.setFlagList = set_list
        self.clearFlagList = clear_list - set_list
```
(idk)

Dragorn421 (commented):
tentative code review

Dragorn421 (approved):
Basic OoT import and export seem to be fine :heavy_check_mark: 

f3d_bleed.py completely flies over my head unfortunately

PR #462, opened 2024-09-15
[F3D Bleed] Fix rendermode not being reverted

Lilaa3:
We cannot just assume that othermodeh or othermodel are always setting the same bits, so instead we write the diff of the current cmd to the one already in the reset dict, this means that if a material sets rendermode and another sets alpha compare (for example) it will still revert the rendermode because those bits will be carried over in the reset dict


Lilaa3:
Thank you @Mineqwerty for bringing this up

Lilaa3:
Also done in #461 a bit better, closing

PR #468, opened 2024-10-01
[Repo Settings] Only set if different

Lilaa3:
Fixes lag at start up from draw layers being set unnecessarily. Some clean up that I already did in #461

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -4,7 +4,7 @@ from bpy.path import abspath
 
 from . import addon_updater_ops
 
-from .fast64_internal.utility import prop_split, multilineLabel, draw_and_check_tab
+from .fast64_internal.utility import prop_split, multilineLabel, prop_group_to_json, set_prop_if_in_data
 
 from .fast64_internal.repo_settings import (
     draw_repo_settings,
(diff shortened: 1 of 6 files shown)


fast64_internal/utility.py
```diff
@@ -1896,3 +1896,12 @@ def create_or_get_world(scene: Scene) -> World:
...
         return bpy.data.worlds.new("Fast64")
+
+
+def set_if_different(owner: object, prop: str, value):
+    if getattr(owner, prop) != value:
+        setattr(owner, prop, value)
+
+
+def set_prop_if_in_data(owner: object, prop_name: str, data: dict, data_name: str):
+    set_if_different(owner, prop_name, data.get(data_name, getattr(owner, prop_name)))
```
Dragorn421, line 1907: I think it would be less confusing with more verbosity
```py
    if data_name in data:
        set_if_different(owner, prop_name, data[data_name])
```

__init__.py
```diff
@@ -4,7 +4,7 @@
 
 from . import addon_updater_ops
 
-from .fast64_internal.utility import prop_split, multilineLabel, draw_and_check_tab
+from .fast64_internal.utility import prop_split, multilineLabel, prop_group_to_json, set_prop_if_in_data
```
Dragorn421, line 7:
```suggestion
from .fast64_internal.utility import prop_split, multilineLabel, set_prop_if_in_data
```

Dragorn421 (commented):
code looks good :+1:

Dragorn421 (approved):

PR #472, opened 2024-10-17
Added shields.io badges into README.md

PatoFlamejanteTV:


Dragorn421:
Thanks but we don't need those badges (or any badge at all really)

PatoFlamejanteTV:
Thats ok :)

PR #478, opened 2024-11-27
why does this exist?

SonicTheHedgeHog64:
"fixes" uv map shift, was exporting a render96 mario for my friend but then the uv broke, looked in the code and found this, don't know if it's supposed to be there.


SonicTheHedgeHog64:
or what purpose it poses

AgentXLP:
And because I don't know what this code does I've elected to remove it entirely and hope for the best ^^

SonicTheHedgeHog64:
?

SonicTheHedgeHog64:
Why would I remove code I don't know what it does?

jesusyoshi54:
This exists because tiles in N64 have a size limit of 1024. If the UV is greater than 1024 it needs to be shifted  so that it does not undergo coordinate overflow.
Have you tested this on N64, or is this just a change for the PC port? If it is, please make this change specifically for PC port exports or to exports that override size restrictions

PR #480, opened 2024-12-01
[OoT] Fixed standalone cutscene export

Yanis002:
The operator wasn't using ``Cutscene.new()``, resulting in export issues


Lilaa3 (approved):
Code looks good (it was actually rather simple to test)

PR #481, opened 2024-12-02
[OoT] Update cutscene macro names

fig02:
Find and replace PR + updated the descriptions to align with what decomp has.
Fixes compatibility issues introduced by https://github.com/zeldaret/oot/pull/2311


Yanis002 (approved):
thanks!

PR #482, opened 2024-12-02
Add missing versions to `oot_versions_items` and add overlay support for quick import

Yanis002:
The quick import file was a bit messy so I decided to improve it a bit in the process, also I wanted to add the missing versions to the enum

Now you can use the quick import feature to import data from overlays, I tested it with [`D_808BCE20`](https://github.com/zeldaret/oot/blob/bdc774058ddeabd78cbe961afda61e3c836135f0/src/overlays/actors/ovl_Bg_Treemouth/z_bg_treemouth_cutscene_data.c#L5)

### DIFF
--- a/fast64_internal/oot/__init__.py
+++ b/fast64_internal/oot/__init__.py
@@ -60,14 +60,19 @@ from .tools import (
 
 oot_versions_items = [
     ("Custom", "Custom", "Custom"),
+    ("ntsc-1.0", "ntsc-1.0", "ntsc-1.0"),
+    ("ntsc-1.1", "ntsc-1.1", "ntsc-1.1"),
+    ("pal-1.0", "pal-1.0", "pal-1.0"),
+    ("ntsc-1.2", "ntsc-1.2", "ntsc-1.2"),
+    ("pal-1.1", "pal-1.1", "pal-1.1"),
     ("gc-jp", "gc-jp", "gc-jp"),
     ("gc-jp-mq", "gc-jp-mq", "gc-jp-mq"),
-    ("gc-jp-ce", "gc-jp-ce", "gc-jp-ce"),
     ("gc-us", "gc-us", "gc-us"),
     ("gc-us-mq", "gc-us-mq", "gc-us-mq"),
+    ("gc-eu-mq-dbg", "gc-eu-mq-dbg", "gc-eu-mq-dbg"),
     ("gc-eu", "gc-eu", "gc-eu"),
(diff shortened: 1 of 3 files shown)


fast64_internal/oot/cutscene/importer/classes.py
```diff
@@ -200,6 +206,9 @@ def getCutsceneList(self):
             # if it's none then there's no cutscene in the file
             return None
 
+        if len(parsedCutscenes) == 0:
+            raise PluginError("ERROR: No cutscene was found.")
```
Lilaa3, line 210: Is this intended for scene imports as well? Deku tree inside won´t import with Cutscene importing on now.

fast64_internal/oot/__init__.py
```diff
@@ -60,14 +60,19 @@
...
     ("gc-us", "gc-us", "gc-us"),
     ("gc-us-mq", "gc-us-mq", "gc-us-mq"),
+    ("gc-eu-mq-dbg", "gc-eu-mq-dbg", "gc-eu-mq-dbg"),
     ("gc-eu", "gc-eu", "gc-eu"),
     ("gc-eu-mq", "gc-eu-mq", "gc-eu-mq"),
-    ("gc-eu-mq-dbg", "gc-eu-mq-dbg", "gc-eu-mq-dbg"),
+    ("gc-jp-ce", "gc-jp-ce", "gc-jp-ce"),
     ("hackeroot-mq", "HackerOoT", "hackeroot-mq"),  # TODO: force this value if HackerOoT features are enabled?
     ("legacy", "Legacy", "Older Decomp Version"),
 ]
```
Lilaa3, line 78: Is the reorder here fine? (Enums are saved as an index, you can keep the old index by adding another element to the tuple with your desired internal unique index if needed)

Lilaa3 (changes requested):
Some questions before I approve

fast64_internal/oot/__init__.py
Yanis002: I don't think it is but considering it's just the oot version it's not really an issue, people might get confused I guess, I wanted to have the same order as decomp's makefile (also since ique versions were added since I need to add them in that enum)

fast64_internal/oot/cutscene/importer/classes.py
Yanis002: I tested and it works properly, though it won't work if you already imported that cutscene before, which might be a concern so I changed it to make it print a warning instead of skipping the import entirely

Dragorn421 (approved):

PR #485, opened 2024-12-16
Added checkbox to import all animations when importing an OoT skeleton

Nokaubure:
Added checkbox to import all animations when importing an OoT skeleton
Supports Link skeleton too

image

### DIFF
--- a/fast64_internal/oot/skeleton/importer/functions.py
+++ b/fast64_internal/oot/skeleton/importer/functions.py
@@ -10,8 +10,8 @@ from ...oot_utility import OOTEnum, ootGetObjectPath, getOOTScale, ootGetObjectH
 from ...oot_texture_array import ootReadTextureArrays
 from ..constants import ootSkeletonImportDict
 from ..properties import OOTSkeletonImportSettings
-from ..utility import ootGetLimb, ootGetLimbs, ootGetSkeleton, applySkeletonRestPose
-
+from ..utility import ootGetLimb, ootGetLimbs, ootGetSkeleton, applySkeletonRestPose, ootGetAnimNames
+from ...tools.quick_import import quick_import_exec
 
 class OOTDLEntry:
     def __init__(self, dlName, limbIndex):
@@ -264,6 +264,7 @@ def ootImportSkeletonC(basePath: str, importSettings: OOTSkeletonImportSettings)
 
     removeDoubles = importSettings.removeDoubles
     importNormals = importSettings.importNormals
+    importAnimations = importSettings.importAnimations
     drawLayer = importSettings.drawLayer
 
     skeletonData = getImportData(filepaths)
@@ -284,8 +285,7 @@ def ootImportSkeletonC(basePath: str, importSettings: OOTSkeletonImportSettings)
         actorScale = ootReadActorScale(basePath, overlayName, isLink)
     else:
         actorScale = getOOTScale(importSettings.actorScale)
-
-    # print(limbList)
+    
     isLOD, armatureObj = ootBuildSkeleton(
         skeletonName,
         overlayName,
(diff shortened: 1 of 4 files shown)


fig02:
Considering theres a whole other tabs for animations, maybe this makes sense as a "import all" button there?

fast64_internal/oot/skeleton/importer/functions.py
```diff
@@ -324,3 +324,20 @@ def ootImportSkeletonC(basePath: str, importSettings: OOTSkeletonImportSettings)
...
+
+
+    if importAnimations:
+
+        if armatureObj:
+            bpy.ops.object.select_all(action='DESELECT')
+            
+            armatureObj.select_set(True)
+            
+            bpy.context.view_layer.objects.active = armatureObj
```
Yanis002, line 336: you can use `selectSingleObject` here (from `utility.py`) I think

fast64_internal/oot/skeleton/importer/functions.py
```diff
@@ -324,3 +324,20 @@ def ootImportSkeletonC(basePath: str, importSettings: OOTSkeletonImportSettings)
         applySkeletonRestPose(restPoseData, armatureObj)
         if isLOD:
             applySkeletonRestPose(restPoseData, LODArmatureObj)
+
+
+    if importAnimations:
+
+        if armatureObj:
```
Yanis002, line 331:
```suggestion
        if armatureObj is not None:
```
I think this is the intent here?

fast64_internal/oot/skeleton/utility.py
```diff
@@ -79,6 +79,13 @@ def ootGetLimb(skeletonData, limbName, continueOnError):
             raise PluginError("Cannot handle skeleton limb named " + limbName + " of type " + limbType)
     return matchResult
 
+def ootGetAnimNames(skeleton_data):
+    #Extracts all animation names that start with 'AnimationHeader' from the given skeleton data.
```
Yanis002, line 83:
```suggestion
    """Extracts all animation names that start with 'AnimationHeader' from the given skeleton data."""
```

Yanis002 (commented):
I'll try this when I can (not an animation expert here)

to make CI happy you need to run `black`, the instructions should be in the readme (if you run into issues feel free)

Nokaubure:
> Considering theres a whole other tab for animations, maybe this makes sense as a "import all" button there?

The idea was to make it similar to the old import plugin where loading the skeleton would load all animations. Separating it would be like if the limbs had to be imported with an "import all limbs" button in DL tab, so not a good idea imo.

Yanis002 (commented):
I can't find the menu you're showing on the screenshot but it seems to work for non-Link objects, with Link I get the skeletons but no animations, also you still need to format your branch to make CI happy

about Fig's suggestion: I don't know what would be better since I never use these features but I think for now it's fine?, if people are annoyed by that (or another dev wants) we'll change that

Nokaubure:
> I can't find the menu you're showing on the screenshot but it seems to work for non-Link objects, with Link I get the skeletons but no animations, also you still need to format your branch to make CI happy
> 

rechecked and it works with Link, it takes like 5 minutes tho since there's a lot of animations but does the job, the menu is in those side tabs

image

Yanis002:
sorry, I meant the menu that shows the animations at the bottom (and I'll try Link again)
image

Nokaubure:
> sorry, I meant the menu that shows the animations at the bottom (and I'll try Link again) image

For this you go to dope sheet editor, then select action editor in the dropdown next to it, and then you will see a dropdown showing all animations

fig02:
"Separating it would be like if the limbs had to be imported with an "import all limbs" button in DL tab, so not a good idea imo"

I dont think the comparison to limbs is really fair. Of course, the tool does not have a separate tab for limbs. It does however have a separate tab for animations. Its breaking pre-existing organization of the tool.
Limb display lists are one piece of the skeleton, while animations are entirely separate.

Dragorn421:
I think a checkbox with the skeleton props is fine, mostly because it's simpler than if it was a button elsewhere and you'd have to somehow specify where you want the animations from.
Except for Link you typically want to import at least one animation anyway to not have to work with the folded model so this is just more convenient 

Note the alternative of a separate "importal all" feature would have its merits too, as it'd allow importing all animations from a different object, for example typically os_anime animations over the various skeletons using them. The checkbox doesn't handle this use case (which is fine to me as a known limitation)

Nokaubure:
> so yeah it seems it doesn't work for me with Link, I found why I think, `ootGetAnimNames` returns nothing since Link's animations are in `link_animetion` or something, and this file isn't read it seems, I'm surprised it works for you because afaict it shouldn't?

Reason it works is because its getting the headers from gameplay_keep, and calling the quick import function which looks for the symbols in link_animetion, so not sure why it doesnt for you.

I'm using Legacy mode, maybe its that...

Yanis002:
I'm using the newer system so yeah that's probably the issue

Yanis002:
alright so I'm testing this again and now it seems to work properly for me :tada: I'm confident it was some issue we fixed related to the assets system from decomp, it's still importing the animations (tried adult link) but so far so good :+1:

PR #486, opened 2024-12-18
[OoT] Move collection operators/functions from oot_utility.py

Yanis002:
There's few things to do before having MM support, the main change being making the OoT module of Fast64 a common thing for both games (during my tests I realised it was basically using the same things as OoT so it's better to merge both)

Anyway, I got lost into MM and then I did that, the next step is using `pathlib` everywhere because I had issues with how it's done currently and using that library fixed it


Lilaa3:
This seems like an easy review, could you just resolve the conflicts beforehand?

Yanis002:
done, sorry for the delay on this

Dragorn421 (approved):

PR #488, opened 2025-01-02
[OoT] Fix 2 minor issues

Yanis002:
found two minor issues, the first one is a bug causing the draw to fail if an object has an empty key value, and while investigating this I found another one where rotations would look for an actor's type without checking if it's using one, which caused another bug with the UI

"fixed" the first one by simply checking if the key is empty, and draw a "delete me" label so users can delete the wrong entries (still no idea how it happened though), and fixed the second one simply by checking if the type is used (now that I think about this might not work after all?)

### DIFF
--- a/fast64_internal/oot/room/properties.py
+++ b/fast64_internal/oot/room/properties.py
@@ -49,6 +49,9 @@ class OOTObjectProperty(PropertyGroup):
 
         if isLegacy:
             objectName = ootData.objectData.ootEnumObjectIDLegacy[self["objectID"]][1]
+        elif self.objectKey == "":
+            # just in case there's invalid objects keys, allows users to delete them
+            objectName = "DELETE ME!"
         elif self.objectKey != "Custom":
             objectName = ootData.objectData.objectsByKey[self.objectKey].name
         else:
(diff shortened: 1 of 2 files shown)


Lilaa3:
Waiting on answer/change to my question in the dev channel

> should 
> ```
> if "Rot" in target:
>     type_value = None
> ```
> also check for type?

Yanis002:
this happens to be an issue I fixed recently, fixing the conflicts solved your question I think

that's hilarious actually, I think what this was fixing was the thing reonu found recently but since it got fixed separately now this PR only fix the object list UI issue lol

fast64_internal/oot/room/properties.py
```diff
@@ -49,6 +49,9 @@ def draw_props(self, layout: UILayout, headerIndex: int, index: int, objName: st
 
         if isLegacy:
             objectName = ootData.objectData.ootEnumObjectIDLegacy[self["objectID"]][1]
+        elif self.objectKey == "":
+            # just in case there's invalid objects keys, allows users to delete them
+            objectName = "DELETE ME!"
```
Dragorn421, line 54: since `""` is not a valid value for objectKey I don't understand how this can happen

The id for deleted entries (see `self.deletedEntry` in `OoT_ObjectData.__init__`) is `"None"`, maybe that's what should be checked here instead?
Or `self.objectKey not in ootData.objectData.objectsByKey` ?

PR #489, opened 2025-01-05
[Draft] Global Game Data

Yanis002:
I'm experimenting ways to get around the fact that you can access `bpy.context` (which I need to get the game editor mode for MM), here's what this branch does:
- move `fast64_internal/oot/data` to `fast64_internal/data` so other games can take advantage of that
- only register oot stuff when a blend loads or the game editor mode is changed
- unregister oot stuff when required
- fetch data from the XML files for oot as it registers/unregisters

To do this I had to use the annotation trick we used before for the detailed actor panel as some enums are requiring `game_data`, and there's some issues with leaving them in the class, so far I only had to do this for `actor_id` but on the other hand I didn't remove `ootData` yet so maybe there's more

Advantages:
- less annoying when you have to deal with two games sharing common things (like OoT and MM)
- allows other games to use it
- idk but overall I think this is better

Drawbacks:
- you can't use `game_data` when registering props under `scene.fast64.oot` and `object.fast64.oot` as this needs to be registered with the addon
- probably other things that I'm not aware of yet? idk

also I deleted `fast64_internal/__init__.py` because this file is completely useless and caused issues with this

---
still experimenting so draft for now but feel free if there's any suggestions


Yanis002:
closing this

PR #491, opened 2025-01-07
[OOT + F3D] More informative error messages for deleted lights

Lilaa3:
I thought of wrapping saveOrGetF3DMaterial as well (with a decorator since it's used in multiple places) to avoid these kind of empty errors in the future, I'm fine with reverting that tho I think it is for the better (in ac I wanted to throw errors where the material wasn't in the args already and it's very annoying, this way everything is consistent)

### DIFF
--- a/fast64_internal/oot/exporter/scene/__init__.py
+++ b/fast64_internal/oot/exporter/scene/__init__.py
@@ -71,10 +71,6 @@ class Scene:
                 )
             except Exception as exc:
                 raise PluginError(f"In alternate, cutscene header {i}: {exc}") from exc
-        altHeader.cutscenes = [
-            SceneHeader.new(f"{name}_header{i:02}", csHeader, sceneObj, transform, i, useMacros)
-            for i, csHeader in enumerate(altProp.cutsceneHeaders, 4)
-        ]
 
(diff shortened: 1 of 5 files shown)


fast64_internal/oot/exporter/scene/__init__.py
```diff
@@ -41,19 +41,36 @@ def new(name: str, sceneObj: Object, transform: Matrix, useMacros: bool, saveTex
...
+                raise PluginError(f"In alternate scene header {header}: {exc}") from exc
+
+        altHeader.cutscenes = []
+        for i, csHeader in enumerate(altProp.cutsceneHeaders, 4):
+            try:
+                altHeader.cutscenes.append(
+                    SceneHeader.new(f"{name}_header{i:02}", csHeader, sceneObj, transform, i, useMacros)
+                )
+            except Exception as exc:
+                raise PluginError(f"In alternate, cutscene header {i}: {exc}") from exc
```
Yanis002, line 73: can't highlight more but look right below, I think you forgot to remove the older block? I'm a bit confused on this part

Yanis002 (approved):
the only thing is that generator I mentioned, otherwise looks ok I was able to export my test scene properly

PR #492, opened 2025-01-15
SM64: Add missing typecasts from Gfx* to SetTileSize* to suppress compiler warnings

gheskett:
Only took me 3 years to finally address this lol


Lilaa3 (approved):
Won´t update existing repos, but that should be fine, thank you!

PR #493, opened 2025-01-17
[OoT] Fix issue where light settings is using the wrong name

Yanis002:
not sure how this wasn't caught earlier lol


gheskett:
Apparently Python 3.8+ supports the `final` keyword, which are more or less the same thing as constants. Apart from that, I think Python convention is to use fully uppercase variable names for anything that should behave as if it should be a constant.

PR #496, opened 2025-02-12
[SM64] Add better exception processing to combined exporter

jesusyoshi54:
Typically when exporting all selected with the combined exporter, exceptions were passed over as it would be common for objects to be selected without gfx or col data and to throw an exception.
Now the exceptions I except to appear during this have an extra argument that I catch during exception processing so that I only pass them (e.g. no data type exceptions), and all other valid exceptions will halt the export


Lilaa3 (approved):
Only thing I can comment on is the fact that now empty collision/gfx errors will include the exc warn string in the error message reported on screen. But that doesn't make any difference to the user so this works. Feel free to change it if that bothers you.

PR #497, opened 2025-02-13
Check if fog data exists

Lilaa3:


jesusyoshi54 (approved):
Does this fix an issue/bug? If so you should link it
Anyway looks good

PR #500, opened 2025-02-22
[SM64]: HackerSM64 3.0 support

s4Ys369:
WIP integration of new features and/or changes to allow fast64 compatibility with HackerSM64 3.0

Currently implemented:
- change how model IDs are written to model_ids.h
- Hide model ID number selection when using 3.0

### DIFF
--- a/fast64_internal/sm64/settings/constants.py
+++ b/fast64_internal/sm64/settings/constants.py
@@ -31,4 +31,5 @@ enum_refresh_versions = [
     ("Refresh 16", "Refresh 16", "Refresh 16"),
     ("", "HackerSM64", ""),
     ("HackerSM64 2.3.0", "HackerSM64 (v2.3.0)", "HackerSM64 (v2.3.0)"),
+    ("HackerSM64 3.0", "HackerSM64 (v3.0)", "HackerSM64 (v3.0)"),
 ]
--- a/fast64_internal/sm64/sm64_objects.py
+++ b/fast64_internal/sm64/sm64_objects.py
@@ -1611,9 +1611,33 @@ class SM64_ExportCombinedObject(ObjectDataExporter):
             PluginError("Could not find model_ids.h")
 
         model_id_lines = open(model_ids, "r").readlines()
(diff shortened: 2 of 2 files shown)


fast64_internal/sm64/settings/constants.py
```diff
@@ -31,4 +31,5 @@
     ("Refresh 16", "Refresh 16", "Refresh 16"),
     ("", "HackerSM64", ""),
     ("HackerSM64 2.3.0", "HackerSM64 (v2.3.0)", "HackerSM64 (v2.3.0)"),
+    ("HackerSM64 3.0", "HackerSM64 (v3.0)", "HackerSM64 (v3.0)"),
```
gheskett, line 34: I would call this 3.0.0, if not just so checking patch version is more trivial/consistent if necessary

Mushie64:
Consider setting the defaults when someone is using HackerSM64 for the following settings:
F3D Global Settings > Microcode > default from F3D to F3DEX2/LX2/ZEX
(https://github.com/Fast-64/fast64/blob/main/fast64_internal/f3d/f3d_material.py#L4988)

SM64 General Settings > SM64 Repo Settings > Compression Format > default from MIO0 to YAY0
(https://github.com/Fast-64/fast64/blob/main/fast64_internal/sm64/settings/properties.py#L73)

...or otherwise handle the defaults sensibly to match the ones from the current repo using HackerSM64.
Adding this comment as a decomp beginner who used the supposedly default settings for basically everything and got this (image below) on my first attempt before asking for help. I heard that "this seems to be a fairly common issue among decomp beginners", which I didn't expect to be the case.
decomp_weirdvisuals

s4Ys369:
An update fast64.json with these appropriate settings has been included in hacker as of September https://github.com/HackerN64/HackerSM64/commit/27fedee76fc4cb034ce69bb75841b25cec161835#diff-4e8d2fc8659d21b104963768ed18776f39ad006af57450eb2530ebe0dc8c4063

PR #501, opened 2025-02-24
Actually Fixes #469

### LINKED ISSUE #469: G_AD_NOISE
Calls this function multiple times + makes useless reverts

SonicTheHedgeHog64:
G_AD_DISABLE -> G_AD_NOISE


gheskett:
My understanding of this is that fast64 still defaults to G_AD_NOISE for SM64 exports, however it needs to be able to revert back to G_AD_DISABLE because SM64 actually defaults to this (else it breaks stuff). Therefore, this should not be changed.

SonicTheHedgeHog64:
maybe i need to stop

PR #503, opened 2025-03-12
[OoT] Fixed writing to the spec file

Yanis002:
Recently decomp changed how the spec file is handled, now there's a split for scenes and overlays (also other files that we don't care here), the `spec` file at the root turned to a folder, which contains the spec file with the includes I mentioned. With this fix it should work properly for both legacy and new decomps (untested for legacy). Also there was an issue where the stuff before the first segment would be misinterpreted if it's not present (which is the case for the scene spec includes).

Note: while it should work properly for normal decomp, it's still broken for HackerOoT specifically because of an `#else` in the scene table

### DIFF
--- a/fast64_internal/oot/exporter/decomp_edit/spec.py
+++ b/fast64_internal/oot/exporter/decomp_edit/spec.py
@@ -3,11 +3,25 @@ import bpy
 import re
 
 from dataclasses import dataclass, field
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


Lilaa3 (changes requested):
Just a suggestion but otherwise works like a charm

Lilaa3 (approved):

PR #504, opened 2025-03-19
[OoT] Fixed actor enums overriding actor types

Yanis002:
currently if an actor is using an enum in its data it will ignore the actor's type value, this PR fix this issue (hopefully that's the only major issue in the actor panel :pray: )


krm01:
Tested this myself, fixes the issue. thanks!

krm01 (approved):

Dragorn421 (approved):

PR #505, opened 2025-03-19
[OoT] Add a toggle for the new actor panel

Yanis002:
some people prefer to use the older system, which this PR adds back (as a checkbox in the OoT "File" Settings)

if you're using the older way because of a bug in the new thing please report it first!

### DIFF
--- a/fast64_internal/oot/actor/properties.py
+++ b/fast64_internal/oot/actor/properties.py
@@ -1,3 +1,5 @@
+import bpy
+
 from bpy.types import Object, PropertyGroup, UILayout
 from bpy.utils import register_class, unregister_class
 from bpy.props import EnumProperty, StringProperty, IntProperty, BoolProperty, CollectionProperty, PointerProperty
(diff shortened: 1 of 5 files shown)


fast64_internal/oot/actor/properties.py
```diff
@@ -442,36 +444,40 @@ def draw_props(self, layout: UILayout, altRoomProp: OOTAlternateRoomHeaderProper
...
 
-        if self.actor_id != "Custom":
-            self.draw_params(actorIDBox, obj)
-        else:
-            prop_split(actorIDBox, self, "actor_id_custom", "")
+        if bpy.context.scene.fast64.oot.use_new_actor_panel:
+            if self.actor_id != "Custom":
+                self.draw_params(actorIDBox, obj)
+            else:
+                prop_split(actorIDBox, self, "actor_id_custom", "")
```
Lilaa3, line 451: The custom id is being hidden but still being exported?

fast64_internal/oot/actor/properties.py
```diff
@@ -504,14 +510,15 @@ def draw_props(
...
 
-        if self.actor.actor_id == "Custom":
-            prop_split(actorIDBox, self.actor, "actor_id_custom", "")
-        else:
-            self.actor.draw_params(actorIDBox, roomObj)
+        if bpy.context.scene.fast64.oot.use_new_actor_panel:
+            if self.actor.actor_id == "Custom":
+                prop_split(actorIDBox, self.actor, "actor_id_custom", "")
+            else:
+                self.actor.draw_params(actorIDBox, roomObj)
```
Lilaa3, line 517: Same here.

Lilaa3 (changes requested):

fast64_internal/oot/actor/properties.py
Yanis002: I'm not sure what you mean in both cases, all this does is display the custom string prop if the custom id was chosen (otherwise draw the param's ui for that actor)

Lilaa3: Yes, but the custom id does not show up under the old actor panel yet is still exported, that's what I was commenting on.

Dragorn421: Yeah as lila is saying this change means you can't set a custom actor id using "old actor panel"
image
Not sure what you intended

Yanis002: should be fixed now

PR #507, opened 2025-04-01
[OoT] Fixed collision export

Yanis002:


krm01 (approved):
removing z64.h actually broke my project in unfortunate ways, since i was using it as a compatibility layer for new fast64 definitions with a very old decomp fork. but that's a me problem i guess. after manually fixing the export to include all the right compatibility defines, it worked for me. so i couldn't test the out-of-the-box compilation, but i assume it works for people on the latest decomp version.

fast64_internal/oot/collision/exporter/to_c/collision.py
```diff
@@ -309,7 +321,6 @@ def exportCollisionToC(
 
             # write file
             path = ootGetPath(exportPath, isCustomExport, "assets/objects/", folderName, False, True)
-            filename = exportSettings.filename if exportSettings.isCustomFilename else f"{name}_collision"
             writeCData(colData, os.path.join(path, f"{filename}.h"), os.path.join(path, f"{filename}.c"))
             if not isCustomExport:
                 addIncludeFiles(folderName, path, name)
```
Lilaa3, line 326: Game wouldn´t compile without this, I assume this is a bug?
```suggestion
                addIncludeFiles(folderName, path, filename)
```

Yanis002: should work properly now (with latest decomp)

Dragorn421 (commented):
There should be some backwards compatibility code path for older repos (currently this would fail compilation due to array_count.h being a newer header)

I think gating it behind something like "`(decomp_path / "include/z64.h").exists()`" if that works would be nice

Yanis002:
done :+1:

Dragorn421 (approved):

PR #508, opened 2025-04-01
[OoT] Update actor scale's default value

Yanis002:
it was mentioned on discord how annoying the value 100 is when exporting actor stuff, this PR change that to 10 (which was the recommended value), also another change: if the auto-detect don't return a value it will default the scale to the one from the `actorScale` value (which is now 10 by default)

discord discussion: https://discord.com/channels/874816709855440926/874819026629574717/1353518470808997980


Lilaa3:
If you can't find an oot tester, become the oot tester.
Works well, but I think this might be relevant https://github.com/Yanis002/fast64/pull/7

Yanis002:
yeah that makes sense

Lilaa3 (approved):
lgtm

PR #509, opened 2025-04-06
[OoT] Fixed UI issue when an actor has no type

Yanis002:
found by @Reonu, also this adds a collectible entry for "item none"


Reonu (approved):
very epic PR, fixed my issue, made me dinner and found my lost cat 👍

PR #510, opened 2025-04-07
[F3D] F3D bleed render mode reset fix

jesusyoshi54:
If you had a material with a changed render mode, and followed that up with a material that only changed the lowest 3 bits (alpha comapre/z src sel) then the bleed reset dictionary would only know to reset the low 3 bits instead of the full render mode.
I have elected to just always change the full othermode_L whenever it is changed.


Lilaa3:
Won't this always add a revert even if not needed? I went to implement this pr now to not lag behind in my own write diff impl and it seems that making othermode hashable is more correct https://github.com/Lilaa3/fast64/commit/01ba728231d343e9b7399453ee7e62bf6bf8ec9c

Lilaa3:
I have this same change in my own pr, so if you'd like to kill two birds with one stone you could also do that

PR #512, opened 2025-04-15
[SM64] Bring more props to combined and merge level enums

### LINKED ISSUE #498: [SM64] Options with level exporting don't have "intro" level as an option
Title. As you can see in the following screenshot, the Options with level exporting don't have the Title Screen level as an option.
My request is to add the Title Screen level as an option in them.
Thank You.

Image

Lilaa3:
There is only one level enum (which I took the time to organize), binary will simply error out with Custom (tho it is not fully out of the possibility to support custom levels in binary in the future), upgrade code added. Dl and animation now share props with combined, animation rework actually uses these but for now both just redraw the prop to keep the ui the same. Imports and address converter left untouched besides enum change, these should not be per scene. Convert to str in getPathAndLevel to avoid issues with posix paths -> abspath

Alternatively, should this be just a pr for level enums? Closes #498


jesusyoshi54:
Should `scene.DLName` be merged with `sm64.combined_export.object_name` be merged with `scene.animName`?
Code looks good and it seems to work well from my testing.

jesusyoshi54 (approved):
lol ok

PR #518, opened 2025-04-27
[OoT] Remove `global.h`, `macros.h` and `z64.h` in the exported includes

Yanis002:
This removes `global.h`, `macros.h` and `z64.h` includes since these files are now deleted in decomp (well, not `z64.h` but it will be removed soon), this changes skeleton, animation, cutscene, DL and scene exporters, the collision one is already done in another PR. I also updated the documentation's wrong links and added "ifdef-define-endif"s in the headers to make it consistent with the other headers from decomp, it exports properly and it looks right but I didn't try to compile (yet)

### DIFF
--- a/fast64_internal/oot/animation/exporter/classes.py
+++ b/fast64_internal/oot/animation/exporter/classes.py
@@ -6,8 +6,9 @@ def convertToUnsignedShort(value: int) -> int:
 
 
 class OOTAnimation:
-    def __init__(self, name):
+    def __init__(self, name, filename: str):
         self.name = toAlnum(name)
+        self.filename = filename
         self.segmentID = None
         self.indices = {}
         self.values = []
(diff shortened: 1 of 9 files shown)


Yanis002:
(reminder for myself: old behavior support when the version is set to legacy)

Dragorn421:
> (reminder for myself: old behavior support when the version is set to legacy)

does this mean this PR is a draft?

Yanis002:
it wasn't initially but when we talked about it on discord I realised I should probably keep things compatible with the legacy version specifically and I didn't take the time to do that yet

Yanis002:
I added backwards compatibility when the OoT version is set to "legacy", lmk if I should instead check if, say, z64.h exists instead of doing `oot_version == "legacy"` like you suggested in #507

Dragorn421:
I do think checking on z64.h would be better, because there's quite a big range of decomp revisions that have version support (so post-legacy) and don't have headers split (so pre-yeet-z64.h)

Dragorn421 (approved):
It does compile pre/post- global.h :)

PR #522, opened 2025-05-17
[F3D] Use set for image and pal dependencies

Lilaa3:
Prevents duplicates, causing the behavior seen in this issue https://discord.com/channels/874816709855440926/1373127651283439636


Dragorn421 (approved):
Exporting did not break :heavy_check_mark:

Issue #527, opened 2025-05-22
[OoT] Cutscene exports don't support IDO


fig02:
I'll admit I'm in the minority here, having a case where I want to mod with IDO instead of GCC.

But, if at all possible, it'd be nice to have a toggle to switch between floats and hex floats, or have the output use the CS_FLOAT macro like in vanilla cutscenes. Without it, cutscenes are completely broken on IDO

Yanis002:
the export is using `DEG_TO_BINANG`, I'm surprised it doesn't work on IDO 🤔 
I'll change that to use `CS_FLOAT` instead then

fig02:
In my case there are camera points that have just a float on its own (I think it's for the roll field, but I'm not at a computer to check). I didn't see DEG_TO_BINANG used anywhere

Yanis002:
oooh you're right, `DEG_TO_BINANG` is only used for actor cues currently (or non-camera stuff at least)

Issue #526, opened 2025-05-22
[OoT] Cutscene exports missing z64player include


fig02:
Z64player.h is needed for player cue IDs. 
I'd recommend taking a look at the includes the new assets system is outputting and mirroring that (maybe for all asset types, just to be safe).

Yanis002:
I noticed that earlier, I fixed it in #525 (I thought it wasn't worth a PR for two lines), also it lacked `z64ocarina.h` for me (because of the ocarina action stuff), I'll look at the extracted includes though it could reveal more missing stuff

PR #528, opened 2025-05-22
[OoT] Scene/Cutscene Export Fixes

Yanis002:
Addresses #526 and #527, I added 3 missing includes and make camera points use `DEG_TO_BINANG` to fix compile issues with IDO (I tested and it seems to work properly now), also I noticed `getOoTRotation` was exporting floats as doubles I think so I fixed that too (though it probably doesn't matter because of the macro but it bothered me)

### DIFF
--- a/fast64_internal/oot/exporter/cutscene/camera.py
+++ b/fast64_internal/oot/exporter/cutscene/camera.py
@@ -1,3 +1,5 @@
+import struct
+
 from dataclasses import dataclass, field
 from ....utility import PluginError, indent
 from ...cutscene.motion.utility import getInteger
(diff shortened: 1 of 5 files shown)


fig02 (commented):
Tested, I still seem to be missing z64player.h.
Im exporting a cutscene into a file on its own, not as a part of a scene file. I think this PR fixes scene cutscenes only.

My header has these:
```c
#include "ultra64.h"
#include "sequence.h"
#include "z64math.h"
#include "z64cutscene.h"
#include "z64cutscene_commands.h"
```

Yanis002:
oops, should work now (I tried the scene exporter, I forgot about this one)

fig02:
Thanks, the includes are fixed now.

The camera is still broken though.
Thinking about it, `DEG_TO_BINANG` does not work.

For example, `DEG_TO_BINANG` converts 60.0f to its binary angle representation which is 0x2AA8.
What really needs to happen though, is the float itself represented as a hex number, so 0x42700000.

So, I think `CS_FLOAT` is necessary

fig02 (approved):
works perfectly now, thank you

fast64_internal/oot/exporter/cutscene/camera.py
```diff
@@ -32,8 +34,9 @@ def getCmd(self):
         if len(self.pos) == 0:
             raise PluginError("ERROR: Pos list is empty!")
 
+        angle_ieee = f"0x{struct.unpack('<I', struct.pack('<f', self.viewAngle))[0]:X}"
```
Dragorn421, line 37:
```suggestion
        viewAngle_ieee = f"0x{struct.unpack('<I', struct.pack('<f', self.viewAngle))[0]:X}"
```

fast64_internal/oot/exporter/cutscene/data.py
```diff
@@ -1,5 +1,6 @@
 import bpy
 import math
+import struct
```
Dragorn421, line 3:
```suggestion
```

Dragorn421 (approved):

PR #530, opened 2025-05-23
[OoT] Add move cutscene bone operator

Dragorn421:
Add buttons to move cutscene bones around in OoT cutscenes, since the exporter (and preview *as of this PR*) use them in the order they're displayed in.
Note rather than being actually *moved*, the bones are deleted and re-created

screenshot:

<details>

image

</details>


fig02 (approved):
works perfectly, thank you.
makes cutscene editing much less stressful!

Yanis002 (commented):
I think the same thing should be done for actor cues but that can wait it's just a thought I had, otherwise it works fine

Yanis002 (approved):

PR #532, opened 2025-05-26
[UI (Super Low priority)] Stop using HIDE_HEADER

Lilaa3:
I showed this off a while back in the development channel, going through some of my old branches to see what's worth keeping and thought it was worth bringing this idea back. It definitely makes fast64 stand out less and it looks a lot better than boxes.
image


Dragorn421 (commented):
I didn't realize that the way this goes about not using HIDE_HEADER is to instead parent every panel to the built-in HIDE_HEADER top panels
That's not much better for integrating seamlessly into blender

What I would like to see is an option (in the addon preferences) for fast64 panels to be regular panels (not using HIDE_HEADER and not being parented to the built-ins). I know that means fast64 won't be at the top, that's actually my reason for wanting it (fast64 gets in the way being stuck there)

PR #533, opened 2025-05-26
[OoT] Exporter Cleanup

Yanis002:
This PR cleans up the older code that was still present, for animations and skeletons I moved the exporters in the `oot/exporter` folder to group them with the other ones, but I didn't improve them, on the other hand I refactored the collision and cutscene exporters to integrate them nicely in the new exporter (and they're much cleaner imo)

I also removed a feature, the one that lets you insert cs motion data into an existing cutscene, for several reasons:
- I don't even know if it still works lol
- I don't like it
- I couldn't find a decent way to integrate it nicely
- It's such an obscure feature anyway, I doubt anyone was using that, and if someone did then simply use "export cs motion data only" and copy paste the output wherever you need, it's not that bad


Dragorn421 (commented):
one nit

Dragorn421 (approved):
Quickly tested export of map, cutscene, skeleton, nothing to report :heavy_check_mark:

PR #541, opened 2025-05-29
Create Game Data and move `fast64_internal/oot/data`

Yanis002:
This PR does the following:
- move "oot data" to a global "data" folder in a z64 subfolder
- remove `ootData` and fix the related issues

this explains why it's a big PR, lots of changes needed to be done to fix the `ootData` stuff but the main focus of that PR is the system handling the XML files

---
Now game data isn't only for OoT (and soon MM), any game can add its own system! To add one, in the `fast64_internal/data` subfolder create a folder for your game (example: `fast64_internal/data/sm64` then you'll need to create your game data system, OoT is using XML files and got functions to read them and create blender-enum item lists based on this, then go to `game_data.py` and add a variable for your game (example: `self.sm64 = SM64_Data()`, and finally you'll need to implement the update, just add an `elif` (or alternatively turn this to a `match-case` statement)


fast64_internal/data/z64/xml/oot_enum_data.xml
```diff
@@ -683,4 +683,126 @@
...
+    <Item Key="sdc_fishing_pond" ID="SDC_FISHING_POND" Name="Fishing Pond (Turibori)" Index="50"/>
+    <Item Key="sdc_ganons_tower_collapse_interior" ID="SDC_GANONS_TOWER_COLLAPSE_INTERIOR" Name="Ganon's Tower (Collapsing) (Ganon Sonogo)" Index="51"/>
+    <Item Key="sdc_inside_ganons_castle_collapse" ID="SDC_INSIDE_GANONS_CASTLE_COLLAPSE" Name="Inside Ganon's Castle (Collapsing) (Ganontika Sonogo)" Index="52"/>
+  </Enum>
+  <Enum Games="OoTMqDbg" Key="global_object" ID="GlobalObjects">
+    <Item Key="object_gameplay_field_keep" ID="OBJECT_GAMEPLAY_FIELD_KEEP" Name="Overworld" Index="1"/>
+    <Item Key="object_gameplay_dangeon_keep" ID="OBJECT_GAMEPLAY_DANGEON_KEEP" Name="Dungeon" Index="2"/>
+    <Item Key="object_invalid" ID="OBJECT_INVALID" Name="None" Index="0"/>
+  </Enum>
+  <Enum Games="MM" Key="cs_spline_rel" ID="CutsceneCamRelativeTo">
```
Dragorn421, line 746: MM stuff in an oot_ file?

Dragorn421 (commented):
will you need to restart blender to switch between OoT and MM? I ask because it seems some props enums are gotten using the current gameMode like
`chest_content: EnumProperty(items=game_data.z64.actors.ootEnumChestContent, default="item_heart")`

even `actor_id: EnumProperty(items=lambda self, context: game_data.z64.actors.getItems(self.actor_user))` looks sus, I'm not sure what happens exactly when items change but some props are already set using the items. I know that at the very least Blender expects python to keep a hold on the previous data for it to not get garbage collected while it's being used on the C side, iirc

Yanis002:
It will reload the z64 parts of the plugin when you switch the game mode (reload as in unregister and register again), currently it's not doing that because it's handled in a draft PR that I plan to close and redo (maybe I already closed it idr), so no need to restart blender, for the enums hardcoded with OoT stuff it will either become a dynamic enum or another prop will be added (idr exactly why but there was cases like this on the mm_dev branch)

fast64_internal/data/z64/xml/oot_enum_data.xml
Yanis002: it's for the "enable MM features" checkbox

fast64_internal/data/z64/data.py
```diff
@@ -0,0 +1,944 @@
...
+            self.ootEnumCameraSType = ootEnumCameraSType
+            self.ootEnumCSListType = ootEnumCSListType
+            self.skeleton_dict = oot_skeleton_dict
+            self.enum_skeleton_mode = oot_enum_skeleton_mode
+        elif self.game == "MM":
+            self.cs_index_start = 1
+            self.cs_list_type_to_cmd["Transition"] = "CS_TRANSITION_LIST"
+            self.cs_list_type_to_cmd["RumbleList"] = "CS_RUMBLE_LIST"
+            self.ootEnumNightSeq = enum_ambiance_id
+            self.ootEnumSkybox = mm_enum_skybox
```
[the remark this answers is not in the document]
Dragorn421: I think in this PR would be fine but if you prefer doing it in a followup PR that's fine by me too

Dragorn421 (approved):

PR #544, opened 2025-05-30
[SM64] Make switch material override stackable and apply layer reverts

Lilaa3:
Also rewrite the code to do it all in one go because ew wtf was that old code

### DIFF
--- a/fast64_internal/f3d/f3d_bleed.py
+++ b/fast64_internal/f3d/f3d_bleed.py
@@ -269,6 +269,10 @@ class BleedGraphics:
         start_cmds = cmd_list.commands  # commands that preceed any jump list
         for jump_list_cmd in fmesh_jump_cmds:
             # bleed mat and tex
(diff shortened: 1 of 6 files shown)


Dragorn421 (commented):
sm64 flies over my head, I'll need a sm64 person to tell me this works / fixes something

Lilaa3:
> sm64 flies over my head, I'll need a sm64 person to tell me this works / fixes something

it definitely works in my main test case, mario. this fixes models that use override draw layer BUT use set rendermode (if you have a material with set rendermode in the opaque layer, it would export a revert for opaque, but then if you overrode that layer it would not change the revert therefor reverting to opaque in say a transparent layer, instead now we see if there is any revert worth adding/replacing when changing layers of a dl) and cases where you apply more than one switch mat override to an hierarchy

Lilaa3:
Fyi gotta fix bleed first for some stupid reason

PR #546, opened 2025-05-30
[OoT] Fix animation exporter expecting an object name with custom paths

Yanis002:
Currently the animation exporter assumes you never use a custom path, which can raise an error with `No name entered for the exporter.` if the object name happens to be empty, this PR addresses that issue

### DIFF
--- a/fast64_internal/oot/animation/operators.py
+++ b/fast64_internal/oot/animation/operators.py
@@ -18,10 +18,17 @@ from ..oot_utility import (
 
 
 def exportAnimationC(armatureObj: bpy.types.Object, settings: OOTAnimExportSettingsProperty):
+    if settings.isCustom:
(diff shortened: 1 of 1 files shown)


fast64_internal/oot/animation/operators.py
```diff
@@ -18,10 +18,15 @@
 
 
 def exportAnimationC(armatureObj: bpy.types.Object, settings: OOTAnimExportSettingsProperty):
+    if settings.isCustom:
+        checkEmptyName(settings.filename)
```
Dragorn421, line 22: shouldn't filename only be checked if isCustomFilename (which can be checked regardless of isCustom)

Dragorn421 (approved):

fig02 (approved):
thank you!

PR #547, opened 2025-06-01
[OoT] Fix the importers

### LINKED ISSUE #524: [OoT] Assets system changes broke importing
Since [zeldaret/oot 1e556e3a3](https://github.com/zeldaret/oot/commit/1e556e3a3de0a485837a1277d82d765c1be690f7) and the introduction of a new assets system, fast64 can no longer import assets.

Workaround: use an older revision of decomp, at least for importing from it.
The latest decomp revision prior to the new assets system is `0c6c112cb9a3a8f12d0f4865743853857bbdc88e`, check it out using git with:
`git checkout 0c6c112cb9a3a8f12d0f4865743853857bbdc88e`

Yanis002:
This PR address "all" the issues with decomp's new assets system while keeping compatiblity with the older system (automatically based on file data content)

I haven't tested this on everything but I'm confident enough on this, any remaining issue should be some edge-case that I missed, my tests were done with HackerOoT and classic up-to-date decomp, both works well from what I can tell (tried hyrule field, jabu, deku tree, the default gerudo skeleton/animations and adult link skeleton/animations)

Something that PR is also probably fixing is importing scene exported from Fast64, I'm not sure if that was broken though but I was able to import my test scene properly

Also note that I agree using scene exporter classes is a bit weird but it will make sense eventually once the scene importer is improved (I think it would be nice to merge both)

closes #524


fast64_internal/f3d/f3d_parser.py
```diff
@@ -847,16 +857,22 @@ def getSizeMacro(self, size: str, suffix: str):
         else:
             return getattr(self.f3d, self.f3d.IM_SIZ[size] + suffix)
 
-    def getImagePathFromInclude(self, path):
+    def getImagePathFromInclude(self, path, skip_base_path: bool = True):
```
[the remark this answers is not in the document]
Yanis002: I meant to keep original behavior, I accidentally inverted it

fast64_internal/oot/importer/scene.py
```diff
@@ -104,28 +106,69 @@ def parseScene(
...
     f3dContext = OOTF3DContext(get_F3D_GBI(), [], bpy.path.abspath(bpy.context.scene.ootDecompPath))
     f3dContext.mat().prim_color = (0.5, 0.5, 0.5, 0.5)
     f3dContext.mat().env_color = (0.5, 0.5, 0.5, 0.5)
 
+    # disable TLUTs only if we're trying to import a scene from the new assets system
+    f3dContext.ignore_tlut = sharedSceneData.use_macros and not sharedSceneData.is_fast64_data
+
     parseMatrices(sceneData, f3dContext, 1 / bpy.context.scene.ootBlenderScale)
     f3dContext.addMatrix("&gMtxClear", mathutils.Matrix.Scale(1 / bpy.context.scene.ootBlenderScale, 4))
+    f3dContext.addMatrix("&gIdentityMtx", mathutils.Matrix.Identity(4))
```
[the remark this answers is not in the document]
Yanis002: apparently that wasn't it unfortunately

fast64_internal/f3d/f3d_parser.py
```diff
@@ -2244,24 +2277,41 @@ def getImportData(filepaths):
...
-            for j in range(4):
-                matrix[j][i] = trueValues[i * 4 + j]
+
+        if "#include" in data:
+            match = re.search(r".*\((.*?)\)", get_include_data(data, strip=True), re.DOTALL | re.MULTILINE)
+            assert match is not None
+            raw_matrix = match.group(1).split(",")
+            for i in range(4):
+                for j in range(4):
+                    matrix[j][i] = float(raw_matrix[i * 4 + j].removesuffix("f"))
```
Dragorn421, line 2297:
```suggestion
                    matrix[i][j] = float(raw_matrix[i * 4 + j].removesuffix("f"))
```
this fixes jabu afaict

fast64_internal/oot/cutscene/classes.py
```diff
@@ -49,7 +49,7 @@ class CutsceneCmdCamPoint(CutsceneCmdBase):
     def __post_init__(self):
         if self.params is not None:
             self.continueFlag = self.params[0]
-            self.camRoll = getInteger(self.params[1])
+            self.camRoll = getInteger(self.params[1], not bpy.context.scene.fast64.oot.is_globalh_present())
```
Dragorn421, line 52: again not a fan of checking for global.h as a "is newer assets system"

what changed here in the assets?

Dragorn421: bruh

how did 0xFFFFFFE2 work previously?

fig02:
The idea with this new assets system in decomp is that some assets will start to be committed over time. As in, not all assets are going to be included from a .inc.c file.
Does the new importer account for both possibilities? data inlined, and data coming from an include?

I dont know fast64 well enough to review the code and check myself, just looking out to help make this compatible for the future with the direction decomp intends to move in.

PR #556, opened 2025-07-01
[OOT/MM] Get enums from `game_data` part 1

Yanis002:
This PR moves enum usage to the `game_data` module, also switches some of the enums to the ones generated from the xml data for the ones that are already present in the xml file, the other ones will be treated in another PR

There's a change in the new actor panel that I don't think will cause issues, basically I moved the generated props to `.oot.XXX` for both games instead of having them directly under `.ootActorProperty`, there's still changes needed for actors though


fast64_internal/z64/actor/operators.py
```diff
@@ -12,7 +12,7 @@ class OOT_SearchChestContentEnumOperator(Operator):
     bl_property = "chest_content"
     bl_options = {"REGISTER", "UNDO"}
 
-    chest_content: EnumProperty(items=game_data.z64.actors.ootEnumChestContent, default="item_heart")
+    chest_content: EnumProperty(items=lambda self, context: game_data.z64.get_enum("chest_content"), default=1)
```
sauraen: Does this work? I thought properties had to have enums which were permanently fixed. What happens if the enum was {A, B, C}, the property was set to B, and then the enum was changed to {D, E, F}? Does the property change to the default value of the new enum? If it keeps the index, what if it was previously C and the enum was changed to {D, E} (out of bounds)?

sauraen: I don't mean if someone edits the XML data but if someone changes the selected game. This could be, import something from OoT, change to MM, export it, or it could be just start Blender in the default scene and change the game. I'm not really worried about correctness, but about Blender not crashing. Anyway just test it.

Yanis002: each game got their own enums, only the operator is common iirc so it should be fine, see `get_prop_name` and `create_game_props` (previously `initOOTActorProperties`) in `actor/properties.py`

Yanis002:
found another tiny issue with my test blend, the fade out seq player type enum is acting strange, I have one blank item and nothing makes sense, this PR's enum and the main branch one are literally the same except for the explicit indices and even after adding the necessary stuff for the seq player it's still broken so idk, not a big deal though you just have to set the value again and that's it but yeah idk, not a big issue anyway so I think it's fine (also I don't think a lot of people will be affected by this anyway, at least I hope lol)

otherwise I fixed the surface type export issue I found the other day so I think this PR should be ready to merge :muscle:

PR #559, opened 2025-07-09
[SM64] Fix mesh instancing, regression from #423

Lilaa3:
Oopsie


jesusyoshi54:
Can you describe what the bug was and how I can test it?

Lilaa3:
> Can you describe what the bug was and how I can test it?
It was checking for geo empty

PR #565, opened 2025-07-16
[F3D + SM64] Dedup mesh names, fixes repeated final revert DLs

Lilaa3:


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -2440,14 +2440,16 @@ def addLight(self, key, value, fMaterial):
...
-        checkUniqueBoneNames(self, meshName, name)
-        self.meshes[meshName] = FMesh(meshName, self.DLFormat)
-
-        self.onAddMesh(self.meshes[meshName], contextObj)
-
-        return self.meshes[meshName]
+    def addMesh(self, name, namePrefix, drawLayer, isSkinned, contextObj, dedup=False):
+        final_name = getFMeshName(name, namePrefix, drawLayer, isSkinned)
+        if dedup:
+            for i in range(1, len(self.meshes) + 2):
```
jesusyoshi54, line 2446: why would you make this a for loop? If you have X copies of revert DLs with the same name, surely there is a better way to make sure each one has a unique name?
This operator is called upon each revert node being added, so you should be able to just reference the length of `self.meshes` to get a unique name, or you could also find the number of meshes with the name, `final_name` within them and then append n+1 to the end of it.

jesusyoshi54 (approved):

PR #569, opened 2025-08-02
[OoT] Fix cutscene export issues

Yanis002:
- fixes an issue where extra cutscenes couldn't be exported
- ~~added a check to check if the path exists for the standalone exporter, for some reasons relative paths are broken for me but it makes no sense at all, I don't see why it wouldn't work but absolute paths works properly~~ actually no forget about that I'm dumb


Dragorn421 (approved):
I didn't test but this is simple and has been tested by others

PR #570, opened 2025-08-04
Remove legacy & unused f3d_render_engine

Dragorn421:
The reason to remove it now besides that it never had any use, is that apparently the bgl module is going away in Blender 5.0 and fast64 using it (through f3d_render_engine.py) prevents enabling fast64 in 5.0


Lilaa3 (approved):
f64render shall reign them all

PR #571, opened 2025-08-04
Fixup 565 normal import

Dragorn421:
Discovered while importing adult link

Now why does this code even run in that case, I have no idea


Yanis002 (approved):
works properly

PR #575, opened 2025-08-22
[Codebase] Restore original functionality to setOrigin

Lilaa3:
the remade version of set origin was not equivalent to the previous one, breaking SM64. we were applying -target_loc instead of the offset between target loc and the original location

### DIFF
--- a/fast64_internal/utility.py
+++ b/fast64_internal/utility.py
@@ -722,6 +722,10 @@ def setOrigin(obj: bpy.types.Object, target_loc: mathutils.Vector):
         active_object=obj,
     ):
         obj.data.transform(mathutils.Matrix.Translation(-offset))
+        # Applying location puts the object origin at world origin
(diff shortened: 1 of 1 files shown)


jesusyoshi54:
does this affect OOT?

Lilaa3:
> does this affect OOT?


I was waiting for dragorn to look at it yeah.

Lilaa3:
Yanis if you have the time to test that would be fine as well

Yanis002:
I'll try to look at it tomorrow

Yanis002 (approved):
should be fine for oot (sorry for the infinite delay on the review)

Lilaa3:
Merging cause it's a pretty awful bug. Thank you yanis :steamhappy:

avereyishere:
so this is why my models weren't exporting correctly, will update pronto!

PR #578, opened 2025-08-30
Update object.py

Santitub:


fast64_internal/z64/object.py
```diff
@@ -4,35 +4,54 @@
...
+                header.objects.objectList.append(obj_id)
+                existing_object_ids.add(obj_id)
+                add_missing_object_to_prop(room_obj, header_index, object_key)
+
+
+def add_missing_objects_to_all_room_headers(room_obj: Object, headers: list[RoomHeader]):
+    """Adds missing objects required by actors to all room headers and the corresponding Blender object."""
+    for index, header in enumerate(headers):
+        if header:
+            add_missing_objects_to_room_header(room_obj, header, index)
```
Yanis002, line 57: why removing code that makes the intent clear? also you removed the final newline
```suggestion
    for index, header in enumerate(headers):
        if header is not None:
            add_missing_objects_to_room_header(room_obj, header, index)

```

Yanis002 (changes requested):
I'm probably being super paranoid but idk I have a feeling this was done through AI or something, either way there's inconsistencies with the codebase style that needs to be addressed, mostly `None` checks being removed and I don't like it at all because it can make things not clear at all

also I appreciate the cleanup and the snake_case, will try this PR later though I wanted to do a code review first

Yanis002:
I'm actually gonna close this PR, checking your profile made me realise how very very very likely AI usage was and how you just makes random PRs to random projects so I don't think you're actually interested in Fast64.

Issue #179, opened 2022-11-16
Property versioning system is broken


kurethedead:
`OOT_ObjectProperties` right now uses an IntProperty to hold its current version. However, in Blender, properties whose values haven't been changed from the defaults do not preserve their values when the default is changed. I haven't checked in depth, but I think SM64 also has this issue. For data added through operators this isn't an issue as the version can be set on creation (ex. F3D material creation), but if a user creates empties without using Fast64 operators, then the version will not be preserved on future updates.

PR #579, opened 2025-08-31
[SM64] Concept: Use desgraph updates to set current versions at runtime

Lilaa3:
Instead of always running upgrade code on new objects/bones/scenes on file load, we make sure they get the new version as soon as possible. This is ideal considering we are gathering properties from keys that could be very likely exist in another addon.

That said, I feel like there has to be a better approach to this? I can´t really figure out one. This does work it just feels hacky I guess, even if this is how materials work.

Related to [#vversionersion](https://github.com/Fast-64/fast64/issues/179)

@Yanis002 Would this work for oot too? Is it even worth it considering oot at least uses oot as a prefix in old names aha.

### DIFF
--- a/fast64_internal/sm64/animation/properties.py
+++ b/fast64_internal/sm64/animation/properties.py
@@ -883,7 +883,7 @@ class SM64_AnimProperties(PropertyGroup):
 
     importing: PointerProperty(type=SM64_AnimImportProperties)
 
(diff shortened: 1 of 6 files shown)


Yanis002:
this should probably work for OoT but is it still upgrading blends like before? I don't really follow what's going on

PR #580, opened 2025-09-03
[SM64] Fix custom bones

Lilaa3:
Someone updated the code for simplicity but didn't test it (I'm very sorry)

Tested levels and armatures.


Anoliah:
About to post an issue but then saw this PR, can confirm it fixes the issue. Merge soon would be nice.

PR #588, opened 2025-09-29
[Codebase] Fix setOrigin AGAIN

Lilaa3:
transform_apply was not neeeded, typehint was missing and offset was being affected by obj.location, we actually just zero it out now before applying the world matrix. also removed the temp_override cause transform already does that.


Reonu (approved):
works on my machine™

PR #597, opened 2025-10-30
[OoT] Improve how lighting is handled

Yanis002:
This adds an extra list of light settings to the time of day mode, in some cases (or by personal preference) you may want to have more than 4 settings in that mode and fast64 is currently forcing that, so that basically address the `TODO` there is about it

Also this PR is fixing a bug where having `LIGHT_MODE_TIME` as the custom setting wouldn't export everything


Yanis002:
will be merged in 24 hours from now unless something else comes up

Issue #602, opened 2025-11-09
[OoT] OoT DL exporter exports 2 DLs with the same name (redefinition) when the DL has opaque and transparent materials


Reonu:
When a mesh contains both opaque and transparent materials and you try to export it with the OoT DL exporter you get an output like this which doesn't make sense and doesn't compile:

<img width="1001" height="594" alt="Image" src="https://github.com/user-attachments/assets/3edfc078-55d7-46b2-8374-cea5c8a8f285" />

Issue #601, opened 2025-11-09
"Export Path" option missing in the F3D exporter in the Fast64 tab


Reonu:
uhhhhhhhhhhhhhhhhhhhhhh

<img width="424" height="222" alt="Image" src="https://github.com/user-attachments/assets/71863d5e-1092-4725-8d6b-c939b93ebcac" />

Running latest main on blender 4.5.3

PR #605, opened 2025-11-10
Fix for issues #601 and #602

Yanis002:
fix for #601 and #602


Reonu (approved):
It do be working, thank you yanis you're the best

PR #622, opened 2025-11-20
[OoT] Introduce feature set

Yanis002:
supersedes #432, the only blocker on the other PR were conflicts and my own review so I addressed both, already marking as merge soon since the review process is already done


Reonu (approved):
Doesn't seem to break anything. I tested exporting HackerOoT-exclusive features, seems to work identically to before.

PR #639, opened 2025-12-03
[OoT] Fixed collision exporter not cleaning up duplicates

### LINKED ISSUE #604: [OoT] Exporting a collision mesh spawns a rotated duplicate of the mesh in the viewport that stays there
After using the OoT collision exporter, a rotated duplicate of the mesh shows up in the viewport and stays there until deleted manually.

Image

Yanis002:
fixes #604


Reonu (approved):
works

PR #640, opened 2025-12-03
[OoT] Fixed DL Exporter using the wrong filename

### LINKED ISSUE #617: [OoT] OoT DL exporter writes wrong includes
With the following setup, where the mesh is called `grass_mesh`, the "Use Custom Filename" box is checked and the custom filename is set to `grass_blades`, Fast64 will write the wrong includes to the object file (`grass_mesh.c` and `grass_mesh.h` instead of `grass_blades.c` and `grass_blades.h`)

Image

Image

Yanis002:
fixes #617

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Fast64
 
-This requires Blender 3.2+. Blender 4.0+ is recommended.
+This requires Blender 3.2 - 5.0.1. Blender 4.0+ is recommended.
 
 Forked from [kurethedead/fast64 on BitBucket](https://bitbucket.org/kurethedead/fast64/src).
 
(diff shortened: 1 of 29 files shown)


fast64_internal/z64/f3d/operators.py
```diff
@@ -50,31 +50,31 @@ def ootConvertMeshToC(
...
+        fModel = OOTModel(filename, DLFormat, None)
         triConverterInfo = TriangleConverterInfo(obj, None, fModel.f3d, finalTransform, getInfoDict(obj))
         fMeshes = saveStaticModel(
             triConverterInfo, fModel, obj, finalTransform, fModel.name, not saveTextures, False, "oot"
         )
 
         # Since we provide a draw layer override, there should only be one fMesh.
         for fMesh in fMeshes.values():
-            fMesh.draw.name = name
+            fMesh.draw.name = filename
```
Dragorn421: I would keep using the object name `toAlnum(originalObj.name)` in these two spots as I think those are relevant for the exported symbol names

PR #668, opened 2026-02-17
[SM64] Fix level exports for base decomp/UltraSM64 repos

Anoliah:
I apparently did not double check that `lvl_set_current_level` was unmodified in HackerSM64 when testing #644 and accidentally caused non hacker repos to throw an error when exporting levels. I have updated the check for final return to instead look for the first non `0`/`FALSE` return, which is still a bit cringe since it prevents the function from being modified too heavily, but it is much more versatile now.


Lilaa3:
@jesusyoshi54 ahemmm awkward

Lilaa3 (approved):
I was hoping scut would review. Changes make sense to me though. Should be good to merge

PR #675, opened 2026-02-27
Fix oot dl import

Dragorn421:
f3dContext.ignore_tlut was set to True for new assets system for skeleton and scenes import, but not for dl imports.


Yanis002:
since it worked for me on HackerOoT (not up-to-date yet) I think we can consider this good to go then

PR #676, opened 2026-02-28
Fix link anim export

Dragorn421:
There were several bugs and missing features:

1) the folderName was still being checked even though isLink=True means it doesn't play a role (the UI doesnt display it)
2) ootGetObjectPath was called for isLink=True with the folderName which again isn't relevant for link anims exports
3) `assets/misc/link_animetion/` and `assets/objects/gameplay_keep/` were not being created if missing (which is the case post "new" assets system)
4) link_animetion.c and gameplay_keep.c were being created in the case of "new" assets system tm, when the proper thing to do is to just export the .c to assets/ and link the relevant .o files in the spec, which is now done in this PR

### DIFF
--- a/fast64_internal/z64/animation/properties.py
+++ b/fast64_internal/z64/animation/properties.py
@@ -31,9 +31,10 @@ class OOTAnimExportSettingsProperty(PropertyGroup):
 
     def draw_props(self, layout: UILayout):
         layout.label(text="Exports active animation on selected object.", icon="INFO")
-        layout.prop(self, "isCustomFilename")
-        if self.isCustomFilename:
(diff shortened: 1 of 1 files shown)


Yanis002 (commented):
seems to work properly

Image

so I guess the order doesn't matter? that's good to know 💪 
I'm not particularily fan of this filename format but I can live with it, also I just realised the custom filename option isn't working for some reasons

Dragorn421:
I ended up hiding the "custom filename" option (if exporting link anims), because link anims are made of two sets of two files, so specifying a single custom filename can't work

Yanis002 (approved):

PR #680, opened 2026-03-08
Fix DL/skeleton importers for committed assets

Dragorn421:
It seems Fast64 was always looking in extracted/ for importing (except for importing Link which was fixed previously already), so this PR checks assets/objects/ first before going to extracted/.
Also change `ootGetIncludedAssetData` to not assume that `[extracted/]assets/*.c` files corresponding to a `[extracted/]assets/*.h` file are named with the same stem (like `assets/objects/gameplay_keep/gameplay_keep.c` for `assets/objects/gameplay_keep/gameplay_keep.h`) and instead look for all .c files that are in the same folder as the .h

I have tested importing
- DL: `gLinkAdultLeftHandHoldingMasterSwordNearDL` from `object_link_boy` (committed)
- DL: `gBoulderFragmentsDL` from `gameplay_keep` (committed)
- DL: `object_bombiwa_DL_0009E0` from `object_bombiwa` (not committed)
- skeleton: adult link
- skeleton: `gZoraSkel` from `object_zo` (committed)
- skeleton: `gGerudoRedSkel` from `object_geldb` (not committed)

### DIFF
--- a/fast64_internal/z64/skeleton/importer/functions.py
+++ b/fast64_internal/z64/skeleton/importer/functions.py
@@ -267,10 +267,21 @@ def ootImportSkeletonC(basePath: str, importSettings: OOTSkeletonImportSettings)
         isLink = False
         restPoseData = None
 
-    filepaths = [
-        ootGetObjectPath(isCustomImport, importPath, folderName, True),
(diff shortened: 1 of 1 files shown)


Yanis002 (approved):
seems to work

PR #681, opened 2026-03-13
[OoT] Fix/workaround decomp extracting params macroified

Dragorn421:
Decomp recently started extracting params macroified like `ENDOOR_PARAMS_NODATA(DOOR_ROOMLOAD, false)` instead of as just raw hex `0x003F`
for example this is an excerpt from `extracted/gc-eu-mq-dbg/assets/scenes/dungeons/Bmori1/Bmori1_scene_02000090_TransitionActorEntryList.inc.c`:
```c
    {
        {
            // { room, bgCamIndex }
            { 11, -1 },
            { 8, -1 },
        }, // sides
        ACTOR_EN_DOOR,
        {  -1502,    523,  -1864 }, // pos
        -0x8000, // rotY
        ENDOOR_PARAMS_NODATA(DOOR_ROOMLOAD, false) /* 0x003F */, // params
    }, // 4
```

This breaks fast64 importing for two unrelated reasons:
- fast64 does not expect additional commas in each entry, so I introduced `split_c_on_commas` which splits on commas but ignoring commas inside parentheses
- fast64 expects to be able to parse the params; this is worked around

The workaround for the second point is to try set the params value, and if it fails then fallback to using the actor as "custom", setting the custom actor id and custom params.

### DIFF
--- a/fast64_internal/z64/exporter/scene/actors.py
+++ b/fast64_internal/z64/exporter/scene/actors.py
@@ -40,6 +40,24 @@ class TransitionActor(Actor):
             + ("\n" + indent + "},\n")
         )
 
+def split_c_on_commas(s: str):
+    parts: list[str] = []
+    part = ""
+    parens_depth = 0
+    for c in s:
+        if c == "(":
+            parens_depth += 1
+        if c == ")":
+            parens_depth -= 1
+            if parens_depth < 0:
+                parens_depth = 0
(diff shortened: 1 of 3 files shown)


Yanis002 (approved):
it seems these changes are working like expected (I tested with kokiri forest and fire temple), though in the future I think we should add proper support for this instead of relying on the `custom` stuff

PR #691, opened 2026-03-30
[OoT] Fix include reading for DL imports

Dragorn421:
fast64 was not looking in extracted/VERSION/ for DL imports, causing e.g. imports of Link DLs like `gLinkAdultLeftHandHoldingMasterSwordNearDL` to fail due to not finding eg `gHilite1Tex_WIDTH`

The fix is to pass a list to `ootGetIncludedAssetData` as `basePaths` instead of the previously singular `basePath` which basically corresponds to the repo include path

This fix could be applied to other import things too (like anim and skeleton import) but it's not necessary at the moment so I'd rather not touch anything there :^)

I did not test this super thoroughly, I'm just assuming it didn't break anything

### DIFF
--- a/fast64_internal/z64/animation/importer/functions.py
+++ b/fast64_internal/z64/animation/importer/functions.py
@@ -69,7 +69,7 @@ def ootImportNonLinkAnimationC(armatureObj, filepath, animName, actorScale, isCu
     animData = getImportData([filepath])
     if not isCustomImport:
         basePath = bpy.path.abspath(bpy.context.scene.ootDecompPath)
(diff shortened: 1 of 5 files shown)


Dragorn421:
Got two reports that this fix works
HM https://discord.com/channels/388361645073629187/451783162859749386/1488113082633949285
HackerN64 https://discord.com/channels/879505781916053544/996101187621359656/1493204280549707806

Yanis002 (approved):

Yanis002:
I forgot about this PR lol, thanks to everyone who tested it :saluting_face:

PR #697, opened 2026-04-15
[SM64] Fix bad FModel names

Lilaa3:
Normalize names before passing them to convert geolayout, but also normalize in the fmodel class like geolayout does.


Anoliah (approved):
Threw some armature exports at it and everything came out fine

PR #706, opened 2026-05-24
[SM64] Fix animation exporter and importer for Blender 5.1

Lilaa3:


Dragorn421 (approved):
any idea why this is required and why the error this fixes pops up only in blender 5.1 ?

Dragorn421:
This fixes errors when import/export ing sm64 animations in Blender 5.1

The changelog for Blender 5.1 mentions upgrading python according to "VFX platform 2026" https://developer.blender.org/docs/release_notes/5.1/python_api/

If we go look at that "VFX platform" specs, we can see that numpy is bumped from 1.26.x to 2.3.x (!) https://vfxplatform.com/component.html?id=numpy

Hence we're experiencing numpy 2.0.0 breaking changes

In particular (see e.g. https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion ) adding an integer (python `int`) to e.g. a `np.uint16` value, now raises an exception on overflow

Hence the fixes in this PR which consist in converting numpy integers to python integers

Dragorn421 (approved):

PR #707, opened 2026-05-26
[OoT] Fix scene_table drawConfig export

Dragorn421:
The draw config value written to the scene_table was using the enum key (like `sdc_default`) instead of the ID (like `SDC_DEFAULT`) (see `oot_enum_data.xml`)


Yanis002 (approved):
I can't test right now but this seems good to me

PR #708, opened 2026-05-26
f3d_parser: fix ast.Str, ast.Num deprecated

Dragorn421:
Those are deprecated since python 3.8 and removed in python 3.14

ast.Constant is used for all constants since python 3.8, and the earliest blender version we support (3.2) ships python 3.10


Lilaa3 (approved):
I think I was already doing this in my own eval function for SM64 custom commands, I should double check this on my arch system at some point. Thank you!

Lilaa3:
> I think I was already doing this in my own eval function for SM64 custom commands, I should double check this on my arch system at some point. Thank you!

Just checked, yeah I did it correctly.

PR #711, opened 2026-05-29
Blender 5.1 support: fix vertex colors import and bump max version in README

Dragorn421:
While testing 5.1.2 I found out that for some reason vertex colors were no longer imported correctly. I still don't know why the deprecated vertex_colors API breaks in how fast64 uses it (I can't replicate it with simpler code)

With that fix oot seems to be working fine on all fronts so bumping the recommended blender version range max to 5.1.2 (the last of the 5.1 releases it seems for now). Testing for other games welcome.

### DIFF
--- a/fast64_internal/f3d/f3d_parser.py
+++ b/fast64_internal/f3d/f3d_parser.py
@@ -1902,10 +1902,10 @@ class F3DContext:
             # There will be one loop for every vertex
             uv_layer[i].uv = self.verts[i].uv
 
-        # The mesh.vertex_colors is deprecated since a long time,
+        # The mesh.vertex_colors API is deprecated since Blender 3.2,
         # and its usage by fast64 here breaks in Blender 5.1 somehow.
         # (can't replicate in simple cases)
-        if bpy.app.version < (5, 1, 0):
+        if bpy.app.version < (3, 2, 0):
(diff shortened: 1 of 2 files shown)


fast64_internal/f3d/f3d_parser.py
```diff
@@ -1902,13 +1902,26 @@ def createMesh(self, obj, removeDoubles, importNormals, callDeleteMaterialContex
...
+        if bpy.app.version < (5, 1, 0):
+            color_layer = mesh.vertex_colors.new(name="Col").data
+            for i in range(len(mesh.loops)):
+                color_layer[i].color = self.verts[i].rgb.to_4d()
+
+            alpha_layer = mesh.vertex_colors.new(name="Alpha").data
+            for i in range(len(mesh.loops)):
+                alpha_layer[i].color = [self.verts[i].alpha] * 3 + [1]
+        else:
+            col_attr = mesh.color_attributes.new("Col", "BYTE_COLOR", "CORNER")
```
Lilaa3, line 1917: This is techinically legacy code, as this has existed since 3.2.
So i think you should either:

- Check for < (3, 2, 0)
- Or just remove the other path

Lilaa3 (changes requested):

Lilaa3 (approved):

PR #716, opened 2026-06-26
[OoT] Auto detect non-literal actor scale

Jameriquiah:
auto detect actor scale was failing for skeleton imports that dont hardcode scale. for example `z_En_Hy` derives it's scale from `sInit2Info` rather than a literal scale (like a `0.01f` for example), causing an import failure if imported with an overlay, for testing i used `object_cne/gHylianYoungWomanSkel`.

auto detect actor scale will now locate scale expressions and detect the proper scale from there.


Dragorn421 (approved):
Also tested:
`object_hata/gFlagpoleSkel`, `ovl_En_Hata`: `Actor_SetScale(&this->dyna.actor, 1.0f / 75.0f);` :white_check_mark: 

I think this feature (actor scale detection) isn't super useful tbh, but it's nice to have and the code is clean and readable so sure

Jameriquiah:
thats fair, most skeletons can be imported without an overlay so its not something strictly necessary. but not entirely sure if _every_ skeleton needs an overlay to import 100% properly. that and more convenient for the user if they are using one since they dont have to bother unchecking stuff or assuming hey this thing doesnt work bc error or anything

PR #719, opened 2026-08-01
Add option to include complete CI format details in texture filenames

LucretiaArc:
Texel format information is missing from the filenames of exported CI textures when using the "Save Textures As PNGs" option, which is a problem when converting those textures. Some toolchains could work around this by assuming the texel format (like OoT does), but to allow projects to fully support exports with PNG CI textures, the texel format should be included in the filename.

I haven't verified the adjustments in `z64/model_classes.py` via export, but the result should be the same when the new option is disabled, and the toolchains for those projects would need to be modified to accept exports with the option enabled.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -113,6 +113,8 @@ class F3D_GlobalSettingsPanel(bpy.types.Panel):
         if context.scene.f3d_type in {"F3DEX3", "T3D"}:
             prop_split(col, context.scene, "packed_normals_algorithm", "Packed normals alg")
         col.prop(context.scene, "saveTextures")
+        if context.scene.saveTextures:
+            col.prop(context.scene, "textureNameIncludesCiFormat")
         col.prop(context.scene, "exportInlineF3D", text="Bleed and Inline Material Exports")
         if context.scene.exportInlineF3D:
             multilineLabel(
@@ -492,7 +494,10 @@ def register():
     bpy.types.Scene.gameEditorMode = bpy.props.EnumProperty(
         name="Game", default="SM64", items=gameEditorEnum, update=gameEditorUpdate
     )
-    bpy.types.Scene.saveTextures = bpy.props.BoolProperty(name="Save Textures As PNGs (Breaks CI Textures)")
+    bpy.types.Scene.saveTextures = bpy.props.BoolProperty(name="Save Textures As PNGs (May Break CI Textures)")
+    bpy.types.Scene.textureNameIncludesCiFormat = bpy.props.BoolProperty(
+        name="Include CI Format In File Name", default=False
+    )
     bpy.types.Scene.exportHiddenGeometry = bpy.props.BoolProperty(name="Export Hidden Geometry", default=True)
     bpy.types.Scene.exportInlineF3D = bpy.props.BoolProperty(
         name="Bleed and Inline Material Exports",
@@ -529,6 +534,7 @@ def unregister():
     del bpy.types.Scene.fullTraceback
     del bpy.types.Scene.ignoreTextureRestrictions
     del bpy.types.Scene.saveTextures
+    del bpy.types.Scene.textureNameIncludesCiFormat
     del bpy.types.Scene.gameEditorMode
     del bpy.types.Scene.exportHiddenGeometry
     del bpy.types.Scene.blenderF3DScale
--- a/__init__.py
+++ b/__init__.py
@@ -103,29 +103,30 @@ class F3D_GlobalSettingsPanel(bpy.types.Panel):
         col = self.layout.column()
         col.scale_y = 1.1  # extra padding
 
-        fast64_settings: Fast64Settings_Properties = context.scene.fast64.settings
+        scene = context.scene
+        fast64_settings: Fast64Settings_Properties = scene.fast64.settings
 
-        col.prop(context.scene, "f3d_simple", text="Simple Material UI")
+        col.prop(scene, "f3d_simple", text="Simple Material UI")
         col.separator()
 
         col.label(text="Saved to Repo Settings file", icon="PROPERTIES")
-        prop_split(col, context.scene, "f3d_type", "Microcode")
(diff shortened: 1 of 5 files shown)


__init__.py
```diff
@@ -492,7 +494,10 @@ def register():
     bpy.types.Scene.gameEditorMode = bpy.props.EnumProperty(
         name="Game", default="SM64", items=gameEditorEnum, update=gameEditorUpdate
     )
-    bpy.types.Scene.saveTextures = bpy.props.BoolProperty(name="Save Textures As PNGs (Breaks CI Textures)")
+    bpy.types.Scene.saveTextures = bpy.props.BoolProperty(name="Save Textures As PNGs (May Break CI Textures)")
+    bpy.types.Scene.textureNameIncludesCiFormat = bpy.props.BoolProperty(
```
Dragorn421: You should add this prop to `Fast64Settings_Properties` instead, then access it with `scene.fast64.settings.yourProp`

this is to avoid polluting the scene "namespace"

Dragorn421 (commented):
Thanks for the PR :)

So based on `getTextureSuffixFromFormat` your proposal is (when the option is checked in fast64 settings) replace `.ci[48].png` with `.ci[48]-{rgba16,ia16}.png`
I think that's a good idea and it looks easy to implement in oot at least
I think it'd be better to use an underscore `_` as separator for consistency with oot's separators in texture file extension information (eg shared tluts are indicated with a `.tlut_theTlutName.` suffix) but lmk why you chose `-` if you have a reason

I'll raise the topic of this suffix idea in the n64brew discord since that may be used there too, may as well discuss it https://discord.com/channels/205520502922543113/974342113850445874/1533042856388853830

Out of curiosity what are you needing IA palettes for? I find them pretty much useless

__init__.py
LucretiaArc: I used a top-level scene property here to match `saveTextures` and other F3D Global Settings properties, since I think it's sensible that this setting be a "sub-setting" of "Save Textures As PNGs".

I'm happy to change it if it makes more sense, but in that case, which of these would be preferred?

- Move the setting to the Fast64 Global Settings panel, and:
  - Appear independently of whether "Save Textures As PNGs" is set, even though it does very little when that setting isn't enabled
  - Appear depending on the value of "Save Textures As PNGs", even though it's in a different panel
- Keep the setting where it is, but access `fast64.settings` in `F3D_GlobalSettingsPanel.draw()`, unlike the other settings there

LucretiaArc:
> So based on `getTextureSuffixFromFormat` your proposal is (when the option is checked in fast64 settings) replace `.ci[48].png` with `.ci[48]-{rgba16,ia16}.png`
> I think that's a good idea and it looks easy to implement in oot at least
> I think it'd be better to use an underscore `_` as separator for consistency with oot's separators in texture file extension information (eg shared tluts are indicated with a `.tlut_theTlutName.` suffix) but lmk why you chose `-` if you have a reason
> I'll raise the topic of this suffix idea in the n64brew discord since that may be used there too, may as well discuss it https://discord.com/channels/205520502922543113/974342113850445874/1533042856388853830

I chose a hyphen mostly just as a sensible default, but if OoT already has an established separator for similar information then I'm happy to go with an underscore, pending any feedback from the N64brew discussion.

> Out of curiosity what are you needing IA palettes for? I find them pretty much useless

I don't have anything specific in mind, I'm more interested in making sure that the format is available for people who want to use it than actually using it in this case. I'm working on HackerSM64's support for CI texture image conversion, and it would be nice to build that in a way that supports IA palettes in case anyone wants to use them.

__init__.py
Dragorn421: option "Keep the setting where it is, but access fast64.settings in F3D_GlobalSettingsPanel.draw(), unlike the other settings there" would be preferred

It just so happens we haven't added settings to the "F3D Global Settings" panel since introducing the scene.fast64 property group, this would be the first one

fast64_internal/utility.py
```diff
@@ -1813,10 +1813,21 @@ def ootGetBaseOrCustomLight(prop, idx, toExport: bool, errIfMissing: bool):
...
 
 
-def getTextureSuffixFromFormat(texFmt):
-    # if texFmt == "RGBA16":
-    #     return "rgb5a1"
-    return texFmt.lower()
+def getTextureSuffixFromFormat(texFmt: str, ciFmt: str | None, isPalette: bool):
+    if bpy.context.scene.textureNameIncludesCiFormat:
+        if ciFmt:
+            fmtName = f"{texFmt}-{ciFmt}"
```
Dragorn421:
```suggestion
            fmtName = f"{texFmt}_{ciFmt}"
```
cf "oot uses _ for something similar"
and rasky from libdragon says he's ok with this

Dragorn421:
If you want you can take a look at the document describing how oot handles pngs, in particular how it handles CI formats (which oot does use, contrarily to sm64 which afaik does not use any) https://github.com/zeldaret/oot/blob/main/docs/assets/images.md

LucretiaArc:
Oops, forgot to comment to notify after marking conversations as resolved. Better late than never, I suppose :)

Dragorn421 (approved):
Sorry I did see the changes just didn't take the time to test until now :)

I'll point out that fast64 does not write file names that work with oot's palette generation, ideally that would be brought in line by updating fast64 before whatever fast64 currently does is applied to another repo and we need different handling per repo
In more details, fast64 for example wrote this
```c
u64 Cube_grayTex_ci8_rgba16[] = {
	#include "assets/objects/testpr/grayTex.ci8_rgba16.inc.c"
};

u64 Cube_grayTex_pal_ci8_rgba16[] = {
	#include "assets/objects/testpr/grayTex.ci8_rgba16.pal"
};
```
where the palette filename is `grayTex.ci8_rgba16.pal` instead of what the build system generates `grayTex.tlut.rgba16.inc.c`

PR #723, opened 2026-08-17
[OoT] fix draw layer suffix always being present

Dragorn421:
# Context

OoT DList exporter

Fast64 allows setting draw layers by material. This splits the mesh by material on export, between opaque/translucent/overlay draw layers, writing out different display lists.

It was reported last year https://github.com/Fast-64/fast64/issues/602 that fast64 however was writing those different display lists using the same name, obviously causing errors on compile due to the duplicate symbol name.

The issue was fixed in https://github.com/Fast-64/fast64/pull/605 by appending a "_drawlayer_dl" suffix to the written draw layer display list

However there have been complaints since then about this behavior since you can't name a single-draw-layer-using mesh properly (like `gMyCubeDL`) and have it be exported under that name, it would export as eg `gMyCubeDL_opaque_dl`

# This PR

This PR makes the suffix be omitted when a mesh is composed of a single draw layer. When several draw layers are present, it uses the draw layer suffixes.

# Testing

I tested:

- DL export with a single draw layer
- DL export with two draw layers
- deku tree import-export
- gerudo skeleton import-export
- epona skinskeleton import-export

### DIFF
--- a/fast64_internal/z64/f3d/operators.py
+++ b/fast64_internal/z64/f3d/operators.py
@@ -32,12 +32,16 @@ from ..utility import (
 
 
 class OOTF3DGfxFormatter(OOTGfxFormatter):
-    def __init__(self, scrollMethod):
+    def __init__(self, scrollMethod, use_draw_layer_suffix=True):
         OOTGfxFormatter.__init__(self, scrollMethod)
+        self.use_draw_layer_suffix = use_draw_layer_suffix
 
     # override the function to give a custom name to the DL array
     def drawToC(self, f3d, gfxList, layer: Optional[str] = None):
-        return gfxList.to_c(f3d, name_override=f"{gfxList.name}_{layer.lower()}_dl")
+        name_override = None
+        if self.use_draw_layer_suffix and layer is not None:
+            name_override = f"{gfxList.name}_{layer.lower()}_dl"
+        return gfxList.to_c(f3d, name_override=name_override)
 
 
 def ootConvertMeshToC(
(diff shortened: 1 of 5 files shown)


fast64_internal/z64/f3d/operators.py
```diff
@@ -71,7 +75,11 @@ def ootConvertMeshToC(
             triConverterInfo, fModel, obj, finalTransform, fModel.name, not saveTextures, False, "oot"
         )
 
-        # Since we provide a draw layer override, there should only be one fMesh.
+        # fMeshes is a dict where keys are draw layer names and values are the mesh for that draw layer.
+        # If there is more than one mesh we need to append a draw layer suffix on export to the original mesh name,
+        # to avoid exporting the same name twice.
+        use_draw_layer_suffix = len(fMeshes) > 1
```
Yanis002, line 81: I assume you can't do `OOTF3DGfxFormatter(ScrollMethod.Vertex, use_draw_layer_suffix=len(fMeshes) > 1)` directly? no big deal I'm just curious

Yanis002 (approved):

fast64_internal/z64/f3d/operators.py
Dragorn421: pretty sure you could, it just seems cleaner to externalize it to a variable and this way I have somewhere clear to put the comment

Trueffeloot:
DL exporting works as it should. I didn't try the other things like importing and exporting a scene though because I don't know how these files should be looking afterwards.

PR #724, opened 2026-08-17
Fix material texture switch on face select

Dragorn421:
The "normal" blender experience is that when you click a face that is basically assigned to a texture, blender switches to that texture in the image views.

This is currently broken in Fast64

This PR fixes the behavior (in some blender versions, see below) by simply selecting the Tex0 texture node in the material library instead of whatever was select before

blend file modified with blender 3.2.2

Testing:
- Fix doesn't work in 4.2.11 on my machine
- Fix works in 4.2.23 on my machine
- Fix works in 4.5.13 on my machine
- Fix works in 5.2.0 on my machine

Note the fix only applies to new materials, including after recreating nodes on "old" materials. Not sure why because stuff should just be linked and automatically update but oh well
https://discord.com/channels/874816709855440926/874816710639767585/1542106252044410921


Dragorn421:
I got confirmation from biobak on discord this works so merging soon tm

PR #726, opened 2026-08-19
[oot] parseScene: show failed path on error

Dragorn421:


fast64_internal/z64/importer/scene.py
```diff
@@ -114,14 +114,22 @@ def parseScene(
...
 
     if not file_path.exists():
-        raise PluginError("ERROR: scene not found!")
+        raise PluginError(f"ERROR: scene not found at {' nor '.join(map(str, tried_file_paths))}!")
 
-    sceneData = file_path.read_text()
+    try:
+        sceneData = file_path.read_text()
+    except Exception:
+        print("file_path =", file_path)
```
Yanis002: I think this should have something to indicate it's the except case otherwise it could be confusing to see this print 🤔

fast64_internal/z64/importer/scene.py
```diff
@@ -114,14 +114,22 @@ def parseScene(
         file_path = Path(sceneFolderPath).resolve() / f"{sceneName}.c"
     is_single_file = True
 
+    tried_file_paths = []
+    tried_file_paths.append(file_path)
```
Yanis002: technically we could save a line by doing this but I don't care much
```suggestion
    tried_file_paths = [file_path]
```

fast64_internal/z64/importer/scene.py
```diff
@@ -114,14 +114,22 @@ def parseScene(
...
     if not file_path.exists():
-        raise PluginError("ERROR: scene not found!")
+        raise PluginError(f"ERROR: scene not found at {' nor '.join(map(str, tried_file_paths))}!")
 
-    sceneData = file_path.read_text()
+    try:
+        sceneData = file_path.read_text()
+    except Exception:
+        print("file_path =", file_path)
+        raise
```
Yanis002, line 131: why not raising a `PluginError` here instead of a print followed by `raise`? but I feel like this keyword got a behavior I don't know about yet (I never saw `raise` being used alone like this before)

Dragorn421: `raise` re-raises the current exception as if it hadn't been caught

doing `raise NewException(...)` swallows the trace of the previous exception which is bad for debugging

could do `raise NewException(...) from e` (with `except Exception as e:`) which chains exceptions and doesn't lose trace information but I don't see a reason to over just reraising without making a PluginError

Yanis002: ahh I see, well I did learn something new :muscle:
it's good as it is currently then

Yanis002 (approved):

PR #727, opened 2026-08-19
[gltf] work around bug in the gltf addon shipped with Blender 5.2.0

Dragorn421:


Dragorn421:
No longer needed, fixed in Blender 5.2.1 which released recently

PR #728, opened 2026-08-19
[OoT] Link anim import: ensure bones rotation mode is "XYZ Euler"

Dragorn421:


Yanis002 (approved):
I assume sometimes it doesn't set it right? how fascinating, I love Blender 🙂

Dragorn421:
I don't think it can but someone just had the issue that their bone rotation modes were wrong so I figured this doesn't hurt

PR #736, opened 2026-09-13
Fix scrolling texture exports for UltraSM64 and base decomp

LucretiaArc:
UltraSM64 and base SM64 decomp don't declare `absi()` and `signum_positive()` in math_util.h, as HackerSM64 does:
- Decomp initially declared these functions in object_helpers2.h
- Starting with refresh 8, these functions are declared in object_helpers.h

This fixes build errors that occur after exporting levels/actors with scrolling textures using those functions. I have verified these exports don't produce errors/warnings related to texture scrolling using:

- HackerSM64 2.4.0, with refresh set to "HackerSM64 (v2.3.0)"
- UltraSM64 latest, with refresh set to "Refresh 16"
- Base decomp latest, with refresh set to "Refresh 16"

Output for refreshes 7 and earlier looks correct to me, but I have no way to build them to verify.


gheskett:
Should note, pretty sure UltraSM64 is still on Refresh 13

LucretiaArc:
> Should note, pretty sure UltraSM64 is still on Refresh 13

That's true, though it doesn't affect the outcome here. I've checked exports using Refresh 13 anyway for good measure, no errors/warnings.

PR #737, opened 2026-09-19
Sort flag sets so exported C is reproducible

Taldariner:
### Problem

Exporting the same scene twice can produce different C files, with nothing changed in the blend file. Only the order of flags inside a command differs:

```diff
-	gsSPClearGeometryMode(G_SHADING_SMOOTH | G_LIGHTING),
+	gsSPClearGeometryMode(G_LIGHTING | G_SHADING_SMOOTH),
```

Reproduction, run it a few times (each run is a fresh process):

```
blender -b --python-expr "import addon_utils; addon_utils.enable('fast64'); from fast64.fast64_internal.f3d.f3d_gbi import SPClearGeometryMode; print(SPClearGeometryMode({'G_LIGHTING','G_SHADING_SMOOTH'}).to_c())"
```

On my machine 6 runs printed `G_LIGHTING | G_SHADING_SMOOTH` five times and `G_SHADING_SMOOTH | G_LIGHTING` once. Exporting a corpus of test meshes, 8 runs of the same export produced 2 distinct sets of files.

This matters when an export is diffed against what is already committed in a decomp repo: the diff shows changes that are not changes. It also makes it impossible to confirm that a change to the exporter left the output alone, which is what I ran into while working on `f3d_writer.py`.

### Cause

`GbiMacro.getattr_virtual` joins iterable fields with `" | "` in iteration order, and several macros keep their flags in a `set`. Python randomizes string hashes per process, so the iteration order of a set of flag names changes from run to run.

### Fix

Sort the values when the field is a `set` or `frozenset`. Lists keep their order, so nothing positional is affected.

This is safe for every field it touches, because they are flag unions that `to_binary` ORs into a single word:

| Macro | Field |
|---|---|
| `SPGeometryMode` | `clearFlagList`, `setFlagList` |
| `SPSetGeometryMode` | `flagList` |
| `SPClearGeometryMode` | `flagList` |
| `SPLoadGeometryMode` | `flagList` |
| `DPSetOtherMode` | `mode0`, `mode1` |

`DPSetRenderMode` has its own positional `to_c` and is not affected. Binary output never depended on the order.

Import is unaffected as well: `f3d_parser` splits flag lists on `|` and does not care about order.

### Verification

A corpus of 22 generated meshes (flat and smooth, vertex colors, fog, UV seams, large texture mode, F3DEX3 packed normals and cel shading) exported to C for F3D, F3DEX2 and F3DEX3, plus a skinned geolayout export, on Blender 4.2.16 LTS:

- before the change, 8 runs of the same export produced **2** distinct outputs; after it, **1**;
- comparing the two trees, 130 files are identical except for 8 reordered flag commands, and the sorted output matches one of the orders the current code already emits;
- `to_binary` output is identical for both orders of the same flags, for every affected macro in F3D, F3DEX, F3DEX2 and F3DEX3.


fast64_internal/f3d/f3d_gbi.py
```diff
@@ -3445,7 +3445,15 @@ def getattr_virtual(self, field, static):
             else:
                 return field.name
         if hasattr(field, "__iter__") and type(field) is not str:
-            return " | ".join(map(str, field)) if len(field) else "0"
+            if not len(field):
+                return "0"
+            values = [str(value) for value in field]
+            if isinstance(field, (set, frozenset)):
+                # Sets iterate in hash order, and Python randomizes string hashes per
```
Dragorn421: If you don't mind I would avoid putting comments on what CPython does, that way I don't have to check they're correct. Just mentioning "sort for reproducible exports" is enough imo

Taldariner: Done.

Dragorn421 (approved):

